PR #3, opened 2019-11-12
changing commit cuz all of hardware except mica black is dumb

ArshanKhanifar:


ArshanKhanifar:
@aashmikamali  can you approve pls

PR #22, opened 2020-01-25
[SOFT-43] MCP23008 GPIO expander

copperium:


libraries/ms-drivers/inc/mcp23008_gpio_expander.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+// GPIO HAL interface for MCP23008 GPIO expander
```
ArshanKhanifar: mention what modules are required to be initialized for this

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
+#include "mcp23008_gpio_expander.h"
+
+#include <stdbool.h>
+#include "i2c.h"
+#include "mcp23008_gpio_expander_defs.h"
+
+StatusCode mcp23008_gpio_init(const Mcp23008I2CAddress i2c_address) {
```
ArshanKhanifar: don't think it's necessary, have the user initialize i2c. We may use i2c with other stuff as well, in that case this module makes it incompatible. In addition, you can't initialize i2c three times if you want to use this module three times.

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
...
+  // i2c_address only used in x86 - should this even have any implementation?
+  I2CSettings i2c_settings = {
+    .speed = I2C_SPEED_FAST,    //
+    .sda = CONFIG_PIN_I2C_SDA,  //
+    .scl = CONFIG_PIN_I2C_SCL,  //
+  };
+  // is this ok to do here?
```
ArshanKhanifar: your intuition is correct :) I don't think it's ok.

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
...
+  if (address->pin >= NUM_MCP23008_GPIO_PINS || settings->direction >= NUM_MCP23008_GPIO_DIRS ||
+      settings->state >= NUM_MCP23008_GPIO_STATES) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // Set the IODIR bit
+  prv_set_reg_bit(address->i2c_address, IODIR, address->pin,
+                  settings->direction == MCP23008_GPIO_DIR_IN);
+
+  mcp23008_gpio_set_state(address, settings->state);
```
ArshanKhanifar: If the direction is input, then you set state what happens? I feel like returning after setting the direction in case the direction is input would be better.

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode mcp23008_gpio_set_state(const Mcp23008GpioAddress *address,
+                                   const Mcp23008GpioState state) {
+  if (address->pin >= NUM_MCP23008_GPIO_PINS || state >= NUM_MCP23008_GPIO_STATES) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // Set the GPIO bit
```
ArshanKhanifar: unnecessary comment, safe to remove.

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
...
+StatusCode mcp23008_gpio_toggle_state(const Mcp23008GpioAddress *address) {
+  if (address->pin >= NUM_MCP23008_GPIO_PINS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // optimization: instead of using set_state and get_state, we read only once
+  uint8_t gpio_data = 0;
+  i2c_read_reg(I2C_PORT, address->i2c_address, GPIO, &gpio_data, 1);
+  gpio_data ^= 1 << address->pin;  // toggle the relevant bit
+  i2c_write_reg(I2C_PORT, address->i2c_address, GPIO, &gpio_data, 1);
```
ArshanKhanifar: some XOR magic! nice!

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,84 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode mcp23008_gpio_get_state(const Mcp23008GpioAddress *address,
+                                   Mcp23008GpioState *input_state) {
+  if (address->pin >= NUM_MCP23008_GPIO_PINS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // Read from GPIO
```
ArshanKhanifar: unnecessary comment.

libraries/ms-drivers/src/x86/mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,63 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode mcp23008_gpio_get_state(const Mcp23008GpioAddress *address,
+                                   Mcp23008GpioState *input_state) {
+  if (address->pin >= NUM_MCP23008_GPIO_PINS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // the gpio module does something strange - check w/ Arshan
```
ArshanKhanifar: what's strange? 🤔

projects/gpio_expander_demo/src/main.c
```diff
@@ -0,0 +1,56 @@
+#include "gpio.h"
+#include "i2c.h"
+#include "log.h"
+
+#define CONFIG_PIN_I2C_SCL \
+  { GPIO_PORT_B, 10 }
+#define CONFIG_PIN_I2C_SDA \
+  { GPIO_PORT_B, 11 }
+
+#define MCP_I2C_ADDR 0x22
```
ArshanKhanifar: feel free to safely remove all these defs.

projects/gpio_expander_demo/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+} Mcp2300OutputConfiguration;
+
+int main(void) {
+  gpio_init();
+  I2CSettings settings = {
+    .speed = I2C_SPEED_FAST,    //
+    .sda = CONFIG_PIN_I2C_SDA,  //
+    .scl = CONFIG_PIN_I2C_SCL,  //
+  };
+  i2c_init(I2C_PORT_2, &settings);
```
ArshanKhanifar: whenever you come in, let's see if we can fix the power dist board and try this again.

libraries/ms-drivers/test/test_mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,254 @@
...
+  TEST_ASSERT_OK(mcp23008_gpio_get_state(&address, &state));
+  TEST_ASSERT_EQUAL(MCP23008_GPIO_STATE_HIGH, state);
+
+  // test toggle high -> low
+  TEST_ASSERT_OK(mcp23008_gpio_toggle_state(&address));
+  TEST_ASSERT_OK(mcp23008_gpio_get_state(&address, &state));
+  TEST_ASSERT_EQUAL(MCP23008_GPIO_STATE_LOW, state);
+}
+
+// Test that an invalid address is caught.
```
ArshanKhanifar: I think this is repeated?

libraries/ms-drivers/test/test_mcp23008_gpio_expander.c
```diff
@@ -0,0 +1,254 @@
...
+  TEST_ASSERT_EQUAL(MCP23008_GPIO_STATE_LOW, state);
+
+  // test initializing to high
+  settings.state = MCP23008_GPIO_STATE_HIGH;
+  TEST_ASSERT_OK(mcp23008_gpio_init_pin(&address, &settings));
+  TEST_ASSERT_OK(mcp23008_gpio_get_state(&address, &state));
+  TEST_ASSERT_EQUAL(MCP23008_GPIO_STATE_HIGH, state);
+}
+
+// Test that an invalid address is caught.
```
ArshanKhanifar: repeating yourself.

libraries/ms-drivers/src/stm32f0xx/mcp23008_gpio_expander.c
copperium: In that case should I remove `state` from `MCP23008GpioSettings`? Alternatively I could check whether the direction is output and only set the state then.

libraries/ms-drivers/src/x86/mcp23008_gpio_expander.c
copperium: Oh gpio's x86 implementation keeps an `s_gpio_pin_input_value` array which is initialized to all 0s and never set, then in `gpio_get_state` it returns that if the direction is set to out (and the actual value of the pin if set to in) with the comment "behave how hardware does when the direction is set to out".

PR #23, opened 2020-01-27
[SOFT-6] Steering

AveryChiu64:
Hey Arshan, this is the same code from the Soft-6 Steering Board branch. However, since there were strange bugs on my computer, I created this new branch. It is the exact same as the old branch.


libraries/ms-helper/inc/exported_enums.h
```diff
@@ -65,3 +65,12 @@ typedef enum EERelayState {
...
   NUM_EE_RELAY_STATES,
 } EERelayState;
+
+typedef enum EESteeringEvents {
+  EE_STEERING_INPUT_HORN,
+  EE_STEERING_RADIO_PPT,
+  EE_STEERING_HIGH_BEAM_FORWARD,
+  EE_STEERING_HIGH_BEAM_REAR,
+  EE_STEERING_REGEN_BRAKE,
+  EE_STEERING_INPUT_CC_TOGGLE_PRESSED
```
ArshanKhanifar: remove this entire enum.

ArshanKhanifar: This shouldn't be in exported enums, it should be in `steering_events.h` and should be part of your project.

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,62 @@
...
+
+  Event e = { 0 };
+  TEST_ASSERT_OK(event_process(&e));
+  TEST_ASSERT_EQUAL(EE_STEERING_HIGH_BEAM_FORWARD, e.id);
+  TEST_ASSERT_EQUAL(GPIO_STATE_LOW, e.data);
+  TEST_ASSERT_EQUAL(STATUS_CODE_EMPTY, event_process(&e));
+}
+
+void test_steering_digital_input_cc_toggle() {
+  GpioAddress *cc_toggle_address = test_get_address(EE_STEERING_INPUT_CC_TOGGLE_PRESSED);
```
ArshanKhanifar: test_get_address gets steering_digital_input as an input, not an event.

PR #24, opened 2020-01-30
Soft 39: centre console

ArshanKhanifar:


ArshanKhanifar:
@g-jessmuir  @GeraldNDA  @copperium  can one of you guys please review and approve this? It has a lot of examples that I can point you guys to when you're writing your own tests... It's pretty huge but I'd appreciate any feedback :)

libraries/ms-common/src/can_ack.c
```diff
@@ -3,6 +3,7 @@
 // ACK requests currently ordered as they were created
 #include "can_ack.h"
 #include <string.h>
+#include "log.h"
```
GeraldNDA: lingering debug?

libraries/ms-helper/inc/exported_enums.h
```diff
@@ -33,6 +33,11 @@ typedef enum {
   NUM_EE_POWER_AUX_SEQUENCES
 } EEPowerAuxSequence;
 
+typedef enum {
+  EE_POWER_OFF_SEQUENCE_CONFIRM_MOTOR_CONTROLLER_DISABLED = 0,
+  NUM_EE_POWER_OFF_SEQUENCES
+} EEPowerOffSequence;
+
```
GeraldNDA: Seems weird to have an enum with just 1 define. This is for a CAN message, right? Maybe it should just be an empty message ...

projects/centre_console/inc/button_press.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+// Receives button presses on GPIO, and raises events.
+// Requires interrupts and gpio to be initialized.
```
GeraldNDA: Just realized that most files don't have something like this, but a small "module comment" would be useful in general. Especially since when things fail due to missing inits, the error isn't verbose at all ... that or some way to actually enforce that it was initialized.

projects/centre_console/inc/button_press.h
```diff
@@ -0,0 +1,21 @@
...
+  CENTRE_CONSOLE_BUTTON_POWER,
+  CENTRE_CONSOLE_BUTTON_NEUTRAL,
+  CENTRE_CONSOLE_BUTTON_HAZARD,
+  CENTRE_CONSOLE_BUTTON_EMERGENCY_STOP,
+  NUM_CENTRE_CONSOLE_BUTTONS
+} CentreConsoleButton;
+
+StatusCode button_press_init(void);
+
+GpioAddress *test_provide_button_addresses(void);
```
GeraldNDA: Weird name imo? I'd have gone with just "get_button_adresses" or smtg probably ...

projects/centre_console/inc/can_tx_retry_wrapper.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
GeraldNDA: This should be a library, no?

projects/centre_console/inc/centre_console_events.h
```diff
@@ -0,0 +1,92 @@
...
+
+typedef enum {
+  CENTRE_CONSOLE_EVENT_CAN_RX = 0,
+  CENTRE_CONSOLE_EVENT_CAN_TX,
+  CENTRE_CONSOLE_EVENT_CAN_FAULT,
+  NUM_CENTRE_CONSOLE_EVENTS  // 3
+} CentreConsoleEvent;
+
+typedef enum {
+  CENTRE_CONSOLE_POWER_EVENT_OFF = NUM_CENTRE_CONSOLE_EVENTS,  // 3
```
GeraldNDA: For power distro in the old car, they did NUM+1 instead. I think I agree with that more since then there's no overlap between the definitions. Also since you have more than two sets of events ... I think it might be better to disclude the "NUM_*_EVENTS" part since it isn't actually the number of events anymore. Not sure if there's a better way to do it ...

projects/centre_console/src/can_tx_retry_wrapper.c
```diff
@@ -0,0 +1,56 @@
...
+  return STATUS_CODE_OK;
+}
+
+static StatusCode prv_can_simple_ack(CanMessageId msg_id, uint16_t device, CanAckStatus status,
+                                     uint16_t num_remaining, void *context) {
+  CanTxRetryWrapperStorage *storage = (CanTxRetryWrapperStorage *)context;
+  if (num_remaining || status) {
+    storage->retry_count++;
+    LOG_DEBUG("retry_count: %d\n", storage->retry_count);
+    if (storage->retry_count >= storage->retries) {
```
GeraldNDA: multiple fault events for a single after retry count and on indefinite retries?

projects/centre_console/inc/drive_fsm.h
```diff
@@ -0,0 +1,30 @@
...
+#include "status.h"
+#include "stdbool.h"
+
+typedef enum {
+  DRIVE_STATE_NEUTRAL = 0,
+  DRIVE_STATE_PARKING,
+  DRIVE_STATE_DRIVE,
+  DRIVE_STATE_REVERSE,
+  NUM_DRIVE_STATES
+} DriveState;
```
GeraldNDA: This is an EE, no?

projects/centre_console/src/button_press.c
```diff
@@ -0,0 +1,47 @@
...
+    gpio_it_register_interrupt(&s_button_addresses[button], &interrupt_settings,
+                               INTERRUPT_EDGE_RISING, prv_button_interrupt_handler,
+                               &s_button_event_lookup[button]);
+  }
+  return STATUS_CODE_OK;
+}
+
+GpioAddress *test_provide_button_addresses(void) {
+  return s_button_addresses;
+}
```
GeraldNDA: Actually I kinda would've expected information to flow the other way. Where main passes this  pin information to the module which creates handlers for it

projects/centre_console/src/drive_fsm.c
```diff
@@ -0,0 +1,294 @@
...
+
+FSM_STATE_TRANSITION(state_set_relays) {
+  FSM_ADD_GUARDED_TRANSITION(DRIVE_FSM_INPUT_EVENT_MCI_RELAYS_CLOSED_DESTINATION_DRIVE,
+                             prv_destination_guard, state_drive);
+  FSM_ADD_GUARDED_TRANSITION(DRIVE_FSM_INPUT_EVENT_MCI_RELAYS_CLOSED_DESTINATION_REVERSE,
+                             prv_destination_guard, state_reverse);
+  FSM_ADD_GUARDED_TRANSITION(DRIVE_FSM_INPUT_EVENT_MCI_RELAYS_OPENED_DESTINATION_NEUTRAL_PARKING,
+                             prv_destination_guard, state_set_motorcontroller_output);
+  FSM_ADD_TRANSITION(DRIVE_FSM_INPUT_EVENT_FAULT, state_fault);
+}
```
GeraldNDA: MCI set output then relays? I thought we had it the other way around. I remember you also had "state_switch_direction" as being the same as set_mci_output which would make sense if the order was reversed ....

projects/centre_console/src/drive_fsm.c
```diff
@@ -0,0 +1,294 @@
...
+
+FSM_STATE_TRANSITION(state_ebrake_fault) {
+  FSM_ADD_TRANSITION(DRIVE_FSM_INPUT_EVENT_FAULT, state_ebrake_fault);
+  FSM_ADD_TRANSITION(DRIVE_FSM_INPUT_EVENT_MCI_EBRAKE_RELEASED, state_neutral);
+}
+
+typedef struct EventPair {
+  EventId success_event;
+  EventId fault_event;
+} EventPair;
```
GeraldNDA: whoa define struct in c file :eyes:

projects/centre_console/src/main.c
```diff
@@ -0,0 +1,28 @@
...
+
+int main(void) {
+  gpio_init();
+  interrupt_init();
+  gpio_it_init();
+  soft_timer_init();
+  event_queue_init();
+
+  button_press_init();
+  LOG_DEBUG("Hello from Centre Console!\n");
```
GeraldNDA: lingering debug?

projects/centre_console/src/power_main_precharge_monitor.c
```diff
@@ -0,0 +1,51 @@
...
+#include "can_ack.h"
+#include "can_msg_defs.h"
+#include "can_unpack.h"
+#include "centre_console_events.h"
+#include "event_queue.h"
+#include "exported_enums.h"
+
+void prv_timer_cleanup(PowerMainPrechargeMonitor *storage) {
+  storage->timer_id = SOFT_TIMER_INVALID_TIMER;
+}
```
GeraldNDA: it's only 1 line tho ... also no `static`

projects/centre_console/src/relay_tx.c
```diff
@@ -0,0 +1,31 @@
+#include "relay_tx.h"
+#include "can_transmit.h"
+
+#define NUM_TX_RETRIES 5
+
```
GeraldNDA: should be in header, no?

projects/centre_console/src/relay_tx.c
```diff
@@ -0,0 +1,31 @@
...
+  CanTxRetryWrapperSettings retry_settings = { .retries = NUM_TX_RETRIES };
+  can_tx_retry_wrapper_init(&storage->can_retry_wrapper_storage, &retry_settings);
+  return STATUS_CODE_OK;
+}
+
+static void prv_tx_relay_state(CanAckRequest *ack_ptr, void *context) {
+  RelayTxStorage *storage = (RelayTxStorage *)context;
+  uint16_t relay_mask = 1 << storage->relay_id;
+  uint16_t relay_state = storage->relay_state << storage->relay_id;
+  CAN_TRANSMIT_SET_RELAY_STATES(ack_ptr, relay_mask, relay_state);
```
GeraldNDA: I feel like you should make a macro for this, it'd make the previous one cleaner as well

projects/centre_console/src/button_press.c
ArshanKhanifar: this is for testing it.

projects/centre_console/inc/drive_fsm.h
ArshanKhanifar: it's the destination states used in the drive_fsm...

projects/centre_console/src/can_tx_retry_wrapper.c
ArshanKhanifar: well on infinite retries I wanna keep raising faults until it passes... we may have to change this behaviour based on the use case later...

projects/centre_console/inc/centre_console_events.h
ArshanKhanifar: yeah, Idk either... and yeah I never really figured out why they did NUM + 1

projects/centre_console/src/can_tx_retry_wrapper.c
ArshanKhanifar: But that was the use case I intended when writing this.

libraries/ms-common/src/can_ack.c
ArshanKhanifar: ah yes

libraries/ms-helper/inc/exported_enums.h
ArshanKhanifar: no this is more explicit, and power off sequence may have more steps to it... (e.g. open relays, unpower this board, etc...)

projects/centre_console/inc/button_press.h
ArshanKhanifar: absolutely!

projects/centre_console/inc/can_tx_retry_wrapper.h
ArshanKhanifar: yeah I'll move it to a library

projects/centre_console/src/button_press.c
GeraldNDA: It's used on line 20 of main tho :0

projects/centre_console/inc/drive_fsm.h
GeraldNDA: Very well. I would have imagined that the broadcasted drive states would be 1:1 with the destination states here. But if not, then this makes sense

projects/centre_console/inc/centre_console_events.h
GeraldNDA: The important thing is that you don't want overlap such that two events have the same ID so it's important that the each event enum has a different value. Not sure what your take on this might be but I think it might be better to do a big enum and then outline the groupings by using variable names and comments ... i.e.
```c
typedef enum {
   // CENTRE_CONSOLE_CAN_EVENT
  CENTRE_CONSOLE_CAN_EVENT_CAN_RX = 0,
  CENTRE_CONSOLE_CAN_EVENT_CAN_TX,
  CENTRE_CONSOLE_CAN_EVENT_CAN_FAULT,

   // CENTRE_CONSOLE_POWER_EVENT
  CENTRE_CONSOLE_POWER_EVENT_OFF
  CENTRE_CONSOLE_POWER_EVENT_ON_MAIN,
  CENTRE_CONSOLE_POWER_EVENT_ON_AUX,
  CENTRE_CONSOLE_POWER_EVENT_CLEAR_FAULT,
  CENTRE_CONSOLE_POWER_EVENT_PUBLISH_FAULT,
  CENTRE_CONSOLE_POWER_EVENT_FAULT,
  CENTRE_CONSOLE_POWER_EVENT_FAULT_POWER_MAIN_SEQUENCE,
  CENTRE_CONSOLE_POWER_EVENT_FAULT_POWER_AUX_SEQUENCE,

  NUM_CENTRE_CONSOLE_EVENTS
};
```

projects/centre_console/src/drive_fsm.c
ArshanKhanifar: it's only used internally :P rest of the modules don't need to know about this.

projects/centre_console/src/relay_tx.c
ArshanKhanifar: yeah I think Imma need this to be in the header to be able to test it 🤔

projects/centre_console/inc/button_press.h
ArshanKhanifar: this function gets used in the unit tests to get all the addresses. As they are internal to this module and shouldn't be used by the rest of the modules... the convention is to prefix functions that are ONLY used in tests with `test_`. In general one should try to avoid them, but if you find yourself redefining everything in your tests, then it makes sense to just write a test function like this. :)

projects/centre_console/src/button_press.c
ArshanKhanifar: oufie really? 👀 forgive me. Main isn't complete yet, and may be old, I'll probably just make it an empty loop for now just so that the builds pass...

projects/centre_console/src/main.c
ArshanKhanifar: A hello message wouldn't hurt. Also, when deploying we can turn debug messages off so they don't slow down the performance.

projects/centre_console/src/drive_fsm.c
ArshanKhanifar: the way state transitions work is that they look up the `destination` property in the storage to detect where they need to go next. If we go from drive to set mci output, it'll then try to set the relay states, which is unnecessary because they're already closed. BTW we're changing this up a bit since we don't have motorcontroller relays anymore... they're controlled via precharge, so drive_fsm is gonna change

projects/centre_console/inc/centre_console_events.h
ArshanKhanifar: We'd have to prefix everything with CENTRE_CONSOLE, and in the case of DRIVE_FSM_ events it's gonna be just too long and inconvenient. Also, the compiler will sometimes give us warnings if we use the wrong Enum data type when declaration and set it to another enum... having their data types separate will give us that added bonus.

projects/centre_console/inc/button_press.h
ArshanKhanifar: Added comments to everything.

ArshanKhanifar:
@GeraldNDA I addressed all your comments. Please look over again and approve it, it has a lot of examples that I can point people to when they're writing their own code. Thanks!

projects/centre_console/src/relay_tx.c
ArshanKhanifar: meh it's a test... you're right tho I'll clean it up in case I need to refactor my code then change this test...

ArshanKhanifar:
@GeraldNDA  please approve

projects/centre_console/inc/centre_console_events.h
GeraldNDA: That's fair I guess. I don't think we currently take advantage of the typing as far as I can tell.

projects/mci/src/main.c
```diff
@@ -42,7 +42,6 @@ int main(void) {
   soft_timer_init();
 
   prv_setup_system_can();
-  prv_mci_storage_init(&s_mci_storage);
```
GeraldNDA: I'm assuming this is to get it to compile?

ArshanKhanifar: ignore main :P

ArshanKhanifar:
@GeraldNDA  please look and approve.

ArshanKhanifar:
@GeraldNDA  finished writing tests for `power_off_sequence`. Please look and approve.

libraries/ms-common/src/can_ack.c
GeraldNDA: still?

projects/centre_console/src/brake_state_monitor.c
```diff
@@ -0,0 +1,10 @@
+#include "brake_state_monitor.h"
```
GeraldNDA: No tests, no functionality?

ArshanKhanifar: My bad, I added `pedal_monitor`, it does the same thing...

ArshanKhanifar:
@GeraldNDA  approve pleaseeeee

PR #27, opened 2020-02-04
Soft 40 mci firmware

g-jessmuir:
Still WIP, a few TODOs


g-jessmuir:
We should probably merge the mcp2515 driver improvements before merging this one since we'll need to update the interface used here anyways.

PR #30, opened 2020-02-05
[SOFT-46] blink_event_generator

copperium:


libraries/ms-common/inc/blink_event_generator.h
```diff
@@ -0,0 +1,32 @@
...
+typedef struct {
+  uint32_t interval_us;
+  uint16_t first_value;  // value of the first event; must be 1 or 0, default is 0 if not specified
+} BlinkEventGeneratorSettings;
+
+typedef struct {
+  uint32_t interval_us;
+  EventId event_id;
+  uint16_t first_value;
+  uint16_t next_value;
```
ArshanKhanifar: change this to `current_state`, and change `first_value` to `default_state`, and typedef an enum called  `BlinkerState`, as in `BLINKER_STATE_ON`, and `BLINKER_STATE_OFF`. Right now the user can pass any `uint16_t` as the first value, which wouldn't really make sense... but if you typedef them as `BlinkerState` it's clear.

libraries/ms-common/src/blink_event_generator.c
```diff
@@ -0,0 +1,43 @@
+#include "blink_event_generator.h"
+
+StatusCode blink_event_generator_init(BlinkEventGeneratorStorage *storage,
+                                      const BlinkEventGeneratorSettings *settings) {
+  if (settings->first_value > 1) {
```
ArshanKhanifar: you can keep this check, but use the typedef.

libraries/ms-common/src/blink_event_generator.c
```diff
@@ -0,0 +1,43 @@
...
+
+static void prv_raise_blink_event(SoftTimerId timer_id, void *context) {
+  BlinkEventGeneratorStorage *storage = context;
+  event_raise_priority(EVENT_PRIORITY_NORMAL, storage->event_id, storage->next_value);
+  storage->next_value = !storage->next_value;
+
+  soft_timer_start(storage->interval_us, &prv_raise_blink_event, storage, &storage->timer_id);
+}
+
+StatusCode blink_event_generator_start(BlinkEventGeneratorStorage *storage, EventId event_id) {
```
ArshanKhanifar: nice! basically calling start will reset it?

PR #37, opened 2020-02-08
[SOFT-90] lights_signal_fsm

copperium:
Notes/questions:
* should it be under `front_power_distribution`?
* should the output events be configurable? Currently they're fixed as `FRONT_POWER_DISTRIBUTION_GPIO_EVENT_SIGNAL_LEFT`, `FRONT_POWER_DISTRIBUTION_GPIO_EVENT_SIGNAL_RIGHT`, `FRONT_POWER_DISTRIBUTION_GPIO_EVENT_SIGNAL_HAZARD`.


projects/front_power_distribution/src/lights_signal_fsm.c
```diff
@@ -0,0 +1,110 @@
...
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_on, state_left_signal);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_right_event, &prv_guard_on, state_right_signal);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_hazard_event, &prv_guard_on, state_hazard_signal);
+}
+
+FSM_STATE_TRANSITION(state_left_signal) {
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_off, state_none);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_right_event, &prv_guard_on, state_right_signal);
```
ArshanKhanifar: I don't think we can go from left to right directly...

ArshanKhanifar: you can remove this

projects/front_power_distribution/src/lights_signal_fsm.c
```diff
@@ -0,0 +1,110 @@
...
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_hazard_event, &prv_guard_on, state_hazard_left_signal);
+}
+
+FSM_STATE_TRANSITION(state_right_signal) {
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_on, state_left_signal);
```
ArshanKhanifar: same goes for right to left

projects/front_power_distribution/src/lights_signal_fsm.c
```diff
@@ -0,0 +1,110 @@
...
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_on, state_hazard_left_signal);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_right_event, &prv_guard_on, state_hazard_right_signal);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_hazard_event, &prv_guard_off, state_none);
+}
+
+FSM_STATE_TRANSITION(state_hazard_left_signal) {
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_off, state_hazard_signal);
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_right_event, &prv_guard_on, state_hazard_right_signal);
```
ArshanKhanifar: I think first left will be turned off, then this transitions to hazard, then right will be turned on, then we'll go to hazard right... so this won't happen either.

projects/front_power_distribution/src/lights_signal_fsm.c
```diff
@@ -0,0 +1,110 @@
...
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_hazard_event, &prv_guard_off, state_left_signal);
+}
+
+FSM_STATE_TRANSITION(state_hazard_right_signal) {
+  SignalFsmStorage *storage = fsm->context;
+  FSM_ADD_GUARDED_TRANSITION(storage->signal_left_event, &prv_guard_on, state_hazard_left_signal);
```
ArshanKhanifar: same

projects/front_power_distribution/src/lights_signal_fsm.c
```diff
@@ -0,0 +1,110 @@
...
+static void prv_state_right_signal_output(Fsm *fsm, const Event *e, void *context) {
+  SignalFsmStorage *storage = context;
+  blink_event_generator_start(&storage->blink_event_generator,
+                              FRONT_POWER_DISTRIBUTION_GPIO_EVENT_SIGNAL_RIGHT);
+}
+
+static void prv_state_hazard_signal_output(Fsm *fsm, const Event *e, void *context) {
+  SignalFsmStorage *storage = context;
+  blink_event_generator_start(&storage->blink_event_generator,
+                              FRONT_POWER_DISTRIBUTION_GPIO_EVENT_SIGNAL_HAZARD);
```
ArshanKhanifar: I actually want this module to be also a library, so take the input events: SIGNAL_*_EVENT and output events: GPIO_SIGNAL_*_EVENT as configuration...

ArshanKhanifar: because both front and rear power distribution will use this :)

projects/front_power_distribution/inc/lights_signal_fsm.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+// FSM for handling transitions between signal states.
```
ArshanKhanifar: update these descriptions after you turn this into a library.

projects/front_power_distribution/test/test_lights_signal_fsm.c
```diff
@@ -0,0 +1,236 @@
...
+#include "interrupt.h"
+#include "lights_signal_fsm.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+#define TEST_BLINK_INTERVAL_US 6000
+
+#define TEST_LIGHTS_SIGNAL_ASSERT_EVENT_RAISED(event_id, event_data) \
```
ArshanKhanifar: I freakin love these!

libraries/ms-common/inc/blink_event_generator.h
```diff
@@ -0,0 +1,42 @@
...
+
+typedef enum {
+  BLINKER_STATE_OFF = 0,
+  BLINKER_STATE_ON,
+  NUM_BLINKER_STATES,
+} BlinkerState;
+
+typedef struct {
+  uint32_t interval_us;
+  BlinkerState default_state;  // if not specified, the default state is BLINKER_STATE_OFF
```
ArshanKhanifar: Nit. comment is a bit long. 
Something like "defaults to BLINKER_STATE_OFF" is enough :)

libraries/ms-common/inc/lights_signal_fsm.h
```diff
@@ -0,0 +1,38 @@
...
+
+typedef struct {
+  // Each event must have a data field of 1 for on or 0 for off.
+  EventId signal_left_input_event;
+  EventId signal_right_input_event;
+  EventId signal_hazard_input_event;
+  EventId signal_left_output_event;
+  EventId signal_right_output_event;
+  EventId signal_hazard_output_event;
+  uint32_t blink_interval_us;  // should this be input or #defined?
```
ArshanKhanifar: it should be input :)

libraries/ms-common/src/blink_event_generator.c
```diff
@@ -0,0 +1,68 @@
...
+  storage->current_state = new_state;
+
+  soft_timer_start(storage->interval_us, &prv_raise_blink_event_callback, storage,
+                   &storage->timer_id);
+}
+
+StatusCode blink_event_generator_start(BlinkEventGeneratorStorage *storage, EventId event_id) {
+  if (prv_is_active(storage)) {
+    if (event_id == storage->event_id) {
+      // restarting with the same event: do nothing to avoid a strange delay
```
ArshanKhanifar: Actually, I think restarting the timer here would be valuable. By that I mean just simply cancelling the old timer and starting a new timer. 

The reason this'll be good is because when we want to sync the rear lights and the front lights, the rear power dist board will send a `sync` message to the front board, and the front board can just simply call `start` and that will do the resetting. That way our lights will also be in sync after longer periods of time.

PR #40, opened 2020-02-10
Soft 122 mci output

GeraldNDA:
Roast me.

Currently has no mapping of precharge during the throttle region. Will test a sample mapping function soon. Will also take a look at different mapping functions but I think we should start with the barebones for now.

Tests are long, but I think it's useful to do it this way since you can get the name of the test. Probably going to look into forward/reverse w/ no pedals case since the current expected value is slightly unexpected. Also might look into why failing tests get stuck in deadlock, but it's a non-issue


projects/mci/inc/motor_controller.h
```diff
@@ -1,8 +1,13 @@
...
 
+#include "exported_enums.h"
+#include "generic_can.h"
+#include "pedal_rx.h"
 #include "precharge_control.h"
 
 typedef struct MotorControllerStorage {
   PrechargeStorage precharge_storage;
+  PedalRxStorage pedal_storage;
+  EEDriveOutput drive_state;
```
ArshanKhanifar: I think having the drive_fsm module expose this via a global variable would be better, than to keeping it here. Similar to how you get pedal values now...

projects/mci/src/mci_output.c
```diff
@@ -0,0 +1,81 @@
...
+  memcpy(&msg.data, data, sizeof(data));
+  generic_can_tx(storage->motor_can, &msg);
+}
+
+static void prv_handle_drive(SoftTimerId timer_id, void *context) {
+  MotorControllerStorage *storage = context;
+  PedalValues pedal_values = pedal_rx_get_pedal_values(&storage->pedal_storage);
+  MotorCanDriveCommand drive_command = { 0 };
+  // TODO(SOFT-122): Make sure test ensures that maps are continues
+  if (storage->drive_state == EE_DRIVE_OUTPUT_OFF) {
```
ArshanKhanifar: instead of accessing it from storage directly,  access it from the drive_fsm. I left Jess a PR to change it.

GeraldNDA:
pedal rx is module that I merged with #34 testing is pretty extensive but I probably should have used the macros. Remind me to do that later ...

ArshanKhanifar:
Neat, Imma use that in centre console!

ArshanKhanifar:
Resolve your merge conflicts please ^_^

PR #44, opened 2020-02-29
[SOFT-128] PCA9539R GPIO expander driver

copperium:
Similar to MCP23008, used on new front power distribution board.


libraries/ms-drivers/inc/i2c_driver_defs.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+// Common definitions for the I2C drivers (i.e. MCP23008 and PCA9539R).
+
+#define I2C_PORT I2C_PORT_2
+
+#define CONFIG_PIN_I2C_SCL \
```
ArshanKhanifar: Unfortunately these won't always be the case. We might put a MCP23008 chip on I2C_PORT_1. You're safe to remove this file entirely, and move the #define 's to the files that use them, which I'm assuming is only going to be the test file. Prefix them with TEST_* so it's clear that the #define is only meant to be used in the test.

PR #45, opened 2020-03-02
Soft 124 mci broadcast

GeraldNDA:
Fully functional in terms of implementation. Will depend on mcp2515 changes to be able to access the required messages.

imo that should be a separate issue than this module since it requires modifying the mcp2515 driver. But can be done before this merge.


projects/mci/test/test_mci_broadcast.c
```diff
@@ -0,0 +1,375 @@
...
+  s_expect_velocity = true;
+  s_expect_bus_measurement = true;
+
+  prv_send_measurements(LEFT_MOTOR_CONTROLLER, TEST_MCI_BUS_MEASUREMENT_MESSAGE, false);
+  prv_send_measurements(RIGHT_MOTOR_CONTROLLER, TEST_MCI_VELOCITY_MESSAGE, false);
+  prv_send_measurements(LEFT_MOTOR_CONTROLLER, TEST_MCI_VELOCITY_MESSAGE, false);
+  prv_send_measurements(RIGHT_MOTOR_CONTROLLER, TEST_MCI_BUS_MEASUREMENT_MESSAGE, false);
+
+  delay_ms(MOTOR_CONTROLLER_BROADCAST_TX_PERIOD_MS + 50);
+  MS_TEST_HELPER_CAN_TX(MCI_CAN_EVENT_TX);
```
ArshanKhanifar: put these four in a function and call that :)

projects/mci/test/test_mci_broadcast.c
```diff
@@ -0,0 +1,375 @@
...
+// Test 1: lb rv lv rb (check 2 output)
+void test_all_measurements_lb_rv_lv_rb() {
+  // motor_can tx and rx should happen immedietly
+  MotorControllerBroadcastStorage broadcast_storage = { 0 };
+  mci_broadcast_init(&broadcast_storage, &s_broadcast_settings);
+
+  s_expect_velocity = true;
+  s_expect_bus_measurement = true;
+
+  prv_send_measurements(LEFT_MOTOR_CONTROLLER, TEST_MCI_BUS_MEASUREMENT_MESSAGE, false);
```
ArshanKhanifar: I like your tests, they're rigorous, and granular. I usually prefer putting the inputs and the outputs in the test itself. I think it'd be neat if you could have your prv_send_measurements function to take the actual measurement value as an input, then in your callbacks you could set a global variable to capture the sent value, then make the assertions at the end of your test instead of the callbacks. I think that would make your tests easier to understand. Cuz when I'm looking at your tests now I'm like, oh two messages got TX and RX'd, now what? then I had to take a look at your callback to see if the assertions were being made. Putting your final assertions in your test makes them flow better :)

projects/mci/inc/mci_broadcast.h
```diff
@@ -0,0 +1,34 @@
...
+typedef struct MotorControllerBroadcastSettings {
+  GenericCan *motor_can;
+  MotorCanDeviceId device_ids[NUM_MOTOR_CONTROLLERS];
+} MotorControllerBroadcastSettings;
+
+typedef struct MotorControllerBroadcastStorage {
+  uint8_t bus_rx_bitset;
+  uint8_t velocity_rx_bitset;
+  MotorControllerMeasurements measurements;
+  MotorCanDeviceId ids[NUM_MOTOR_CONTROLLERS];
```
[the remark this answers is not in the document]
GeraldNDA: Prefer not to. It's technically not the full can id (technically) since you also have to include the message. This only counts as the device id and `can_device_ids` sounds like a pain for only internal use imo ...

PR #46, opened 2020-03-02
Soft 126 fix pedal rx

ArshanKhanifar:
@GeraldNDA I added the timeout as an input to make the tests run faster. The rest of the api should be the same. Take a look at the tests to see what I mean by writing "granular" tests. Please approve so I can start using this in `centre_console`


ArshanKhanifar:
@GeraldNDA  good catch! Yes I admit I did fit two tests into one for that one. didn't seem proper to rewrite the entire thing again only to put it in a loop.

PR #48, opened 2020-03-05
[SOFT-130] charger firmware

g-jessmuir:
Added code and tests for modules charger_control_pilot_monitor and charger_connection_sense


projects/charger/src/charger_control_pilot_monitor.c
```diff
@@ -0,0 +1,24 @@
...
+#include "event_queue.h"
+#include "gpio.h"
+#include "pwm_input.h"
+#include "status.h"
+
+void handle_pwm_event(Event e) {
+  if (e.id == PWM_READING_REQUEST) {
+    PwmInputReading reading = { 0 };
+    pwm_input_get_reading(PWM_TIMER_3, &reading);
+    event_raise(PWM_READING_VALUE, reading.dc_percent);
```
ArshanKhanifar: convert this to a current value, look at the specification.

projects/charger/src/main.c
```diff
@@ -0,0 +1,43 @@
+/*
```
ArshanKhanifar: /* */ style comments are not accepted.  Take a look at here: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards

projects/charger/test/test_connection_sense.c
```diff
@@ -0,0 +1,21 @@
...
+  gpio_init();
+  soft_timer_init();
+}
+
+void teardown_test(void) {}
+
+void test_init() {
+  // Tests init of timer reading ADC
+  TEST_ASSERT_OK(connection_sense_init());
+}
```
ArshanKhanifar: add more tests for connection sense. Mock out the `adc` module, and have it return different values to cover all the different branches. Put delays in your test and make sure that your charger connection state is getting polled correctly and it reports the values correctly.

projects/charger/test/test_control_pilot_monitor.c
```diff
@@ -0,0 +1,35 @@
...
+  TEST_ASSERT_EQUAL(e.id, PWM_READING_VALUE);
+}
+
+// Tests that raising a reading request will cause an event to be raised
+// with the read value
+void test_control_pilot_event_handle() {
+  event_raise(PWM_READING_REQUEST, 0);
+  Event e = { 0 };
+  // Process event raised here
+  event_process(&e);
```
ArshanKhanifar: use the `MS_TEST_HELPER_ASSERT_NEXT_EVENT` macro here.

ArshanKhanifar: also mock the `pwm_input` module, and make sure that the duty-cycle-to-current conversion is happening correctly.

projects/charger_test/src/main.c
```diff
@@ -0,0 +1,123 @@
+#include <stdbool.h>
```
ArshanKhanifar: I like test projects like this. naming convention for this should be something like: `test_<descriptive_test_name>`.  Over here seems like you're testing `pwm`,  `can_rx`, and `evse_connection_state`. So split it into three different projects: `test_charger_pwm`, `test_charger_mcp_rx`, and `test_charger_evse_conn`. Having these be granular will be super useful when we want to test with hardware.

g-jessmuir:
To be continued fall term 2020

PR #53, opened 2020-03-08
[SOFT-135] test mcp2515

g-jessmuir:
project to help smoke test the mcp2515, should work on charger interface and mci


ArshanKhanifar:
@g-jessmuir why this fail?

PR #60, opened 2020-04-19
[SOFT-148] Add fault handling and fault clearing to centre console.

ArshanKhanifar:
I'll need to explain this again, probably in a video... but basically the faults that occur during a transition are not that critical: aka. that state can go back to the state it previously was at...


libraries/ms-common/src/ms_test_helper_can.c
```diff
@@ -0,0 +1,25 @@
...
+#include "gpio.h"
+#include "interrupt.h"
+#include "soft_timer.h"
+
+StatusCode initialize_can_in_test(CanStorage *storage, SystemCanDevice device, EventId tx_event,
+                                  EventId rx_event, EventId fault_event) {
+  event_queue_init();
+  gpio_init();
+  interrupt_init();
+  soft_timer_init();
```
GeraldNDA, line 14: Might be unclear that this is happening if someone else were to use this.

 Also calling `_init` more than once should result in an error ... `status_ok_or_return` at the very least?

libraries/ms-common/inc/event_queue.h
```diff
@@ -48,6 +48,8 @@ StatusCode event_raise_priority(EventPriority priority, EventId id, uint16_t dat
 // for legacy compatibility.
 #define event_raise(id, data) event_raise_priority(EVENT_PRIORITY_NORMAL, (id), (data))
 
+#define event_raise_no_data(id) event_raise((id), (0))
+
```
GeraldNDA, line 52: Looks like the other definitions here have a descriptive comment, might wanna add one for this too.

projects/centre_console/src/drive_fsm.c
```diff
@@ -123,8 +125,12 @@ static void prv_set_motorcontroller_output(Fsm *fsm, const Event *e, void *conte
   DriveFsmStorage *storage = (DriveFsmStorage *)context;
   storage->current_state = DRIVE_STATE_TRANSITIONING;
   DestinationTransitionInfo info = s_destination_transition_lookup[storage->destination];
-  DriveFsmFault fault = { .fault_reason = DRIVE_FSM_FAULT_REASON_MCI_OUTPUT,
-                          .fault_state = info.mci_drive_output };
+  StateTransitionFault fault = {
+    .state_machine = DRIVE_FSM_STATE_MACHINE,
+    .fault_reason = (DriveFsmFaultReason){ .step = DRIVE_FSM_TRANSITION_STEP_MCI_OUTPUT,
+                                           .state = info.mci_drive_output }
+                        .raw
```
GeraldNDA, line 132: Is the `(DriveFsmFaultReason)` necessary here? I don't think I've seen that anywhere else.
Same with the `.raw`

GeraldNDA: Similar style thing used below, so I guess if it isn't then change those too

projects/centre_console/inc/centre_console_fault_reason.h
```diff
@@ -5,33 +5,35 @@
...
-  CENTRE_CONSOLE_FAULT_REASON_PRECHARGE_NOT_INITIATED,
-  CENTRE_CONSOLE_FAULT_REASON_MOTOR_RELAY_FAILED_TO_CLOSE,
-  NUM_CENTRE_CONSOLE_FAULT_REASONS
-} CentreConsoleFaultReason;
+// Used as the event data for the CENTRE_CONSOLE_POWER_EVENT_FAULT, and
+// DRIVE_FSM_INPUT_EVENT_FAULT events.
+typedef union StateTransitionFault {
+  struct {
+    StateMachine state_machine : 3;
+    uint8_t fault_reason : 8;
```
GeraldNDA, line 20: Actually, shouldn't this be `DriveFsmFaultReason` that way you don't have to use that yucky `((DriveFsmFaultReason){ ... }).raw` stuff?

projects/centre_console/inc/fault_monitor.h
```diff
@@ -0,0 +1,19 @@
...
+#define FAULT_MONITOR_WATCHDOG_TIMEOUT 4000
+
+typedef enum {
+  FAULT_STATUS_FAULT = 0,
+  FAULT_STATUS_ACK_FAULT,
+  FAULT_STATUS_OK,
+  NUM_FAULT_STATUS
+} FaultStatus;
+
+FaultStatus *get_fault_status(void);
```
GeraldNDA, line 17: I understand why you did this but I don't think this is a good idea ... One option would be to create a test-only setter method.

projects/centre_console/src/power_fsm.c
```diff
@@ -109,10 +85,21 @@ bool power_fsm_process_event(PowerFsmStorage *power_fsm, const Event *event) {
...
+static EventId s_event_lookup[] = {
+  [CENTRE_CONSOLE_POWER_EVENT_ON_MAIN] = POWER_MAIN_SEQUENCE_EVENT_BEGIN,
+  [CENTRE_CONSOLE_POWER_EVENT_ON_AUX] = POWER_AUX_SEQUENCE_EVENT_BEGIN,
+  [CENTRE_CONSOLE_POWER_EVENT_OFF] = POWER_OFF_SEQUENCE_EVENT_BEGIN,
+};
+
+static void prv_power_state_transitioning(Fsm *fsm, const Event *e, void *context) {
+  PowerFsmStorage *power_fsm = (PowerFsmStorage *)context;
+  power_fsm->previous_state = power_fsm->current_state;
+  prv_set_current_state(context, POWER_STATE_TRANSITIONING);
```
GeraldNDA, line 97: nitpick: should've used `power_fsm` here

libraries/ms-common/src/ms_test_helper_can.c
g-jessmuir: maybe change to something like `initialize_can_and_dependencies_in_test` or similar given that the convention is to just include the inits in the setup_test function

g-jessmuir:
Travis seems to be stuck, going to close this and open a new PR.

PR #62, opened 2020-04-27
[SOFT-138] Sync events for power distribution

copperium:
This changes `lights_signal_fsm` and `blink_event_generator` to support sending sync CAN messages and receiving sync events (converted from CAN by `power_distribution_can_rx_event_mapper`).


libraries/ms-common/src/lights_signal_fsm.c
```diff
@@ -80,17 +104,43 @@ static void prv_state_hazard_signal_output(Fsm *fsm, const Event *e, void *conte
...
   storage->signal_hazard_output_event = settings->signal_hazard_output_event;
+  storage->sync_event = settings->sync_event;
+  storage->sync_behaviour = settings->sync_behaviour;
+  storage->num_blinks_between_syncs = settings->num_blinks_between_syncs;
+  storage->blink_counter = 0;
 
   BlinkEventGeneratorSettings blinker_settings = {
     .interval_us = settings->blink_interval_us,
     .default_state = BLINKER_STATE_OFF,  // all lights default to off
+    .callback = &prv_blink_event_raised_callback,
```
g-jessmuir, line 142: in C the `&` is optional for function pointers and in most of our other code we don't include it, so if you want to be a hero you could remove them from the function pointers in this file, but I won't hold you to it

libraries/ms-common/src/lights_signal_fsm.c
```diff
@@ -80,17 +104,43 @@ static void prv_state_hazard_signal_output(Fsm *fsm, const Event *e, void *conte
   blink_event_generator_start(&storage->blink_event_generator, storage->signal_hazard_output_event);
 }
 
+static void prv_blink_event_raised_callback(BlinkerState new_state, void *context) {
+  SignalFsmStorage *storage = context;
+
+  // we only send sync events on rising edges
+  if (new_state == BLINKER_STATE_ON) {
+    storage->blink_counter++;
```
g-jessmuir, line 112: isn't the `blink_counter` only used by the sync event sender? should the increment be inside the following if statement?

libraries/ms-common/src/lights_signal_fsm.c
```diff
@@ -15,6 +16,14 @@ static bool prv_guard_off(const Fsm *fsm, const Event *e, void *context) {
   return e->data == 0;
 }
 
+// A bit of an abuse of the guard system: used to stop the blinker during sync event handling.
```
g-jessmuir, line 19: I am ok with this, good call

PR #63, opened 2020-05-05
[SOFT-51] power_distribution_publish_data

copperium:
The module's `power_distribution_publish_data_publish` function is called with measurements from `power_distribution_current_measurements`. It then calls a "transmitter" callback with all of the current measurements specified in the configuration. The transmitter in the standard front/rear configurations sends a CAN message with id `SYSTEM_CAN_MESSAGE_FRONT_CURRENT_MEASUREMENT` or `SYSTEM_CAN_MESSAGE_REAR_CURRENT_MEASUREMENT`.


projects/power_distribution/src/power_distribution_publish_data_config.c
```diff
@@ -0,0 +1,72 @@
...
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_3,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_4,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_5,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_6,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_7,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_8,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_9,
+          REAR_POWER_DISTRIBUTION_CURRENT_SPARE_10,
+      },
+  .num_currents_to_publish = 22,
```
g-jessmuir, line 71: is there a reason this and the earlier `22` aren't set with `#define`?

copperium: Those are the lengths of the array just before them, so a `#define` is kind of redundant and would be easier to forget to update

PR #65, opened 2020-05-11
SOFT[137] Steering Stalk

AveryChiu64:


projects/steering/src/steering_control_stalk.c
```diff
@@ -0,0 +1,71 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_digital_input.h"
+#include "steering_events.h"
+
+// Will be edited later
+#define STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE 1000
```
GeraldNDA, line 26: might wanna mention units somewhere for someone looking at this later. Also what's are these not the final voltages?

projects/steering/src/steering_control_stalk.c
```diff
@@ -0,0 +1,71 @@
...
+#define STEERING_CONTROL_STALK_RIGHT_SIGNAL_VOLTAGE 2000
+#define STEERING_CC_INCREASE_SPEED_VOLTAGE 3000
+#define STEERING_CC_DECREASE_SPEED_VOLTAGE 4000
+#define STEERING_CC_BRAKE_PRESSED_VOLTAGE 5000
+#define TOL 100
+
+// Function prototype
+void control_stalk_callback(uint16_t data, PeriodicReaderId id, void *context);
```
GeraldNDA, line 34: Should be in the header or private (`static` and `prv_...``)

projects/steering/src/steering_control_stalk.c
AveryChiu64: Yeah, I haven't got a chance to check the actual voltages from the control stalk

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,122 @@
...
+CanSettings can_settings = { .device_id = STEERING_CAN_DEVICE_ID,
+                             .bitrate = CAN_HW_BITRATE_500KBPS,
+                             .rx_event = STEERING_CAN_EVENT_RX,
+                             .tx_event = STEERING_CAN_EVENT_TX,
+                             .fault_event = STEERING_CAN_FAULT,
+                             .tx = { GPIO_PORT_A, 12 },
+                             .rx = { GPIO_PORT_A, 11 },
+                             .loopback = true };
+
+int count = 0;
```
GeraldNDA, line 42: might wanna prefix these

libraries/ms-helper/inc/adc_periodic_reader.h
```diff
@@ -0,0 +1,49 @@
...
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "gpio_mcu.h"
+#include "interrupt.h"
+#include "log.h"
+#include "misc.h"
+#include "soft_timer.h"
+#include "status.h"
```
GeraldNDA, line 12: Only include headers you use in this file. i.e. type definitions. Similar problem in .c files I think

libraries/ms-helper/src/x86/adc_periodic_reader.c
```diff
@@ -0,0 +1,82 @@
+#include "adc_periodic_reader.h"
```
GeraldNDA, line 1: is this any different than the stm version?

libraries/ms-helper/src/x86/adc_periodic_reader.c
```diff
@@ -0,0 +1,82 @@
...
+void prv_callback(SoftTimerId timer_id, void *context) {
+  for (size_t i = 0; i < NUM_PERIODIC_READER_IDS; i++) {
+    uint16_t data = 0;
+    AdcChannel channel;
+    adc_get_channel(s_storage[i].address, &channel);
+    adc_read_converted(channel, &data);
+    s_storage[i].data = data;
+    if (s_storage[i].activated) {
+      s_storage[i].callback(s_storage[i].data, i, s_storage[i].context);
+    }
```
GeraldNDA, line 26: Should probably not have read the data if this reader was not activated

libraries/ms-helper/src/x86/adc_periodic_reader.c
```diff
@@ -0,0 +1,82 @@
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "gpio_mcu.h"
+#include "interrupt.h"
+#include "log.h"
+#include "misc.h"
+#include "soft_timer.h"
+#include "status.h"
+
+#define TIMER_INTERVAL 50
```
GeraldNDA, line 13: This should probably be an init option

libraries/ms-helper/src/x86/adc_periodic_reader.c
```diff
@@ -0,0 +1,82 @@
...
+  soft_timer_start_millis(TIMER_INTERVAL, prv_callback, NULL, NULL);
+}
+
+StatusCode adc_periodic_reader_init() {
+  soft_timer_start_millis(TIMER_INTERVAL, prv_callback, NULL, NULL);
+
+  // Disable all ADCs
+  for (size_t i = 0; i < NUM_PERIODIC_READER_IDS; i++) {
+    s_storage[i].activated = false;
+  }
```
GeraldNDA, line 37: This is fine, but not really necessary ...

GeraldNDA: You can probably just do ` = {0};` at the top if you're worried about it

projects/steering/inc/steering_events.h
```diff
@@ -0,0 +1,21 @@
...
+  STEERING_HIGH_BEAM_FORWARD_EVENT,
+  STEERING_HIGH_BEAM_REAR_EVENT,
+  STEERING_REGEN_BRAKE_EVENT,
+  STEERING_INPUT_CC_TOGGLE_PRESSED_EVENT,
+  NUM_DIGITAL_STEERING_EVENTS,
+} SteeringDigitalEvent;
+
+typedef enum {
+  // Analog
+  STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL = 7,
```
GeraldNDA, line 16: Should probably be explicit i.e.
```suggestion
  STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL = NUM_DIGITAL_STEERING_EVENTS + 1,

```

projects/steering/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+    .rx = { GPIO_PORT_A, 11 },
+  };
+
+  can_init(&s_can_storage, &can_settings);
+
+  Event e = { .id = 0, .data = 0 };
+
+  while (true) {
+    while (event_process(&e)) {
+      steering_can_process_event(&e);
```
GeraldNDA, line 58: This probably doesn't work for 2 reasons:
1. there needs to be a can_process_event
2. There's no rx handler which generates the steering_can events

projects/steering/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+    .rx_event = STEERING_CAN_EVENT_RX,
+    .tx_event = STEERING_CAN_EVENT_TX,
+    .fault_event = STEERING_CAN_FAULT,
+    .tx = { GPIO_PORT_A, 12 },
+    .rx = { GPIO_PORT_A, 11 },
+  };
+
+  can_init(&s_can_storage, &can_settings);
+
+  Event e = { .id = 0, .data = 0 };
```
GeraldNDA, line 54: Can probably just be 
```suggestion
  Event e = { 0 };

```

projects/steering/src/steering_can.c
```diff
@@ -0,0 +1,48 @@
...
+    CAN_TRANSMIT_CRUISE_CONTROL_COMMAND(EE_CRUISE_CONTROL_COMMAND_TOGGLE);
+  } else if (e->id == STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL) {
+    CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_SIGNAL_LEFT, (EELightState)e->data);
+  } else if (e->id == STEERING_CONTROL_STALK_EVENT_RIGHT_SIGNAL) {
+    CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_SIGNAL_RIGHT, (EELightState)e->data);
+  } else if (e->id == STEERING_CC_EVENT_INCREASE_SPEED) {
+    CAN_TRANSMIT_CRUISE_CONTROL_COMMAND(EE_CRUISE_CONTROL_COMMAND_INCREASE);
+  } else if (e->id == STEERING_CC_EVENT_DECREASE_SPEED) {
+    CAN_TRANSMIT_CRUISE_CONTROL_COMMAND(EE_CRUISE_CONTROL_COMMAND_DECREASE);
+  }
```
GeraldNDA, line 46: dare I say that this is a good time to use a `switch/case` statement ... doesn't seem like a lookup table is possible here anyways

projects/steering/src/steering_control_stalk.c
```diff
@@ -0,0 +1,71 @@
...
+#include "steering_digital_input.h"
+#include "steering_events.h"
+
+// Will be edited later
+#define STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE 1000
+#define STEERING_CONTROL_STALK_RIGHT_SIGNAL_VOLTAGE 2000
+#define STEERING_CC_INCREASE_SPEED_VOLTAGE 3000
+#define STEERING_CC_DECREASE_SPEED_VOLTAGE 4000
+#define STEERING_CC_BRAKE_PRESSED_VOLTAGE 5000
+#define TOL 100
```
GeraldNDA, line 31: "VOLTAGE_TOLERANCE_MV" ? Naming is pretty unclear imo

projects/steering/src/steering_control_stalk.c
```diff
@@ -0,0 +1,71 @@
...
+
+// Function prototype
+void control_stalk_callback(uint16_t data, PeriodicReaderId id, void *context);
+
+// Needs to be edited for the actual stalk
+AdcPeriodicReaderSettings reader_settings = { .address = { .port = GPIO_PORT_A, .pin = 3 },
+                                              .callback = control_stalk_callback };
+
+// Stores event id of the event that was just raised
+SteeringAnalogEvent prev = NUM_STEERING_EVENTS;
```
GeraldNDA, line 41: Might wanna prefix these as well, goes for anywhere else that I didn't catch this too I guess

projects/steering/src/steering_control_stalk.c
AveryChiu64: Hmm I added this in the header already? However, it is only for testing.

GeraldNDA: Since you include that header, you shouldn't need the prototype here, I believe ... Also I feel like this should be private and then mock the adc read to return the values you want the periodic reader to get in the test.

libraries/ms-helper/src/x86/adc_periodic_reader.c
AveryChiu64: Yup, I need to seperate them since ADC doesent work on x86

libraries/ms-helper/src/x86/adc_periodic_reader.c
AveryChiu64: Hmm I could change it to be an init option, would it affect the functionality a lot? My tests rely on this exact interval of 50ms

libraries/ms-helper/src/x86/adc_periodic_reader.c
AveryChiu64: This will not read data if the reader is not activated?

libraries/ms-helper/src/x86/adc_periodic_reader.c
GeraldNDA: you can make it a constant in the test that you pass to the periodic reader. In the case of steering tests it should be a "public" define that the tests have access to and use I guess. I don't think it should be a big deal to change, but would make things cleaner if they end up being changed later

projects/steering/test/test_control_stalk.c
```diff
@@ -0,0 +1,114 @@
...
+
+#define STEERING_CAN_DEVICE_ID 0x1
+#define STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE 1000
+#define STEERING_CONTROL_STALK_RIGHT_SIGNAL_VOLTAGE 2000
+#define STEERING_CC_INCREASE_SPEED_VOLTAGE 3000
+#define STEERING_CC_DECREASE_SPEED_VOLTAGE 4000
+#define STEERING_CC_BRAKE_PRESSED_VOLTAGE 5000
+#define INVALID_VOLTAGE 6000
+#define VOLTAGE_TOLERANCE_MV 100
+#define TIMER_INTERVAL_MS 50
```
g-jessmuir, line 27: could you make this any lower for the test? we run all test cases whenever we commit to a PR, so it would save time, even if it's just a little

projects/steering/test/test_control_stalk.c
```diff
@@ -0,0 +1,114 @@
...
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_control_stalk.h"
+#include "steering_digital_input.h"
+#include "steering_events.h"
+#include "test_helpers.h"
+
+#define STEERING_CAN_DEVICE_ID 0x1
+#define STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE 1000
```
g-jessmuir, line 20: these should be defined in a header, so you don't have to redefine them here

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,115 @@
...
+#include "interrupt_def.h"
+#include "ms_test_helpers.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_digital_input.h"
+#include "steering_events.h"
+#include "test_helpers.h"
+
+#define STEERING_CAN_DEVICE_ID 0x1
```
g-jessmuir, line 16: should be defined in a header somewhere

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,115 @@
...
+
+typedef enum {
```
g-jessmuir, line 18: again, these should be defined in the steering_events.h header

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,115 @@
...
+
+StatusCode prv_test_cc_toggle_rx_cb_handler(const CanMessage *msg, void *context,
+                                            CanAckStatus *ack_reply) {
+  TEST_ASSERT_EQUAL(SYSTEM_CAN_MESSAGE_CRUISE_CONTROL_COMMAND, msg->msg_id);
+  count++;
+  return STATUS_CODE_OK;
+}
+
+void setup_test(void) {
+  TEST_ASSERT_OK(gpio_init());
```
g-jessmuir, line 59: no TEST_ASSERT_OK necessary for libraries as mentioned previously

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,115 @@
...
+  gpio_it_init();
+  soft_timer_init();
+  TEST_ASSERT_OK(steering_digital_input_init());
+  TEST_ASSERT_OK(can_init(&s_can_storage, &can_settings));
+}
+
+void test_steering_digital_input_horn() {
+  TEST_ASSERT_OK(
+      can_register_rx_handler(SYSTEM_CAN_MESSAGE_HORN, prv_test_horn_rx_cb_handler, NULL));
+  GpioAddress *horn_address = test_get_address(STEERING_INPUT_HORN_EVENT);
```
g-jessmuir, line 71: again, these can be #defined in the header so you don't have to have this function, goes for other instances in this file as well

projects/steering/test/test_steering_digital_input.c
```diff
@@ -0,0 +1,115 @@
...
+                                   (uint16_t)GPIO_STATE_LOW);
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+  TEST_ASSERT_OK(steering_can_process_event(&e));
+  MS_TEST_HELPER_CAN_TX_RX(STEERING_CAN_EVENT_TX, STEERING_CAN_EVENT_RX);
+  TEST_ASSERT_EQUAL(3, count);
+}
+
+void test_invalid_can_message() {
+  // provide an invalid id
+  Event e = { .id = 16, .data = 0 };
```
g-jessmuir, line 111: nitpick: #define INVALID_STEERING_CAN_ID could be used for clarity

projects/steering/inc/steering_events.h
```diff
@@ -0,0 +1,21 @@
...
+  STEERING_CONTROL_STALK_EVENT_RIGHT_SIGNAL,
+  STEERING_CC_EVENT_INCREASE_SPEED,
+  STEERING_CC_EVENT_DECREASE_SPEED,
+  NUM_STEERING_EVENTS
```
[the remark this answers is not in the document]
AveryChiu64: Naming it NUM_ANALOG_STEERING_EVENTS would be incorrect since the enum starts at 7, not 0

projects/steering/src/steering_digital_input.c
```diff
@@ -0,0 +1,65 @@
...
+EventId s_steering_event_lookup_table[NUM_STEERING_DIGITAL_INPUTS] = {
+  [STEERING_DIGITAL_INPUT_HORN] = STEERING_INPUT_HORN_EVENT,
+  [STEERING_DIGITAL_INPUT_RADIO_PPT] = STEERING_RADIO_PPT_EVENT,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_FORWARD] = STEERING_HIGH_BEAM_FORWARD_EVENT,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_REAR] = STEERING_HIGH_BEAM_REAR_EVENT,
+  [STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE] = STEERING_REGEN_BRAKE_EVENT,
+  [STEERING_DIGITAL_INPUT_CC_TOGGLE] = STEERING_INPUT_CC_TOGGLE_PRESSED_EVENT
+};
+
+GpioAddress *test_get_address(int digital_input_id) {
```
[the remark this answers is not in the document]
AveryChiu64: If I did this the implementation would be messed up since I would need to manually set up interrupts for each one. Also, if I remember correctly Arshan told me to put the lookup table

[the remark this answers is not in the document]
g-jessmuir: Couldn't you still have the lookup table, just also have the #defines? this way you can keep your implementation but not have the helper function

projects/steering/inc/steering_events.h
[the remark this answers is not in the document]
g-jessmuir: yeah this is true actually, I think I just haven't seen events used this way. should be fine, maybe leave a comment.

projects/steering/src/steering_digital_input.c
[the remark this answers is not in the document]
AveryChiu64: Also I tried this and its not working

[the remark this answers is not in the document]
g-jessmuir: hmm it's used this way in power distro inc/power_distribution_pin_defs.h: 
```
#define POWER_DISTRIBUTION_I2C_SCL_PIN \
  { GPIO_PORT_B, 10 }
```
maybe try that?

projects/steering/src/main.c
```diff
@@ -0,0 +1,50 @@
...
+#include "gpio_it.h"
+#include "gpio_mcu.h"
+#include "interrupt.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_control_stalk.h"
+#include "steering_digital_input.h"
+#include "steering_events.h"
+#define STEERING_CAN_DEVICE_ID 0x1
```
g-jessmuir, line 13: this should be included in `can_msg_defs.h`, include that and use the value `SYSTEM_CAN_DEVICE_STEERING`. You'll probably also have to add codegen-tooling to the makefile, check the mci project makefile for details I think. Sorry, I think I missed this on first review.

projects/steering/src/steering_control_stalk.c
```diff
@@ -0,0 +1,50 @@
...
+                                              .callback = control_stalk_callback };
+
+// Stores event id of the event that was just raised
+static SteeringAnalogEvent prev = NUM_STEERING_EVENTS;
+
+void control_stalk_callback(uint16_t data, PeriodicReaderId id, void *context) {
+  if (data > STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE - VOLTAGE_TOLERANCE_MV &&
+      data < STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE + VOLTAGE_TOLERANCE_MV &&
+      prev != STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL) {
+    event_raise((EventId)STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL, data);
```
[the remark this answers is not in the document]
g-jessmuir: ^^Did you try this?

projects/steering/src/steering_digital_input.c
```diff
@@ -0,0 +1,75 @@
+#include "steering_digital_input.h"
+#include "event_queue.h"
+#include "gpio_it.h"
+#include "interrupt_def.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_events.h"
+
+GpioAddress s_steering_address_lookup_table[NUM_STEERING_DIGITAL_INPUTS] = {
```
g-jessmuir, line 10: this should be static

projects/steering/src/steering_digital_input.c
```diff
@@ -0,0 +1,75 @@
...
+  [STEERING_DIGITAL_INPUT_RADIO_PPT] = RADIO_PPT_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_FORWARD] = HIGH_BEAM_FORWARD_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_REAR] = HIGH_BEAM_REAR_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE] = REGEN_BRAKE_TOGGLE_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_CC_TOGGLE] = CC_TOGGLE_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_CC_INCREASE_SPEED] = CC_INCREASE_SPEED_GPIO_ADDR,
+  [STEERING_DIGITAL_INPUT_CC_DECREASE_SPEED] = CC_INCREASE_SPEED_GPIO_ADDR,
+};
+
+EventId s_steering_event_lookup_table[NUM_STEERING_DIGITAL_INPUTS] = {
```
g-jessmuir, line 21: this should also be static

projects/steering/src/steering_digital_input.c
```diff
@@ -0,0 +1,75 @@
...
+  [STEERING_DIGITAL_INPUT_RADIO_PPT] = STEERING_RADIO_PPT_EVENT,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_FORWARD] = STEERING_HIGH_BEAM_FORWARD_EVENT,
+  [STEERING_DIGITAL_INPUT_HIGH_BEAM_REAR] = STEERING_HIGH_BEAM_REAR_EVENT,
+  [STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE] = STEERING_REGEN_BRAKE_EVENT,
+  [STEERING_DIGITAL_INPUT_CC_TOGGLE] = STEERING_DIGITAL_INPUT_CC_TOGGLE_PRESSED_EVENT,
+  [STEERING_DIGITAL_INPUT_CC_INCREASE_SPEED] = STEERING_CC_INCREASE_SPEED_EVENT,
+  [STEERING_DIGITAL_INPUT_CC_DECREASE_SPEED] = STEERING_CC_DECREASE_SPEED_EVENT,
+};
+
+GpioAddress *get_address(int digital_input_id) {
```
g-jessmuir, line 32: is this used anywhere? you can remove it right?

projects/steering/src/steering_digital_input.c
```diff
@@ -0,0 +1,75 @@
...
+  return &s_steering_address_lookup_table[digital_input_id];
+}
+
+EventId *get_event(int digital_input_id) {
```
g-jessmuir, line 36: I don't see this one used anywhere either

projects/steering/inc/steering_control_stalk.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+// Module to set up analog readers to read voltages and convert
+// them to signal events
+#include "adc_periodic_reader.h"
+#include "event_queue.h"
+#include "status.h"
+
+#define VOLTAGE_TOLERANCE_MV 100
+#define STEERING_CONTROL_STALK_LEFT_SIGNAL_VOLTAGE 1000
+#define STEERING_CONTROL_STALK_RIGHT_SIGNAL_VOLTAGE 2000
```
g-jessmuir, line 10: should this be `_MV` for millivolts? I think that's what our adc reads return

libraries/ms-helper/rules.mk
```diff
@@ -9,7 +9,7 @@
 $(T)_DEPS := ms-common
 
 ifeq (x86,$(PLATFORM))
-$(T)_EXCLUDE_TESTS := mcp2515
```
GeraldNDA, line 12: Now that x86 has an adc implementation, you shouldn't need this?

g-jessmuir: Avery was hitting a lot of issues getting travis to run the test successfully, so I made the call to just skip it on x86. I think it's alright, hopefully doesn't cause issues down the line.

PR #70, opened 2020-05-23
Soft 169 pca9539r smoke test

mitchellostler:
Created smoke test for pca9539r. May need to add input testing in the future


projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+#include "i2c.h"
+#include "interrupt.h"
+#include "log.h"
+#include "pca9539r_gpio_expander.h"
+#include "soft_timer.h"
+#include "test_helpers.h"
+#include "wait.h"
+
+// this used to set test mode. If set to 1, will toggle gpio states for outputs
+#define TEST_MODE_OUTPUT 1
```
copperium, line 16: You don't seem to ever use this. What's "test mode"?

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
+/* Smoke test for pca9539r gpio expander. i2c and other settings based off of
+   test_pca9539r_gpio_expander.c Program sets all pins as inputs, and low state, then outputs with
+   high state, and checks configuration registers against expected results Then toggles gpio state
+   on all pins every second */
```
copperium, line 4: Generally we prefer a series of single-line `//` comments rather than block comments. Also, could you separate your sentences with periods?

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+
+#include "gpio.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "log.h"
+#include "pca9539r_gpio_expander.h"
+#include "soft_timer.h"
+#include "test_helpers.h"
```
copperium, line 12: `test_helpers.h` shouldn't be included outside of a test.

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+// this used to set test mode. If set to 1, will toggle gpio states for outputs
+#define TEST_MODE_OUTPUT 1
+
+// these i2c settings may need adjustment - taken from test_pca9539r_gpio_expander.c
+#define TEST_I2C_PORT I2C_PORT_2
+#define TEST_CONFIG_PIN_I2C_SCL \
+  { GPIO_PORT_B, 10 }
+#define TEST_CONFIG_PIN_I2C_SDA \
+  { GPIO_PORT_B, 11 }
+#define VALID_I2C_ADDRESS 0x74
```
copperium, line 24: Please rename these to something like `I2C_PORT`, `I2C_SCL_PIN`, `I2C_SDA_PIN`, and `I2C_ADDRESS` respectively and move the address beside the port. You should then give instructions to the hardware person using this on how to use these parameters.
* `I2C_ADDRESS` is the I2C address of the PCA9539R
* `I2C_PORT` is which I2C port to use (we have two; `I2C_PORT_1` has SDA on PB9 and SCL on PB8, `I2C_PORT_2` has SDA on PB11 and SCL on PB10)
* `I2C_SDA_PIN` and `I2C_SCL_PIN` are the GPIO addresses of the SDA (data) and SCL (clock) I2C pins.

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+    if (get_state == *state) {
+      LOG_DEBUG("State for pin %d_%d set and read correctly\n", address.pin / 8, address.pin % 8);
+    } else {
+      LOG_DEBUG("State for pin %d_%d incorrect\n", address.pin / 8, address.pin % 8);
+    }
+  }
+  *state =
+      (*state == PCA9539R_GPIO_STATE_HIGH ? PCA9539R_GPIO_STATE_LOW : PCA9539R_GPIO_STATE_HIGH);
+  LOG_DEBUG("GPIO state = %d\n", *state);
+  soft_timer_start_millis(1000, prv_soft_timer_callback_output, (void *)state, NULL);
```
copperium, line 95: Nit: I don't think you have to cast to `void *` here

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/* TODO: look at triggering interrupt on input*/
```
copperium, line 98: What do you mean by this? Also, replace this with a single-line comment, though `make lint` will complain about the lack of a Jira ticket in the todo; you can just use `// TODO(SOFT-169): ...`

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+  LOG_DEBUG("Initializing all pins in...\n");
+  pca9539r_init_all_pins(PCA9539R_GPIO_DIR_IN);
+  pca9539r_check_all_pin_states(PCA9539R_GPIO_STATE_LOW);
+  LOG_DEBUG("Initializing all pins out...\n");
+  pca9539r_init_all_pins(PCA9539R_GPIO_DIR_OUT);
+  pca9539r_check_all_pin_states(PCA9539R_GPIO_STATE_HIGH);
+  LOG_DEBUG("GPIO initialization complete. Now beginning toggling of GPIO states\n");
+  Pca9539rGpioState state;
+  // Toggles gpio, compares expected values against registers, first read should be ignored
+  soft_timer_start_millis(100, prv_soft_timer_callback_output, (void *)&state, NULL);
```
copperium, line 116: Nit again: unnecessary cast to `void *`

projects/smoke_pca9539r/src/main.c
g-jessmuir: re: 'generally we prefer', it's actually in our (not often looked at) coding standards: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards

projects/smoke_pca9539r/src/main.c
g-jessmuir: this is a good point, I've added a note about including instructions on the confluence page for smoke tests. https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1263927334/Smoke+Tests

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+  { GPIO_PORT_B, 10 }
+#define TEST_CONFIG_PIN_I2C_SDA \
+  { GPIO_PORT_B, 11 }
+#define VALID_I2C_ADDRESS 0x74
+
+void setup_test(void) {
+  I2CSettings i2c_settings = {
+    .speed = I2C_SPEED_FAST,         //
+    .sda = TEST_CONFIG_PIN_I2C_SDA,  //
+    .scl = TEST_CONFIG_PIN_I2C_SCL,  //
```
g-jessmuir, line 30: remove unnecessary comments

projects/smoke_pca9539r/src/main.c
g-jessmuir: (outside of unit tests)

projects/smoke_pca9539r/src/main.c
g-jessmuir: also, maybe put the 1000 as a #define since it may be useful to change the frequency

projects/smoke_pca9539r/src/main.c
g-jessmuir: @copperium do we use the PCA9539R for input at all, or is it just for output? if it's just for output, we won't smoketest for input and we can just remove the todo.

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,123 @@
...
+int main() {
+  interrupt_init();
+  soft_timer_init();
+  setup_test();
+
+  Pca9539rGpioState test_state;
+  LOG_DEBUG("Testing GPIO initialization...\n");
+  LOG_DEBUG("Initializing all pins in...\n");
+  pca9539r_init_all_pins(PCA9539R_GPIO_DIR_IN);
+  pca9539r_check_all_pin_states(PCA9539R_GPIO_STATE_LOW);
```
g-jessmuir, line 109: re: todo / checking if we use the chip for input, if we don't I think we can remove this check and init

projects/smoke_pca9539r/src/main.c
copperium: @g-jessmuir Good point, power distribution never uses the PCA9539R for input, so the todo can be removed.

projects/smoke_pca9539r/src/main.c
copperium: Interesting, searching `projects/*` for `//` gives 756 results, whereas searching for `*/` gives 15.

mitchellostler: Updated instructions, made some nice clean one line comments

projects/smoke_pca9539r/src/main.c
mitchellostler: Commented in both possible configurations so that they can be easily changed

projects/smoke_pca9539r/src/main.c
mitchellostler: added option to define wait time, and removed all void* casts

projects/smoke_pca9539r/src/main.c
mitchellostler: Ok, won't worry about it

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,120 @@
...
+    .speed = I2C_SPEED_FAST,
+    .sda = PIN_I2C_SDA,
+    .scl = PIN_I2C_SCL,
+  };
+  i2c_init(I2C_PORT, &i2c_settings);
+  pca9539r_gpio_init(I2C_PORT, PCA9539_I2C_ADDRESS);
+}
+
+// initialize all pins to in/out - must be called after pca9539r_gpio_init
+StatusCode pca9539r_init_all_pins(Pca9539rGpioDirection direction) {
```
copperium, line 39: A final nit I didn't catch earlier: this should be `static` and called `prv_pca9539r_init_all_pins`

projects/smoke_pca9539r/src/main.c
```diff
@@ -0,0 +1,120 @@
...
+  }
+  for (Pca9539rPinAddress pin = PCA9539R_PIN_IO0_0; pin < NUM_PCA9539R_GPIO_PINS; pin++) {
+    address.pin = pin;
+    status_ok_or_return(pca9539r_gpio_init_pin(&address, &gpio_settings));
+  }
+  return STATUS_CODE_OK;
+}
+
+// reads back configuration registers and compares them to input state
+StatusCode pca9539r_check_all_pin_states(Pca9539rGpioState state) {
```
copperium, line 60: This too should be `static` and called `prv_pca9539r_check_all_pin_states`

mitchellostler: For sure, will make the change now. Still learning on the code front, so will keep this convo as reference and the next one will be a lot shorter haha

mitchellostler: Thanks for the review!

PR #71, opened 2020-05-25
[SOFT-185] implement x86 adc.c

g-jessmuir:
Most code copied from the STM version.

### DIFF
--- a/libraries/ms-common/src/x86/adc.c
+++ b/libraries/ms-common/src/x86/adc.c
@@ -7,18 +7,12 @@
 // adc_read_raw should always return 4090.
 // Vdda locked at 3300 mV.
 // adc_read_converted should always return close to 2V
+// temperature reading always returns 293 kelvin.
 
 #define ADC_RETURNED_VOLTAGE_RAW 2500
 #define ADC_CONTINUOUS_CB_FREQ_MS 50
-
-// TS_CAL addresses obtained from section 3.10.1 of the specific device
-// datasheet
-#define ADC_TS_CAL1 0x1FFFF7b8
-#define ADC_TS_CAL2 0x1FFFF7c2
-
-// ADC_VREFINT_CAL address obtained from section 3.10.2 of the specific device
-// datasheet
-#define ADC_VREFINT_CAL 0x1FFFF7ba
+#define ADC_TEMP_RETURN 293
+#define ADC_VDDA_RETURN 3300
 
 typedef struct AdcInterrupt {
   AdcCallback callback;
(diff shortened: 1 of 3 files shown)


g-jessmuir:
@GeraldNDA 
> LGTM I think. Do the tests not pass already, I don't see why they shouldn't ...

We previously didn't run adc tests on x86, hence the makefile change. They wouldn't have passed because everything just returned `STATUS_CODE_UNIMPLEMENTED`, but they pass now.

> One thing to consider is that the adc_callback isn't used by anyone. I didn't realize this before but maybe we should modify the interface so that the callback can modify the reading?

Hmm good point. Do you have any recommendations on how to do this without needing significant refactoring? To me registering a callback to modify the reading isn't any more or less work than mocking `adc_read_raw` or `adc_read_converted` to return a custom value.

GeraldNDA:
RE:tests; My b, read the changlelog backwards 😅

RE:callback; Current callback returns nothing and takes in the channel and some context. I think ideally it should also either take in the reading value by reference (and status_ok return maybe) or take in a reading value and return the modified reading value.

libraries/ms-common/src/x86/adc.c
```diff
@@ -1,23 +1,172 @@
...
+
+static AdcInterrupt s_adc_interrupts[NUM_ADC_CHANNELS];
+
+static bool s_active_channels[NUM_ADC_CHANNELS];
+
+// Formula obtained from section 13.9 of the reference manual. Returns reading
+// in kelvin
+static uint16_t prv_get_temp(uint16_t reading) {
+  uint16_t ts_cal1 = *(uint16_t *)ADC_TS_CAL1;
+  uint16_t ts_cal2 = *(uint16_t *)ADC_TS_CAL2;
```
copperium, line 37: You seem to be dereferencing a hardcoded pointer on x86? This will fail with a segfault, the `ts_cal` values themselves should probably be hardcoded like Vdda. I'm surprised there are no tests for temp/vdda.

libraries/ms-common/src/x86/adc.c
```diff
@@ -1,23 +1,172 @@
...
+#define ADC_CONTINUOUS_CB_FREQ_MS 50
+
+// TS_CAL addresses obtained from section 3.10.1 of the specific device
+// datasheet
+#define ADC_TS_CAL1 0x1FFFF7b8
+#define ADC_TS_CAL2 0x1FFFF7c2
+
+// ADC_VREFINT_CAL address obtained from section 3.10.2 of the specific device
+// datasheet
+#define ADC_VREFINT_CAL 0x1FFFF7ba
```
copperium, line 21: This is never used (it's an artifact from the stm32 version of `prv_get_vdda`), should be removed.

libraries/ms-common/src/x86/adc.c
```diff
@@ -1,23 +1,172 @@
...
+// Vdda locked at 3300 mV.
+// adc_read_converted should always return close to 2V
+
+#define ADC_RETURNED_VOLTAGE_RAW 2500
+#define ADC_CONTINUOUS_CB_FREQ_MS 50
+
+// TS_CAL addresses obtained from section 3.10.1 of the specific device
+// datasheet
+#define ADC_TS_CAL1 0x1FFFF7b8
+#define ADC_TS_CAL2 0x1FFFF7c2
```
copperium, line 17: Like I commented below, these hardcoded pointers won't work on x86. Should be removed, I'd replaced them with hardcoded `TS_CAL` values.

g-jessmuir:
@GeraldNDA 
> RE:callback; Current callback returns nothing and takes in the channel and some context. I think ideally it should also either take in the reading value by reference (and status_ok return maybe) or take in a reading value and return the modified reading value.

Hmm, not sure I'm convinced on the need for a return value, but I edited the x86 implementation of `adc_read_raw` to allow you to mock the value with a callback. I also provided a test case to demonstrate how you'd do this. I'm kinda just too lazy to rewrite all the callback usage in the tests, do you think this is ok?

g-jessmuir:
> Seems good, just the one issue the tests didn't catch with `prv_get_temp`.

Wow good catch, I was definitely moving too fast copying and pasting code here. Also added a testcase for reading temperature.

GeraldNDA:
If you passed in the reading by reference in the new callback then you would just have to change the function signature. That being said reordering it like you did should be sufficient. Only "concern" is the difference between how callbacks work on stm and x86. But I think it should be fine. So up to you.

PR #72, opened 2020-05-26
Firmware 102: A & B counter

tristanle22:
counterAB

### DIFF
--- a/projects/counter/inc/counter.h
+++ b/projects/counter/inc/counter.h
@@ -1,0 +1,10 @@
+#pragma once
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+// Callback function for counter 1
+static void counter_a_timeout(SoftTimerId timer_id, void *context);
+
+// Callback function for counter 2
+static void counter_b_timeout(SoftTimerId timer_id, void *context);
--- a/projects/counter/inc/counter.h
+++ b/projects/counter/inc/counter.h
@@ -4,7 +4,7 @@
 #include "soft_timer.h"
 
 // Callback function for counter 1
-static void counter_a_timeout(SoftTimerId timer_id, void *context);
+void counter_a_timeout(SoftTimerId timer_id, void *context);
 
 // Callback function for counter 2
-static void counter_b_timeout(SoftTimerId timer_id, void *context);
+void counter_b_timeout(SoftTimerId timer_id, void *context);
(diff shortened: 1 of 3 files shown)


projects/counter/src/main.c
```diff
@@ -0,0 +1,47 @@
...
+  LOG_DEBUG("Counter B: %d\n", counter->count_b);
+
+  soft_timer_start_millis(COUNTER_TIMEOUT_MILI_SECS, counter_a_timeout, counter, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counter = {
```
g-jessmuir, line 36: please initialize this to zero with `Counters counter = { 0 }` to match convention

projects/counter/src/main.c
```diff
@@ -0,0 +1,47 @@
+#include "counter.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_TIMEOUT_MILI_SECS 500
```
g-jessmuir, line 7: please just suffix the #define `_MS` to denote milli-seconds, to match with convention

projects/counter/inc/counter.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+// Callback function for counter 1
+static void counter_a_timeout(SoftTimerId timer_id, void *context);
```
g-jessmuir, line 7: in general we'd consider these functions private and thus would not include them in the header. private functions are prefixed `prv_`, but the specs were not 100% clear on using only one file so this is ok. Functions in the header should not be declared static.

projects/counter/src/main.c
```diff
@@ -0,0 +1,47 @@
...
+  uint8_t count_a;
+  uint8_t count_b;
+} Counters;
+
+static void counter_a_timeout(SoftTimerId timer_id, void *context) {
+  Counters *counter = context;
+  counter->count_a++;
+  LOG_DEBUG("Counter A: %d\n", counter->count_a);
+
+  soft_timer_start_millis(COUNTER_TIMEOUT_MILI_SECS, counter_b_timeout, counter, NULL);
```
g-jessmuir, line 19: in the specs it mentioned using only one soft timer 🤔🤔🤔 I'll give you a pass here because it works though.

tristanle22: oh oops :sweat_smile:

g-jessmuir:
complete!

PR #75, opened 2020-05-27
Soft 999 james carr pries fw 102

jcarrpries:


projects/hello_world/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counters;
+  counters.counter_a = 0;
+  counters.counter_b = 0;
+  counters.half_second = false;
+
+  soft_timer_start_millis(500, callback, &counters, NULL);
```
g-jessmuir, line 40: please use a `#define ... 500` for the 500 milliseconds with a descriptive name

projects/hello_world/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+#include "soft_timer.h"  // Software timers for scheduling future events.
+#include "wait.h"        // for the wait function
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+  bool half_second;
+} Counters;
+
+void callback(SoftTimerId timer_id, void *context) {
```
g-jessmuir, line 13: please use `static void`, and prefix the function name with `prv_` to denote the function is private to this file/module. Could also be made more descriptive by using `prv_timer_callback`.

projects/hello_world/src/main.c
```diff
@@ -0,0 +1,46 @@
+#include "delay.h"  // For real-time delays
```
g-jessmuir, line 1: please remove unnecessary include

projects/hello_world/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+  counters->half_second = !counters->half_second;
+
+  soft_timer_start_millis(500, callback, counters, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counters;
```
g-jessmuir, line 35: we usually initialize an empty struct with `= { 0 };`, please use this for clarity. For reference, doing this initializes the memory of the struct to zero.

projects/hello_world/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+  counters.counter_a = 0;
+  counters.counter_b = 0;
+  counters.half_second = false;
```
g-jessmuir, line 38: With the above comment, you won't need to explicitly set these values here.

projects/hello_world/src/main.c
jcarrpries: will this work for the boolean being set to false? I understand why it would work for just numerical values.

g-jessmuir: yes, a false boolean in C is equivalent to 0

g-jessmuir:
LGTM

PR #76, opened 2020-05-27
Soft 191 pre commit hook

AveryChiu64:


.travis.yml
```diff
@@ -97,12 +97,14 @@ before_script:
   - pylint --version
 
 script:
+  - make test_format || travis_terminate 1;
+  - make lint || travis_terminate 1;
+  - make pylint || travis_terminate 1;
   - make build_all PLATFORM=x86
+  - make clean
   - make build_all PLATFORM=stm32f0xx
+  - make clean
```
g-jessmuir, line 106: how about we reorder this to be
```
make build_all PLATFORM=stm32f0xx
make clean 
make build_all PLATFORM=x86
make test_all PLATFORM=x86
```
that way we can avoid a build

PR #77, opened 2020-05-30
[SOFT-188] SPV1020 MPPT driver

copperium:
Supports issuing each command to the SPV1020s and extracting data from the status byte.

Did you know you can have both x86/stm32-specific C files and a common one?

### DIFF
--- a/libraries/ms-drivers/inc/spv1020_mppt.h
+++ b/libraries/ms-drivers/inc/spv1020_mppt.h
@@ -1,0 +1,47 @@
+#pragma once
+
+// Driver for the SPV1020 maximum power point tracker.
+// Requires GPIO and SPI to be initialized.
+
+// This driver only issues commands over SPI and does not select between MPPTs. To use this driver
+// with multiple SPV1020s, make a wrapper.
+
+// Note that this driver does not attempt to convert data from the SPV1020's internal ADC values.
+// The reference voltage is 1.25V.
+
+#include <stdint.h>
+#include <stdbool.h>
+#include "status.h"
+
+// Issue a SHUT command to the SPV1020. This shuts down the SPV1020.
+StatusCode spv1020_shut(void);
+
+// Issue a Turn ON command to the SPV1020. This turns on the SPV1020 after a previous SHUT command.
+// It is only required after a SHUT command as the SPV1020s turn on automatically on power up.
+StatusCode spv1020_turn_on(void);
+
+// Read current through the SPV1020. Current is 10 bits, converted by the SPV1020's internal ADC.
+StatusCode spv1020_read_current(uint16_t *current);
+
+// Read the SPV1020's input voltage. Voltage is 10 bits, converted by the SPV1020's internal ADC.
+StatusCode spv1020_read_voltage_in(uint16_t *vin);
+
+// Read PWM from the SPV1020. This is 9 bits.
+StatusCode spv1020_read_pwm(uint16_t *pwm);
+
+// Read a status byte from the SPV1020. The status byte should then be passed to
+// |spv1020_is_overcurrent|, |spv_1020_is_overvoltage|, and |spv1020_is_overtemperature| to check
+// each flag in the status byte.
+StatusCode spv1020_read_status(uint8_t *status);
+
+// Return whether an overcurrent (OVC) bit is set in the status byte. If an OVC bit is set,
+// |ovc_branch| will be set to 1, 2, 3, or 4, indicating on which branch the overcurrent took place;
+// 1 indicates the least significant OVC bit and 4 is the most significant. Otherwise, |ovc_branch|
+// is set to 0.
+bool spv1020_is_overcurrent(uint8_t status, uint8_t *ovc_branch);
+
+// Return whether the overvoltage (OVV) bit is set in the status byte.
+bool spv1020_is_overvoltage(uint8_t status);
+
+// Return whether the overtemperature (OVT) bit is set in the status byte.
+bool spv1020_is_overtemperature(uint8_t status);
--- a/libraries/ms-drivers/inc/spv1020_mppt.h
+++ b/libraries/ms-drivers/inc/spv1020_mppt.h
@@ -21,12 +21,15 @@ StatusCode spv1020_shut(SpiPort port);
 StatusCode spv1020_turn_on(SpiPort port);
 
 // Read current through the SPV1020. Current is 10 bits, converted by the SPV1020's internal ADC.
+// This is either the input current or the average phase current (input current / 4 / duty cycle).
 StatusCode spv1020_read_current(SpiPort port, uint16_t *current);
 
 // Read the SPV1020's input voltage. Voltage is 10 bits, converted by the SPV1020's internal ADC.
 StatusCode spv1020_read_voltage_in(SpiPort port, uint16_t *vin);
 
-// Read PWM from the SPV1020. This is 9 bits.
+// Read the PWM duty cycle from the SPV1020. This is 9 bits. Note that as per the application note,
+// the PWM duty cycle can vary from 5% to 90% with a step of 0.2%, giving 425 possible values, so
+// this likely gives the "step number" of the duty cycle (duty cycle = 5% + 85% * (pwm/425)).
 StatusCode spv1020_read_pwm(SpiPort port, uint16_t *pwm);
 
 // Read a status byte from the SPV1020. The status byte should then be passed to
@@ -43,3 +46,7 @@ bool spv1020_is_overvoltage(uint8_t status);
 
 // Return whether the overtemperature (OVT) bit is set in the status byte.
 bool spv1020_is_overtemperature(uint8_t status);
+
+// Return whether the CR bit is set in the status byte. We don't know what the CR bit is, but it
+// might be useful to log it.
+bool spv1020_is_cr_bit_set(uint8_t status);
--- a/libraries/ms-drivers/inc/spv1020_mppt.h
+++ b/libraries/ms-drivers/inc/spv1020_mppt.h
@@ -3,42 +3,40 @@
 // Driver for the SPV1020 maximum power point tracker.
 // Requires GPIO and SPI to be initialized.
 
-// This driver only issues commands over SPI and does not select between MPPTs. To use this driver
-// with multiple SPV1020s, make a wrapper.
+// This driver only issues commands the given SPI port and does not select between MPPTs. To use
+// this driver with multiple SPV1020s, make a wrapper.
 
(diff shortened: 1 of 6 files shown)


libraries/ms-drivers/src/spv1020_mppt.c
```diff
@@ -0,0 +1,25 @@
+#include "spv1020_mppt.h"
+#include "spv1020_mppt_defs.h"
+
+// This file implements common functionality between stm32 and x86. See stm32f0xx/spv1020_mppt.c
```
AveryChiu64, line 4: Hmm, why is this file needed if it is neither for x86 or stm32?

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
```diff
@@ -0,0 +1,45 @@
...
+
+// Send a command to SPV1020 and get a response. |rx_len| must be <= 2.
+// It's safe to pass NULL as |rx_data| if |rx_len| is 0.
+static StatusCode prv_send_command(SpiPort port, uint8_t command, uint8_t *rx_data, size_t rx_len) {
+  if (port >= NUM_SPI_PORTS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // We don't use spi_exchange since that assumes 0x00 is NOP, whereas SPV1020 appears to use 0x01.
+  spi_cs_set_state(port, GPIO_STATE_LOW);
```
AveryChiu64, line 14: Why is the GPIO_STATE_LOW set twice for the same port?

libraries/ms-drivers/test/test_spv1020_mppt.c
```diff
@@ -0,0 +1,136 @@
...
+  spi_init(TEST_SPI_PORT, &spi_settings);
+}
+void teardown_test() {}
+
+// Test that we can successfully send the SHUT and Turn ON commands.
+void test_shut_and_turn_on(void) {
+  TEST_ASSERT_OK(spv1020_shut(TEST_SPI_PORT));
+  TEST_ASSERT_OK(spv1020_turn_on(TEST_SPI_PORT));
+  TEST_ASSERT_OK(spv1020_shut(TEST_SPI_PORT));
+  TEST_ASSERT_OK(spv1020_shut(TEST_SPI_PORT));  // 2 in a row
```
AveryChiu64, line 38: Wait if it is already shut why are they allowed to keep sending this command?

libraries/ms-drivers/inc/spv1020_mppt.h
```diff
@@ -0,0 +1,52 @@
...
+// be set to a 4-bit bitmask indicating which, if any, branches have overcurrent.
+bool spv1020_is_overcurrent(uint8_t status, uint8_t *ovc_branches);
+
+// Return whether the overvoltage (OVV) bit is set in the status byte.
+bool spv1020_is_overvoltage(uint8_t status);
+
+// Return whether the overtemperature (OVT) bit is set in the status byte.
+bool spv1020_is_overtemperature(uint8_t status);
+
+// Return whether the CR bit is set in the status byte. We don't know what the CR bit is, but it
```
AveryChiu64, line 50: Wait what does CR stand for?

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
copperium: Whoops, fixed that

libraries/ms-drivers/src/spv1020_mppt.c
copperium: That file has common functionality, the `spv1020_is_X` functions are the same between x86 and stm32.

libraries/ms-drivers/inc/spv1020_mppt.h
copperium: It shows up in the datasheet, I don't know what it stands for and neither does Micah, but it might be good to log it.

libraries/ms-drivers/test/test_spv1020_mppt.c
copperium: I suppose they aren't supposed to, but since the module doesn't keep state I guess this test is to make sure nothing funky happens if it's called multiple times. I'll add a warning when shut is called when it's already shut on x86.

libraries/ms-drivers/src/spv1020_mppt.c
AveryChiu64: Oh I see, so that saves you the from having to make the two files with duplicate functions

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
```diff
@@ -0,0 +1,48 @@
...
+// This file implements stm32-specific functionality. See ../spv1020_mppt.c for the rest.
+
+// Send a command to SPV1020 and get a response. |rx_len| must be <= 2.
+// It's safe to pass NULL as |rx_data| if |rx_len| is 0.
+static StatusCode prv_send_command(SpiPort port, uint8_t command, uint8_t *rx_data, size_t rx_len) {
+  if (port >= NUM_SPI_PORTS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  // We don't use spi_exchange since that assumes 0x00 is NOP, whereas SPV1020 appears to use 0x01.
```
g-jessmuir, line 13: I'm not sure this is correct. In the command list NOP is 0x01, but on page 12 paragraph 3 it says:
> Therefore the master appends a number of NOPs (0x00) to the command so that the entire data can be transmitted

It makes a lot more sense to me for this to be left as 0x00, matching spi_exchange. I think we should request a `smoke_spi` test with exchange before jumping into this weird custom spi implementation.

My theory is that the *command word* NOP is different from the SPI NOP, and that the SPI NOP is really 0x00.

I hate this datasheet.

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
```diff
@@ -0,0 +1,48 @@
...
+StatusCode spv1020_read_current(SpiPort port, uint16_t *current) {
+  return prv_send_command(port, COMMAND_READ_CURRENT, (uint8_t *)current, 2);
+}
+
+StatusCode spv1020_read_voltage_in(SpiPort port, uint16_t *vin) {
+  return prv_send_command(port, COMMAND_READ_VIN, (uint8_t *)vin, 2);
+}
+
+StatusCode spv1020_read_pwm(SpiPort port, uint16_t *pwm) {
+  return prv_send_command(port, COMMAND_READ_PWM, (uint8_t *)pwm, 2);
```
g-jessmuir, line 43: I think it makes sense to convert the PWM to the actual value here rather than the steps since it requires no calibration, but if you'd rather leave that to the user I'm ok with it. In other places we use uints as decimal by using 123.4% = `1234`.

libraries/ms-drivers/test/test_spv1020_mppt.c
```diff
@@ -0,0 +1,136 @@
...
+  TEST_ASSERT_BITS_LOW(0b1111110000000000, vin);
+}
+
+// Test that we can successfully read PWM and it's in the stated range.
+void test_read_pwm(void) {
+  uint16_t pwm = 0xFFFF;  // over 9 bits
+  TEST_ASSERT_OK(spv1020_read_pwm(TEST_SPI_PORT, &pwm));
+
+  // make sure the 6 highest bits are low, so it's a 9-bit value
+  TEST_ASSERT_BITS_LOW(0b1111111000000000, pwm);
```
g-jessmuir, line 67: could add an assertion that pwm is less than expected number of steps rather than just a 9 bit number, depends on if you add the conversion to actual pwm from the 9 bits.

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
```diff
@@ -0,0 +1,57 @@
...
+  return prv_send_command(port, COMMAND_READ_CURRENT, (uint8_t *)current, 2);
+}
+
+StatusCode spv1020_read_voltage_in(SpiPort port, uint16_t *vin) {
+  return prv_send_command(port, COMMAND_READ_VIN, (uint8_t *)vin, 2);
+}
+
+StatusCode spv1020_read_pwm(SpiPort port, uint16_t *pwm) {
+  uint16_t raw_pwm;
+  StatusCode code = prv_send_command(port, COMMAND_READ_PWM, (uint8_t *)raw_pwm, 2);
```
g-jessmuir, line 44: I believe this should be (uint8_t *)&raw_pwm, might need more brackets

libraries/ms-drivers/inc/spv1020_mppt_defs.h
```diff
@@ -0,0 +1,18 @@
...
+#define COMMAND_READ_CURRENT 0x04
+#define COMMAND_READ_VIN 0x05
+#define COMMAND_READ_PWM 0x06
+#define COMMAND_READ_STATUS 0x07
+
+// Bitmasks for extracting information from the status byte.
+#define OVC_BITMASK 0b1111000
+#define OVV_BITMASK 0b0000100
+#define OVT_BITMASK 0b0000010
+#define CR_BITMASK 0b0000001
```
GeraldNDA, line 18: Typically in driver code (and in general) we prefix these. In thise case I guess it'd be `SPV1020_` may or may not want to shorten BITMASK to MASK and COMMAND to CMD to compensate.

libraries/ms-drivers/src/x86/spv1020_mppt.c
```diff
@@ -0,0 +1,67 @@
...
+static bool s_is_shut = false;
+
+StatusCode spv1020_shut(SpiPort port) {
+  if (port >= NUM_SPI_PORTS) {
+    return STATUS_CODE_INVALID_ARGS;
+  }
+
+  // this doesn't occur on stm32 but is useful for debugging
+  if (s_is_shut) {
+    LOG_WARN("Issuing SHUT command to SPV1020 MPPT when it's already shut\r\n");
```
GeraldNDA, line 22: Why carriage returns?

copperium: Minicom needs carriage returns to move to the beginning of the new line like a typewriter. There's a setting to auto-insert carriage returns, but that was the rational behind putting them in. (I suppose this is x86 so we aren't using minicom anyways, I'll remove them.)

libraries/ms-drivers/src/stm32f0xx/spv1020_mppt.c
copperium: Whoops, fixed

PR #79, opened 2020-05-30
changed address to PA0 instead of PA1

JarvisWeng:


g-jessmuir:
wait I already assigned this to avery, can you do the board validation one? check jira

PR #81, opened 2020-06-01
Firmware 103 Submission

tristanle22:


projects/adc_reader/src/adc_reader.c
```diff
@@ -0,0 +1,66 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "interrupt_def.h"
+#include "wait.h"
+
+typedef struct {
```
g-jessmuir, line 8: by convention, we include the struct name before and after the definition, i.e.
```
typedef struct AdcReader { ... } AdcReader;
```

PR #82, opened 2020-06-02
[SOFT-202] Add headers / layout for BMS carrier

g-jessmuir:
Design here: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1290895361/Detailed+Design please review as well. Thanks!

### DIFF
--- a/projects/bms_carrier/inc/bms.h
+++ b/projects/bms_carrier/inc/bms.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <stdint.h>
+
+#include "debouncer.h"
+#include "status.h"
+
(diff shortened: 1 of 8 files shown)


g-jessmuir:
> Guess I'll take this and rebase and fool around with the structure as needed. The old AFE/Current Sense code was pretty intermingled with this I guess. I assume the plan is to relocate them?

For the AFE code I'm sorta counting on you to create whatever API makes the most sense and isn't too much work to implement (hopefully copying lots of code).

For current sense it was just implemented in two drivers and one module. Since we have different requirements (much faster data rate, different adc) I think it's fine to scrap the old code.

PR #85, opened 2020-06-04
Added x86 implementation for SPI

AveryChiu64:


libraries/ms-common/src/x86/spi.c
```diff
@@ -1,10 +1,70 @@
 #include "spi.h"
+#include "log.h"
+#include "spi_mcu.h"
+
+static GpioState cs_curr_state = GPIO_STATE_HIGH;
```
GeraldNDA, line 5: Should probably have the `s_` since static

libraries/ms-common/src/x86/spi.c
```diff
@@ -1,10 +1,70 @@
...
+  } else if (settings->mode >= NUM_SPI_MODES) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid SPI mode.");
+  }
+  return STATUS_CODE_OK;
+}
+
+StatusCode spi_tx(SpiPort spi, uint8_t *tx_data, size_t tx_len) {
+  LOG_DEBUG("Sending Data...\n");
+  for (size_t i = 0; i < tx_len; i++) {
+    LOG_DEBUG("%d\n", tx_data[i]);
```
g-jessmuir, line 20: sorry last thing, could you log these in hex? I think that makes it a bit easier since raw bytes aren't usually seen in decimal. Do `LOG_DEBUG("0x%x\n", tx_data[i]);` for here and the rx call as well please.

PR #87, opened 2020-06-06
[SOFT-206] Implement data_store solar module

copperium:
`data_store` provides an abstraction layer between reading the data and doing something with it. It stores all of the data collected by sense et al., then raises a `DATA_READY_EVENT`. Consumers of the data (`logger`, `fault_monitor`, etc) can retrieve the data directly from `data_store` to avoid overloading the event queue. All data is stored as `uint16_t`.

To store and retrieve data, we use a `DataPoint` enum with one entry for each data point we need (e.g. `DATA_POINT_VOLTAGE_1`, `DATA_POINT_MPPT_CURRENT_3`, etc.).

The `data_store_enter()` function takes a `DataPoint` parameter to identify which data point it is, as well as the data point as a `uint16_t`. It overwrites that data point in the module’s storage.

The `data_store_done()` function will be called by `sense` after a batch of data is read. It raises a `DATA_READY_EVENT` which will be received by the consumers of the data.

The `data_store_get()` function takes a `DataPoint` parameter and a `uint16_t*` and sets the pointer to the value of the data point.

### DIFF
--- a/projects/solar/src/data_store.c
+++ b/projects/solar/src/data_store.c
@@ -4,7 +4,7 @@
 
 static uint16_t s_data_store[NUM_DATA_POINTS];
 
-StatusCode data_store_enter(DataPoint data_point, uint16_t value) {
+StatusCode data_store_set(DataPoint data_point, uint16_t value) {
   if (data_point >= NUM_DATA_POINTS) {
     return status_code(STATUS_CODE_INVALID_ARGS);
   }
@@ -18,7 +18,7 @@ StatusCode data_store_done(void) {
 }
 
 StatusCode data_store_get(DataPoint data_point, uint16_t *value) {
-  if (!value || data_point >= NUM_DATA_POINTS) {
+  if (value == NULL || data_point >= NUM_DATA_POINTS) {
     return status_code(STATUS_CODE_INVALID_ARGS);
   }
   *value = s_data_store[data_point];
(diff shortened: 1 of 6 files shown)


projects/solar/src/data_store.c
```diff
@@ -0,0 +1,26 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode data_store_done(void) {
+  // the data ready event has no associated data
+  return event_raise_priority(DATA_READY_EVENT_PRIORITY, DATA_READY_EVENT, 0);
+}
+
+StatusCode data_store_get(DataPoint data_point, uint16_t *value) {
+  if (!value || data_point >= NUM_DATA_POINTS) {
```
GeraldNDA, line 21: Should explicitly check if `value == NULL`

projects/solar/test/test_data_store.c
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Test that |data_store_enter| responds correctly on an invalid index.
+void test_data_store_enter_invalid_index(void) {
+  TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, data_store_enter(NUM_DATA_POINTS, 0));
+}
+
+// Test that |data_store_get| responds correctly on an invalid index.
+void test_data_store_get_invalid_index(void) {
+  TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, data_store_enter(NUM_DATA_POINTS, 0));
```
GeraldNDA, line 39: I think `TEST_ASSERT_NOT_OK` would be sufficient here

copperium:
data_store_set is a better name, changed it

PR #89, opened 2020-06-06
soft_999_Jenish_Patel_fw_102

jenishpatel2147:


soft_999_Jenish_Patel_fw_102
```diff
@@ -0,0 +1,45 @@
+// Blink an LED
```
g-jessmuir, line 1: remove unncecessary comment please

soft_999_Jenish_Patel_fw_102
```diff
@@ -0,0 +1,45 @@
...
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define HALF_SECOND 500  //
```
g-jessmuir, line 8: since this is a value in milliseconds, our convention is to suffix it with `_MS`, please rename

soft_999_Jenish_Patel_fw_102
```diff
@@ -0,0 +1,45 @@
...
+
+typedef struct CounterStorage {
+  uint16_t counter_a;
+  uint16_t counter_b;
+} CounterStorage;
+
+void prv_timer_call(const SoftTimerId timer_id, void *empty) {
```
g-jessmuir, line 15: why is it called empty? could be confusing, please rename

soft_999_Jenish_Patel_fw_102
```diff
@@ -0,0 +1,45 @@
...
+  CounterStorage *storage = empty;
+  storage->counter_a++;
+  storage->counter_b++;
+
+  // Prints Every half-second
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  // Prints Every Second
+  if (storage->counter_b % 2 == 0) {
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b / 2);
```
g-jessmuir, line 26: interesting solution, but yea this works. Could be confusing for someone to read though. I'm ok with this for here, but I think a better solution is to only increment every half second because it makes it more obvious what's happening.

g-jessmuir:
Just add the makefile and hit 'ready for review' then we can call it done

jenishpatel2147:
Hi Jess,

I placed them in it's own folder, inside of the Projects folder, it didn't catch any errors.

g-jessmuir:
You seem to have increased the number of files by a lot, could you take another look? some are different from each other, not entirely sure which ones are most recent

jenishpatel2147:
Hi, yeah I deleted all the files, and re-uploaded only the correct ones.

projects/soft_999_Jenish_Patel_fw_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+  // Prints Every half-second
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  // Prints Every Second
+  if (storage->counter_b % 2 == 0) {
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b / 2);
+  }
+
+  // start the timer again, so it keeps periodically flipping coins
+  soft_timer_start_millis(HALF_SECOND_MS, prv_timer_call, &storage, NULL);
```
g-jessmuir, line 29: I think you have an error here with the type of `storage`, check the parameters that `soft_timer_start` takes. Also check your output (run it locally) to check for errors.

PR #92, opened 2020-06-09
Soft 173 ads1259 driver

mitchellostler:
### Initial Draft PR for New ads1259 Driver
- ads1259_adc_defs.h is complete. Macros have been added to compute value for register configuration depending on whether different options are enabled or disabled. 

- ads1259_adc.h may still need work, settings and storage structs are fully defined, as well as function stubs, but may need to add other typedefs

- ads1259_init() logic is complete, and as all functions needed for it to operate. Calibration is done initially on register configuration

- ads1259_get_data() needs more work. Message rx is complete, but processing data needs revision

- TODO: add error handling and calibration during operation. Calibration is just a single command (for each type of calibration), but need to figure out where it will be implemented and after what offset/gain value calibration needs to occur.

Let me know if you have any questions


libraries/ms-drivers/inc/ads1259_adc.h
```diff
@@ -0,0 +1,44 @@
...
+//Time in ms needed after calibration for each data rate
+typedef enum {
+  ADS1259_CALIBRATION_TIME_10 = 1900,
+  ADS1259_CALIBRATION_TIME_17 = 1140,
+  ADS1259_CALIBRATION_TIME_50 = 380,
+  ADS1259_CALIBRATION_TIME_60 = 318,
+  ADS1259_CALIBRATION_TIME_400 = 49,
+  ADS1259_CALIBRATION_TIME_1200 = 17,
+  ADS1259_CALIBRATION_TIME_3600 = 7,
+  ADS1259_CALIBRATION_TIME_3600 = 3,
```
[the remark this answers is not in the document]
mitchellostler: @g-jessmuir Do you think this would be better in a lookup array with the indices as the SPS values?

libraries/ms-drivers/inc/ads1259_adc_defs.h
```diff
@@ -0,0 +1,91 @@
...
+#define ADS1259_ADDRESS_OFC2 0x05
+#define ADS1259_ADDRESS_FSC0 0x06
+#define ADS1259_ADDRESS_FSC1 0x07
+#define ADS1259_ADDRESS_FSC2 0x08
+#define NUM_REGISTERS 0x09
+
+// ADS1259 Register Configurations - LSB to MSB
+//These values can be changed based on needed configurations
+
+// CONFIG0
```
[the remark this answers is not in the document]
mitchellostler: @g-jessmuir Basically what I did with the macros is create flags for all the configurable settings, and then combine them into a hex value to write to the registers. The idea behind this was that all that would be needed to change the settings for the adc would be flipping a bit in these settings, and then the macro would do the rest. Is this not ideal?

libraries/ms-drivers/src/stm32f0xx/ads159_adc.c
```diff
@@ -0,0 +1,115 @@
...
+
+
+static void prv_stop_rdatac(Ads1259Storage* storage) {
+    uint8_t payload[] = { ADS1259_STOP_READ_DATA_CONTINUOUS };
+    delay_us(100); // (SOFT-173 TODO) Must wait at leas 2^16 clk cycles before sending first command - change?
+    spi_tx(storage->spi_port, payload, sizeof(payload));
+}
+
+static StatusCode prv_configure_registers(Ads1259Storage* storage) {
+    uint8_t payload[] =  { ADS1259_WRITE_REGISTER | ADS1259_ADDRESS_CONFIG0, 3, ADS1259_CONFIG0_SETTINGS,
```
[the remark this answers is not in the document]
mitchellostler: @g-jessmuir I'm a bit confused by this, as there are only 3 configuration registers, but in the data sheet example (page 40) they write to the other registers with their power-up reset values. What do you think, should I write the reset values or just the configurations? I could also just send the reset commands later on in the init function which would have the same effect.
Screen Shot 2020-06-11 at 5 12 26 PM

g-jessmuir:
Recommended steps:
1. move the driver to the common drivers folder (not x86 or stm32). There's no super intricate hardware interaction so I think it's fine to leave it common. (also change filename to ads1259_adc.c rather than ads159_ads.c :slightly_smiling_face: )
2. Include "interrupt.h" and interrupt_init() to your test file before initializing soft timers so that it doesn't crash
3. In your test, mock spi_tx / spi_rx / spi_exchange so that you can assert the proper thing is returned in your tests

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,147 @@
...
+      rx_data[1] = 0xFF;
+      rx_data[2] = 0xFF;
+      rx_data[3] = (uint8_t)((0x2FD + ADS1259_CHECKSUM_OFFSET) & 0x7F);
+      break;
+  }
+  return STATUS_CODE_OK;
+}
+
+static void prv_error_handler(Ads1259StatusCode code, void *context) {
+  if (code == ADS1259_STATUS_CODE_OUT_OF_RANGE) LOG_DEBUG("ADC READING IS OUT OF RANGE\n");
```
g-jessmuir, line 103: rather than just logging, what you can do is create a static `Ads1259StatusCode s_gotten_code` and set it with whatever the error handler is called with. Then, you can assert in tests that the code is what's expected. E.g.
```
static Ads1259StatusCode s_gotten_code;
static void prv_error_handler(... ...) {
  s_gotten_code = code;
}
...
void test_correct_error_code(void) {
  // whatever you need to do to get the error handler called
  ...
  TEST_ASSERT_EQUAL(expected_code, s_gotten_code);
}
```

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,147 @@
...
+      break;
+    // Start of Ads1259_get_conversion_data()
+    case ADS1259_START_CONV:
+      LOG_DEBUG("CONVERSIONS STARTED \n");
+      break;
+    case ADS1259_READ_DATA_BY_OPCODE:
+      LOG_DEBUG("SETTING CONVERSION DATA\n");
+      rx_data[0] = 0xFF;
+      rx_data[1] = 0xFF;
+      rx_data[2] = 0xFF;
```
g-jessmuir, line 95: rather than just returning `0xFF` here, what you can do is have a global static `uint8_t[4] s_rx_data` that you can set in tests then set the `rx_data` here to that static one. That way, you can set it to return various things in tests, then assert that your driver acts correctly according to what it gets in the rx.

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,147 @@
...
+
+void test_ads1259_init() {
+  uint32_t test_raw = 0xFFFFFF;
+  ads1259_get_conversion_data(&storage);
+  delay_ms(TEST_DATA_SETTLING_TIME_MS);
+  TEST_ASSERT_EQUAL(storage.conv_data.MSB | storage.conv_data.MID | storage.conv_data.LSB, 0xFF);
+  // TEST_ASSERT_EQUAL(storage.conv_data.raw, test_raw);
+  TEST_ASSERT_EQUAL(storage.reading, (test_raw >> 4) * EXTERNAL_VREF / pow(2, 20));
+  LOG_DEBUG("VOLTAGE READING: %lf\n", storage.reading);
+}
```
g-jessmuir, line 147: a few more tests to add:

- driver returns the correct voltage for a zero voltage
- driver returns correct error code if out of range
- driver returns correct error code if the checksum isn't correct
- driver returns correct voltage for some other arbitrary voltage that isn't full-scale or zero (can also use this test to assert your endianness logic works correctly)

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -0,0 +1,136 @@
...
+static void prv_conversion_callback(SoftTimerId timer_id, void *context) {
+  Ads1259Storage *storage = (Ads1259Storage *)context;
+  Ads1259StatusCode code;
+  uint8_t payload[] = { ADS1259_READ_DATA_BY_OPCODE };
+  spi_exchange(storage->spi_port, payload, 1, (uint8_t *)&storage->rx_data, NUM_ADS_RX_BYTES);
+  code = prv_checksum(storage);
+  if (code) {
+    (*storage->handler)(code, NULL);
+  }
+  storage->conv_data.MSB = storage->rx_data.LSB;
```
g-jessmuir, line 98: Why are you flipping these? the MSB should always be the MSB since you flipped the order in the structs

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,147 @@
...
+  TEST_ASSERT_EQUAL((registers.FSC_0 | registers.FSC_1 | registers.FSC_2), 0);
+  TEST_ASSERT_EQUAL((registers.OFC_0 | registers.OFC_1 | registers.OFC_2), 0);
+  TEST_ASSERT_EQUAL(registers.CONFIG_0, register_lookup[0]);
+  TEST_ASSERT_EQUAL(registers.CONFIG_1, register_lookup[1]);
+  TEST_ASSERT_EQUAL(registers.CONFIG_2, register_lookup[2]);
+}
+
+void teardown_test(void) {}
+
+void test_ads1259_init() {
```
g-jessmuir, line 139: please rename test to be more descriptive, doesn't seem like you're testing `init` here

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,147 @@
...
+  uint32_t test_raw = 0xFFFFFF;
+  ads1259_get_conversion_data(&storage);
+  delay_ms(TEST_DATA_SETTLING_TIME_MS);
+  TEST_ASSERT_EQUAL(storage.conv_data.MSB | storage.conv_data.MID | storage.conv_data.LSB, 0xFF);
```
g-jessmuir, line 143: our assertions are the other way around: the expected comes first, then the actual. Check the function parameters

libraries/ms-drivers/test/test_ads1259_adc.c
```diff
@@ -0,0 +1,268 @@
...
+  delay_ms(TEST_DATA_SETTLING_TIME_MS);
+  TEST_ASSERT_EQUAL(-50, s_storage.reading);
+
+  // test with min readable pos data
+  s_test_mode = ADS1259_MODE_MIN_POS_DATA;
+  ads1259_get_conversion_data(&s_storage);
+  delay_ms(TEST_DATA_SETTLING_TIME_MS);
+  TEST_ASSERT_EQUAL(0.000095, s_storage.reading);
+
+  // test with min readable pos data
```
mitchellostler, line 232: realize pos should be neg here

PR #94, opened 2020-06-10
Soft 196 adt7470

Max-MZ:
initial starting block

### DIFF
--- a/libraries/ms-drivers/inc/adt_7470_fan_controller.h
+++ b/libraries/ms-drivers/inc/adt_7470_fan_controller.h
@@ -1,0 +1,41 @@
+#pragma once
+
+// Reads current values from 2 ADC pins on the BTS 7200 switched between with a selection pin.
+// Requires GPIO, interrupts, soft timers, and ADC to be initialized in ADC_MODE_SINGLE.
+// If using with MCP23008, requires I2C to be initialized.
+
+#include "adc.h"
+#include "gpio.h"
+#include "i2c.h"
+#include "soft_timer.h"
+
+typedef void (*Adt7470DataCallback)();
+
+typedef struct {
+  I2CPort i2c_port;
+  GpioAddress *fan_1_pin;
+  GpioAddress *fan_2_pin;
+  GpioAddress *fan_3_pin;
+  GpioAddress *fan_4_pin;
+  uint32_t interval_us;
+  Adt7470DataCallback
+      callback;  // set to NULL for no callback, probably don't need since manual control
+  void *callback_context;
+} Adt7470Settings;
+
+typedef struct {
+  GpioAddress *fan_1_pin;
+  GpioAddress *fan_2_pin;
+  GpioAddress *fan_3_pin;
+  GpioAddress *fan_4_pin;
+  uint32_t interval_us;
+  SoftTimerId timer_id;
+  Adt7470DataCallback callback;
+  void *callback_context;
+} Adt7470Storage;
+
+// Initialize the Adt7200 with the given settings; the select pin is an STM32 GPIO pin.
+StatusCode adt_7200_init_stm32(Adt7470Storage *storage, Adt7470Settings *settings);
+
+// Translate and write the new speed
+StatusCode apt7470_set_speed(I2CPort port, uint8_t *status);
--- a/libraries/ms-drivers/inc/adt_7470_fan_controller.h
+++ b/libraries/ms-drivers/inc/adt_7470_fan_controller.h
@@ -5,7 +5,7 @@
 #include "i2c.h"
 #include "soft_timer.h"
 
-typedef void (*Adt7470DataCallback)();
+typedef void (*Adt7470DataCallback)(void *context);
 
 typedef struct {
   I2CPort i2c_port;
@@ -13,9 +13,10 @@ typedef struct {
   GpioAddress *fan_2_pin;
   GpioAddress *fan_3_pin;
   GpioAddress *fan_4_pin;
-  uint32_t interval_us;
+  uint32_t interval_ms;
   Adt7470DataCallback callback;  // set to NULL for no callback
   void *callback_context;
+  I2CPort i2c;
 } Adt7470Settings;
 
 typedef struct {
@@ -23,17 +24,19 @@ typedef struct {
   GpioAddress *fan_2_pin;
   GpioAddress *fan_3_pin;
   GpioAddress *fan_4_pin;
-  uint32_t interval_us;
+  uint32_t interval_ms;
   SoftTimerId timer_id;
   Adt7470DataCallback callback;
   void *callback_context;
+  I2CPort i2c;
 } Adt7470Storage;
 
 // probably a better name for these
-typedef enum { FAN_1 = 0, FAN_2, FAN_3, FAN_4 } Fan;
+typedef enum { FAN_1 = 0, FAN_2, FAN_3, FAN_4 } FAN_NUM;
 
-// Initialize the Adt7200 with the given settings; the select pin is an STM32 GPIO pin.
-StatusCode adt_7200_init_stm32(Adt7470Storage *storage, Adt7470Settings *settings);
+// Initialize the Adt7470 with the given settings; the select pin is an STM32 GPIO pin.
+StatusCode adt7470_init(Adt7470Storage *storage, Adt7470Settings *settings);
 
 // Translate and write the new speed
-StatusCode apt7470_set_speed(I2CPort port, uint8_t *status, Fan Fan, uint16_t ADR7470_I2C_ADDRESS);
+StatusCode adt7470_set_speed(I2CPort port, uint8_t speed, uint8_t FAN_PWM_ADDR,
+                             uint16_t ADR7470_I2C_ADDRESS);
--- a/libraries/ms-drivers/inc/adt_7470_fan_controller.h
+++ b/libraries/ms-drivers/inc/adt_7470_fan_controller.h
@@ -1,9 +1,5 @@
 #pragma once
 
-// Reads current values from 2 ADC pins on the BTS 7200 switched between with a selection pin.
-// Requires GPIO, interrupts, soft timers, and ADC to be initialized in ADC_MODE_SINGLE.
-// If using with MCP23008, requires I2C to be initialized.
-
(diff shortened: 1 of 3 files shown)


libraries/ms-drivers/inc/adt_7470_fan_controller.h
```diff
@@ -0,0 +1,39 @@
...
+  GpioAddress *fan_3_pin;
+  GpioAddress *fan_4_pin;
+  uint32_t interval_us;
+  SoftTimerId timer_id;
+  Adt7470DataCallback callback;
+  void *callback_context;
+} Adt7470Storage;
+
+// probably a better name for these
+typedef enum { FAN_1 = 0, FAN_2, FAN_3, FAN_4 } Fan;
```
g-jessmuir, line 33: By convention (and for utility), at the end of enums we add a `NUM_...` value e.g. `NUM_FANS`. This helps iteration so we can do `for (uint8_t i = 0; i < NUM_FANS; i++)` which is a bit more descriptive

libraries/ms-drivers/inc/adt_7470_fan_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+// Initialize the Adt7200 with the given settings; the select pin is an STM32 GPIO pin.
+StatusCode adt_7200_init_stm32(Adt7470Storage *storage, Adt7470Settings *settings);
```
g-jessmuir, line 36: should be fine to just call it `adt7470_init`, no need for stm32. Also, where did 7200 come from?

libraries/ms-drivers/src/adt_7470_fan_controller.c
```diff
@@ -0,0 +1,61 @@
...
+    .sda = I2C1_SDA,
+    .scl = I2C1_SCL,
+  };
+
+  i2c_init(I2C_PORT_1, &i2c1_settings);
+
+  return STATUS_CODE_OK;
+}
+
+StatusCode bts_7240_init(Adt7470Storage *storage, Adt7470Settings *settings) {
```
g-jessmuir, line 30: bts 7240 ?? I swear you're just using random letters and numbers at this point 😅

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
```diff
@@ -0,0 +1,55 @@
...
+
+#define ADT7470_PWM1_DUTY_CYCLE 0x32
+#define ADT7470_PWM2_DUTY_CYCLE 0x33
+#define ADT7470_PWM3_DUTY_CYCLE 0x34
+#define ADT7470_PWM4_DUTY_CYCLE 0x35
+
+#define ADT7470_FAN_MODE_REGISTER_1 0x68  // bits 6 and 7 contorl the mode for PWM2 and PWM1
+#define ADT7470_FAN_MODE_REGISTER_2 0x69  // bits 6 and 7 contorl the mode for PWM4 and PWM3
+
+#define ADT7470_MANUAL_MODE_MASK 00001111
```
g-jessmuir, line 25: I think you should define these with the `0b` prefix so they're actually the correct mask

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
```diff
@@ -0,0 +1,55 @@
...
+  0x80  // Setting a direction bit to 1 in the GPIO configuration register makes the corresponding
+        // GPIO pin an output.
+        //  Clearing the direction bit to 0 makes it an input.  bits 7, 5, 3, 1
+#define ADT7470_GPIO_POLARITY_REGISTER \
+  0x80  // Setting a polarity bit to 1 makes the corresponding GPIO pin active high. Clearing the
+        // polarity bit to 0 makes it active low.
+
+// how are these read? on the schematic it says I2C address: 0101100_R/W and hex values for read or
+// write. assuming those are tagged on after a don't care for a total of 16 bits?
+#define ADR7470_I2C_ADDRESS_1 010111  // 15 bits?
```
g-jessmuir, line 44: per the schematics I guess the address is dependent on whether you're reading or writing. This is abstracted by our i2c library, you can check libraries/ms-common/src/stm32f0xx/i2c.c for implementation details. In our implementation the address gets shifted left one bit, so I think you actually want to take the "right justified" address. Maybe review the i2c notes on how addressing works, or ask me for clarification
image

g-jessmuir: Also fine to just use hex, not binary here

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
```diff
@@ -0,0 +1,55 @@
...
+#define ADR7470_I2C_ADDRESS_2 0101100
+
+#define I2C1_SDA \
```
g-jessmuir, line 47: These pins aren't quite right, and should be passed in as parameters for i2c settings to init rather than defined here

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
```diff
@@ -0,0 +1,55 @@
...
+  { .port = GPIO_PORT_B, .pin = 16 }
+#define I2C1_SCL \
+  { .port = GPIO_PORT_B, .pin = 1 }
+
+#define SET_SPEED_NUM_BYTES 2
+
+// need to create masks for temperature interrupt, since we aren't handling that in this driver
```
g-jessmuir, line 54: we're not worrying about temperature with this since we have thermistors on the batteries, you can skip this part 🎉

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
```diff
@@ -0,0 +1,55 @@
...
+// need to set up interrupt? somehow not sure how it is raised yet.
```
g-jessmuir, line 55: what interrupt are you talking about here?

libraries/ms-drivers/src/adt_7470_fan_controller.c
```diff
@@ -0,0 +1,61 @@
+#include "adt_7470_fan_controller.h"
+
+#include <stddef.h>
+
+#include "adt_7470_fan_controller_defs.h"
+
+#define STM32_GPIO_STATE_SELECT_OUT_0 GPIO_STATE_LOW
+#define STM32_GPIO_STATE_SELECT_OUT_1 GPIO_STATE_HIGH
```
g-jessmuir, line 8: What are these defines used for?

libraries/ms-drivers/src/adt_7470_fan_controller.c
```diff
@@ -0,0 +1,61 @@
...
+// need to set interrupt once fan goes out of range
+// PWM duty cycle is set from 0-100, in steps of 0.39 (0 - 0xFF)
+StatusCode apt7470_set_speed(I2CPort port, uint8_t *speed, Fan FAN, uint16_t ADR7470_I2C_ADDRESS) {
+  // need to select duty fan - correct fan duty cycle
+  return i2c_write_reg(I2C_PORT_1, ADR7470_I2C_ADDRESS, ADT7470_PWM2_DUTY_CYCLE, speed,
```
g-jessmuir, line 13: you should be using `port` as the argument instead of `I2C_PORT_1` I think

libraries/ms-drivers/src/adt_7470_fan_controller.c
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  memset(storage, 0, sizeof(Adt7470Storage));
+
+  // 4 gpio's need to be set to PWM
+  storage->fan_1_pin = settings->fan_1_pin;
+  storage->fan_2_pin = settings->fan_2_pin;
+  storage->fan_3_pin = settings->fan_3_pin;
+  storage->fan_4_pin = settings->fan_4_pin;
+  storage->interval_us = settings->interval_us;
```
g-jessmuir, line 42: what's this interval used for? might be more clear if it's in milliseconds since that's generally the unit we use

libraries/ms-drivers/inc/adt_7470_fan_controller.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "adc.h"
+#include "gpio.h"
+#include "i2c.h"
+#include "soft_timer.h"
+
+typedef void (*Adt7470DataCallback)();
```
g-jessmuir, line 8: wouldn't hurt to include a `void *context` parameter, makes it simpler to use externally

libraries/ms-drivers/inc/adt_7470_fan_controller_defs.h
g-jessmuir: Actually maybe it makes more sense to just leave these as parameters when calling either public function, so that the user (fan control module in bms carrier) can more easily set them individually

libraries/ms-drivers/src/adt_7470_fan_controller.c
```diff
@@ -0,0 +1,61 @@
...
+#define STM32_GPIO_STATE_SELECT_OUT_1 GPIO_STATE_HIGH
+// need to set interrupt once fan goes out of range
+// PWM duty cycle is set from 0-100, in steps of 0.39 (0 - 0xFF)
+StatusCode apt7470_set_speed(I2CPort port, uint8_t *speed, Fan FAN, uint16_t ADR7470_I2C_ADDRESS) {
+  // need to select duty fan - correct fan duty cycle
+  return i2c_write_reg(I2C_PORT_1, ADR7470_I2C_ADDRESS, ADT7470_PWM2_DUTY_CYCLE, speed,
+                       SET_SPEED_NUM_BYTES);
+}
+
+static StatusCode prv_init_common(Adt7470Storage *storage) {
```
g-jessmuir, line 17: I think you can just include this step in the general init since it only happens once ish

g-jessmuir:
- need to also expose "grab_status()"
- need to register interrupt on SMBALERT pin (pin should be in setttings) (callback function pointer also should be in settings)

Make driver 'static', no state should be stored in driver.
When interrupt handler is called, module should grab state from both fan controllers.

PR #96, opened 2020-06-12
[SOFT-187] MCP3427 ADC

copperium:
After many trials and tribulations and several instances of misunderstanding the driver, it seems to finally be ready. I've tried to document and test it thoroughly, and also made a bunch of improvements to the base driver.

The driver uses an FSM to continually read from the MCP3427 regardless of whether it's in continuous or one-shot mode (requesting a conversion each time if the latter). The FSM, as far as I can tell, is so that the cycle is interrupted by the event queue and other events are allowed to process.

I also fixed a bug in `ms_test_helpers.h` that bit me in the tests.


libraries/ms-common/inc/ms_test_helpers.h
```diff
@@ -55,16 +55,16 @@
 #define MS_TEST_HELPER_ASSERT_EVENT(event, e_id, e_data) \
   ({                                                     \
     TEST_ASSERT_OK(event_process(&(event)));             \
-    TEST_ASSERT_EQUAL((e_id), event.id);                 \
-    TEST_ASSERT_EQUAL((e_data), event.data);             \
+    TEST_ASSERT_EQUAL((e_id), (event).id);               \
```
AveryChiu64, line 58: How did this fix your test? What difference does the bracket make? I'm interested to know what happens when there is a bracket and when there is no bracket.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+FSM_STATE_TRANSITION(channel_2_readback) {
+  Mcp3427Storage *storage = fsm->context;
+  FSM_ADD_TRANSITION(storage->data_trigger_event, channel_1_trigger);
+}
+
+static uint16_t prv_get_chip_identifier(Mcp3427Storage *storage) {
+  // used to gate events we raised to only this MCP3427
+  return (storage->addr << 8) | (storage->port == I2C_PORT_1 ? 0 : 1);
```
AveryChiu64, line 47: Does this work if you don't use a ternary operator since it should evaluate to 1 or 0?

AveryChiu64: Eg when I do printf("%d", ((1 << 1) | (1 == 1))); , it gives me 3

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
```diff
@@ -0,0 +1,180 @@
...
+
+  StatusCode status = i2c_write(storage->port, storage->addr, &config, MCP3427_NUM_CONFIG_BYTES);
+  // start the state machine ready to transition to channel_1_trigger
+  fsm_init(&storage->fsm, MCP3427_FSM_NAME, &channel_2_readback, storage);
+  return status;
+}
+
+StatusCode mcp3427_register_callback(Mcp3427Storage *storage, Mcp3427Callback callback,
+                                     void *context) {
+  if (storage == NULL || callback == NULL) {
```
AveryChiu64, line 149: I think you can do !(storage && callback) but what you have is fine, this isn't really a big deal

AveryChiu64: This appears in other parts too, so storage == NULL can become !storage or smthing. As I said this isn't really a big deal, up to you

libraries/ms-common/inc/ms_test_helpers.h
copperium: Macros work by simple substitution, so whatever you put in the place of the `event` parameter gets substituted whenever the C preprocessor sees `event`. With a single identifier this works fine with no brackets, but I was passing `*e`, which caused replacements of e.g. `*e.id`, which tries to dereference `e.id` and fails. Putting brackets fixes the precedence issue. It's generally good practice to put brackets around every macro parameter (and the macro itself) every time you use them.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
copperium: It will work without a ternary operator, I thought that would be a little less clear though.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
copperium: Lol, people gave the opposite comments when I did `!storage` last time. `== NULL` is probably slightly clearer.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
```diff
@@ -0,0 +1,180 @@
...
+// Note: we always read from both channels before reporting, taking 100ms between reports. If timing
+// is an issue, an optimization for boards (i.e. solar) using only one channel is to only read that
+// channel, so reports take 50ms.
+
+// We use a combination of the address and I2C port in the event data to only transition on events
+// that we raised. Possible optimization if necessary: make prv_get_chip_identifier public, then
+// keep a reverse lookup table of identifier -> MCP3427 storage and only call the appropriate
+// MCP3427 when receiving an MCP3427 event. That way, using n MCP3427s generates n calls to
+// |mcp3427_process_event| per cycle rather than n^2 if we pass every event to every MCP3427.
+// (I think this was done on MSXII.)
```
copperium: I think this might be worth implementing since we can have a theoretical maximum of 18 MCP3427s, meaning 324 calls to `mcp3427_process_event` each cycle. I'll implement it.

copperium:
Finally figured out my git hooks settings, hopefully there shouldn't be any more of these fun surprise commits.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
```diff
@@ -0,0 +1,196 @@
...
+#include "fsm.h"
+#include "log.h"
+#include "mcp3427_adc_defs.h"
+#include "soft_timer.h"
+
+// Note: we always read from both channels before reporting, taking 100ms between reports. If timing
+// is an issue, an optimization for boards (i.e. solar) using only one channel is to only read that
+// channel, so reports take 50ms.
+
+#define MCP3427_FSM_NAME "MCP3427 FSM"
```
g-jessmuir, line 11: still don't know why we bother with these, maybe we should just open a ticket to remove the fsm name but this is fine here

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
```diff
@@ -0,0 +1,196 @@
...
+  Mcp3427Storage *storage = fsm->context;
+  FSM_ADD_TRANSITION(storage->data_trigger_event, channel_1_trigger);
+}
+
+static uint8_t prv_get_chip_identifier(Mcp3427Storage *storage) {
+  // Used to gate events we raised to only this MCP3427.
+  // ID is 4 bits, with the "base address" (the 3 least significant bits of the I2C address) as the
+  // most significant bits and the I2C port encoded as the LSB.
+  uint8_t base_addr = storage->addr ^ (MCP3427_DEVICE_CODE << 3);
+  return (base_addr << 1) | (storage->port == I2C_PORT_1 ? 0 : 1);
```
g-jessmuir, line 50: is there any reason you're cramming this into a `uint8_t`? couldn't you just define a chip identifier struct? I think this is fine, but I feel like best practice is probably to not have complex expressions like this since they tend to be most error prone.

libraries/ms-drivers/test/test_mcp3427_adc.c
```diff
@@ -0,0 +1,371 @@
...
+
+  I2CSettings i2c_settings = {
+    .speed = I2C_SPEED_FAST,
+    .sda = TEST_CONFIG_PIN_I2C_SDA,
+    .scl = TEST_CONFIG_PIN_I2C_SCL,
+  };
+  i2c_init(TEST_I2C_PORT, &i2c_settings);
+
+  s_times_callback_called = s_times_fault_callback_called = 0;
+  s_callback_context = s_fault_callback_context = NULL;
```
g-jessmuir, line 72: I'm faaaaairly certain this breaks google C++ style guide but can't prove it with a skim through. Mostly just a nitpick though, whatever really.

libraries/ms-drivers/test/test_mcp3427_adc.c
```diff
@@ -0,0 +1,371 @@
...
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+  TEST_ASSERT_EQUAL(1, s_times_callback_called);            // callback called since it's CH2
+  TEST_ASSERT_EQUAL(callback_context, s_callback_context);  // context was passed
+
+  // the fault callback shouldn't have been called ever
+  TEST_ASSERT_EQUAL(0, s_times_fault_callback_called);
+  TEST_ASSERT_EQUAL(NULL, s_fault_callback_context);
+
+  // clean up for a possible next round
+  s_callback_context = NULL;
```
g-jessmuir, line 112: you skipped this in your test for the sense module since it's in setup_test. Could also put it in teardown_test, but I don't think this should go here.

libraries/ms-drivers/src/stm32f0xx/mcp3427_adc.c
copperium: I'm cramming it into the rightmost 4 bits of a `uint8_t` because I'm using the chip ID as an index for `s_id_to_storage_cache`, and I'm also using it as the data for the events so that we can direct them to the appropriate storage. This would be a lot harder to do with a struct.

g-jessmuir: Fair enough. Something we have in centre console is a 
```
typedef union FaultReason {
    struct {
        uint8_t area;
        uint8_t reason;
    } fields;
    uint16_t raw;
} FaultReason;
```
But To be honest that's not really any easier to work with than packing into an int since it tends to add a lot of lines wherever you declare them. For events you pass in the `fault.raw`, then when unpacking them you do `FaultReason fault = { .raw = e.data };` then access it via `fault.fields.reason` etc.

libraries/ms-drivers/inc/mcp3427_adc.h
```diff
@@ -0,0 +1,108 @@
...
+  uint16_t sensor_data[NUM_MCP3427_CHANNELS];
+  EventId data_trigger_event;
+  EventId data_ready_event;
+  Mcp3427Callback callback;
+  void *context;
+  Mcp3427FaultCallback fault_callback;
+  void *fault_context;
+  Mcp3427SampleRate sample_rate;
+  Fsm fsm;
+  Mcp3427Channel current_channel;  // only used on x86
```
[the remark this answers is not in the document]
copperium: The stm32 version reads the current channel from one bit in `config`, but I figured this was clearer for x86.

PR #97, opened 2020-06-13
Soft 203 i2c implementation x86

AveryChiu64:


libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,65 @@
...
+  return STATUS_CODE_OK;
 }
 
 StatusCode i2c_read_reg(I2CPort i2c, I2CAddress addr, uint8_t reg, uint8_t *rx_data,
                         size_t rx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Reading Register Address...\n");
```
copperium, line 44: See above: "Reading Register Address" -> "Reading %d bytes from register %d over I2C". I don't think logging the placeholder data we're reading is necessary.

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,65 @@
...
+  for (size_t i = 0; i < rx_len; i++) {
+    // Insert dummy data
+    if (i % 2 == 0) {
+      rx_data[i] = 1;
+    }
```
copperium, line 49: See above re: this won't populate odd indices

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,65 @@
...
+  return STATUS_CODE_OK;
 }
 
 StatusCode i2c_write_reg(I2CPort i2c, I2CAddress addr, uint8_t reg, uint8_t *tx_data,
                          size_t tx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Writing Register Address...\n");
```
copperium, line 60: See above: "Writing Register Address" -> "Writing %d bytes to register 0x%x over I2C", also see if there's a way to put the logging on one line (not a big deal if there isn't).

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,62 @@
...
 StatusCode i2c_write(I2CPort i2c, I2CAddress addr, uint8_t *tx_data, size_t tx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Sending %ld bytes over I2C\n", tx_len);
+  for (size_t i = 0; i < tx_len; i++) {
+    LOG_DEBUG("0x%x", tx_data[i]);
+  }
+  LOG_DEBUG("\n");
```
copperium, line 34: `LOG_DEBUG` inserts a prefix like `[0] libraries/src/ms-common/i2c.c:50:` before its argument every time, so this won't work for putting the bytes on one line.

To put everything on one line, I'd recommend removing the \n from the "Sending %ld bytes over I2C" log and then using printf for the other logs.

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,62 @@
...
                          size_t tx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Writing %ld bytes from register %d over I2C\n", tx_len, reg);
+  for (size_t i = 0; i < tx_len; i++) {
+    LOG_DEBUG("0x%x", tx_data[i]);
+  }
+  LOG_DEBUG("\n");
```
copperium, line 60: Same thing here.

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,63 @@
...
 StatusCode i2c_write(I2CPort i2c, I2CAddress addr, uint8_t *tx_data, size_t tx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Sending %ld bytes over I2C\n", tx_len);
+  for (size_t i = 0; i < tx_len; i++) {
+    printf("0x%x ", tx_data[i]);
+  }
+  printf("\n");
```
copperium, line 35: Shoot, I just realized... this is going to break filtering the verbosity of the log. Could you surround the for loop and printf with `if (LOG_LEVEL_DEBUG >= LOG_LEVEL_VERBOSITY) {..}` please? That way the printfs will behave nicely like `LOG_DEBUG`.

(Also, could you change the `\n` in the `LOG_DEBUG("Sending %ld bytes over I2C\n")` to "`: `" so that the whole message is on one line?)

libraries/ms-common/src/x86/i2c.c
```diff
@@ -1,23 +1,63 @@
...
                          size_t tx_len) {
-  return STATUS_CODE_UNIMPLEMENTED;
+  if (i2c >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
+  }
+  LOG_DEBUG("Writing %ld bytes to register %d over I2C\n", tx_len, reg);
+  for (size_t i = 0; i < tx_len; i++) {
+    printf("0x%x ", tx_data[i]);
+  }
+  printf("\n");
```
copperium, line 61: Same thing here.

PR #98, opened 2020-06-13
[SOFT-148] centre console fault handling

g-jessmuir:
Updated to fault to off from a bps heartbeat fault


g-jessmuir:
**This PR comes with the caveat that we should probably rework some of the sequences to reflect updated hardware, and probably change the power off sequence so that it's unable to fault

libraries/ms-common/inc/event_queue.h
```diff
@@ -48,6 +48,9 @@ StatusCode event_raise_priority(EventPriority priority, EventId id, uint16_t dat
 // for legacy compatibility.
 #define event_raise(id, data) event_raise_priority(EVENT_PRIORITY_NORMAL, (id), (data))
 
+// Raises an event at default priority with 0 for data
+#define event_raise_no_data(id) event_raise((id), (0))
```
GeraldNDA, line 52: Nit: You don't need the brackets on the zero here.

libraries/ms-common/inc/ms_test_helpers.h
```diff
@@ -67,6 +68,9 @@
     TEST_ASSERT_EQUAL((e_data), event.data);                  \
   })
 
+#define MS_TEST_HELPER_ASSERT_NEXT_EVENT_ID(event, e_id) \
+  MS_TEST_HELPER_ASSERT_NEXT_EVENT((event), (e_id), (0))
```
GeraldNDA, line 72: Same here

projects/centre_console/src/fault_monitor.c
```diff
@@ -0,0 +1,50 @@
...
+#include "power_fsm.h"
+#include "status.h"
+#include "watchdog.h"
+
+#define NO_FAULT 0
+
+typedef struct FaultMonitorStorage {
+  uint16_t bps_fault_bitset;
+  WatchdogStorage watchdog_storage;
+} FaultMonitorStorage;
```
GeraldNDA, line 16: Shouldn't this be in the header, also what's the point of storing the fault_bitset?

projects/centre_console/src/power_fsm.c
```diff
@@ -1,86 +1,75 @@
...
+                             power_state_aux);
+  FSM_ADD_GUARDED_TRANSITION(CENTRE_CONSOLE_POWER_EVENT_CLEAR_FAULT, prv_guard_clear_fault,
+                             power_state_main);
+  FSM_ADD_TRANSITION(CENTRE_CONSOLE_POWER_EVENT_OFF, power_state_transitioning);
 }
 
 void prv_set_current_state(void *context, PowerState state) {
   PowerFsmStorage *storage = (PowerFsmStorage *)context;
   storage->current_state = state;
 }
```
GeraldNDA, line 58: I assume this might've made sense on the old code ... but it's literally one line so I'd say export this logic?

This function does seem useful if it stored the previous state (maybe even the destination state) while setting the current one imo, but you'd have to change your logic accordingly

projects/centre_console/src/power_off_sequence.c
```diff
@@ -59,6 +60,7 @@ static StatusCode prv_can_ack_everything_turned_off(CanMessageId msg_id, uint16_
 }
 
 static void prv_state_turn_off_everything(Fsm *fsm, const Event *e, void *context) {
+  printf("turning off everything\n");
```
GeraldNDA, line 63: Lingering debug?

projects/centre_console/src/fault_monitor.c
g-jessmuir: True we don't do anything with the bitset, so I just made it a static watchdog storage

projects/centre_console/src/power_off_sequence.c
g-jessmuir: whoopsie

PR #101, opened 2020-06-13
[SOFT-146] added resistance to thermistor

JarvisWeng:


g-jessmuir:
Can you remove unrelated files like the mppt tests? thanks

JarvisWeng:
> Can you remove unrelated files like the mppt tests? thanks

Ah, must have made a branch off the one I was working on.

projects/power_selection/src/resistance_to_temp.c
```diff
@@ -0,0 +1,50 @@
...
+  { 80.56, 1234 },    { 81.67, 1190 },    { 82.78, 1149 },    { 83.89, 1109 },
+  { 85.00, 1070 },    { 86.11, 1034 }
+};
+
+double resistance_to_temp(double resistance) {
+  // given resistance, use the table to find the temperature.
+  double temp = UINT16_MAX;
+
+  // find the first resistance value that is lower
+  for (int i = 0; i < 114; ++i) {
```
g-jessmuir, line 42: Please use #define for 114, I know it seems trivial but it just makes it that much easier to read. Also please put a quick comment defining the structure of the array, i.e. what's the first value at each index, what's the second. It wouldn't hurt to define an enum (local to the `.c` file) for this I think? Would improve clarity but I won't make you.

projects/power_selection/src/power_selection.c
```diff
@@ -36,8 +37,12 @@ static uint16_t s_aux_temp = 0;
 uint16_t prv_status_checker() {
   adc_read_raw(aux_channels[AUX_ADC_VOLT_CHANNEL], &s_aux_volt);
   adc_read_raw(aux_channels[AUX_ADC_TEMP_CHANNEL], &s_aux_temp);
-  // LOG_DEBUG("AUX Volatge Data: %d\n", aux_volt);
-  // LOG_DEBUG("AUX Temp Data: %d\n", aux_temp);
+  LOG_DEBUG("AUX Volatge Data: %d\n", s_aux_volt);
+  LOG_DEBUG("AUX Temp Voltage Data: %d\n", s_aux_temp);
+  double resistance = 33000.0 / (double)(s_aux_temp / 1000.0) - 10000;
```
g-jessmuir, line 42: It's not great to leave this here since you'll need it for tests, you could add the calculation as a #define in the header that takes in a parameter, e.g.
```
#define temp_to_res(r) 33000.0 / (double)((r) / 1000.0) - 10000
```
This way you can reuse it exactly in test.

projects/power_selection/src/resistance_to_temp.c
```diff
@@ -0,0 +1,50 @@
+#include "stdint-gcc.h"
```
g-jessmuir, line 1: why do you need to include this? can't you just include <stdint.h>?

projects/power_selection/src/resistance_to_temp.c
JarvisWeng: What do you mean by a local enum?

g-jessmuir: I think it would be more clear if you defined an enum `typedef enum { RES_TEMP_TABLE_TEMP = 0, RES_TEMP_TABLE_RES, NUM_RES_TEMP_TABLE_INDICES } ResTempTableIndex;`, then when you're referring to `resistance_temperature_in_C[i][1]` you can say `resistance_temperature_in_C[i][RES_TEMP_TABLE_RES` to indicate you're accessing the resistance in the table. I think this is more clear than just a 1 or a 0

JarvisWeng: ah i see. I'll do that

projects/power_selection/src/resistance_to_temp.c
```diff
@@ -0,0 +1,61 @@
...
+  // given resistance, use the table to find the temperature.
+  double temp = UINT16_MAX;
+
+  // checks the highest possible resistance value first
+  // then goes down until it finds the resistance that's lower than it
+  for (int i = 0; i < NUM_OF_RESISTANCES; ++i) {
+    // the higher the resistance, the lower the temperature
+    // so finds the first resistance value that is lower, else returns an impossible resistance
+    if (resistance_temperature_in_C[i][RES_TEMP_TABLE_RES] < resistance) {
+      temp = resistance_temperature_in_C[i][0];
```
g-jessmuir, line 55: can you use the enum value instead of 0 here

projects/power_selection/src/resistance_to_temp.c
```diff
@@ -0,0 +1,61 @@
...
+
+typedef enum {
+  RES_TEMP_TABLE_TEMP = 0,
+  RES_TEMP_TABLE_RES,
+  NUM_RES_TEMP_TABLE_INDICES
+} ResTempTableIndex;
+
+// from
+// http://2avrmz2nom8p47cc28p2743e-wpengine.netdna-ssl.com/wp-content/uploads/2010/11/Thermistor_10K-2.pdf
+const double resistance_temperature_in_C[NUM_OF_RESISTANCES][2] = {
```
g-jessmuir, line 13: can you use the enum value instead of 2 here? also please make this `static` and prefix the name to `s_`

PR #103, opened 2020-06-14
WIP: [SOFT-161] Remove the 'power_distribution' prefix

cathcai:
Removed "power_distribution" prefix in all files and renamed header dependencies appropriately within files.


projects/power_distribution/inc/can_rx_event_mapper_config.h
```diff
@@ -2,7 +2,7 @@
 
 // Standard configurations for power_distribution_can_rx_event_mapper.
```
copperium, line 3: Change the module name here too?

projects/power_distribution/inc/current_measurement_config.h
```diff
@@ -2,7 +2,7 @@
 
 // Standard hardware configurations for power_distribution_current_measurement.
```
copperium, line 3: Also change the module name here

projects/power_distribution/inc/gpio_config.h
```diff
@@ -2,7 +2,7 @@
 
 // Standard configurations for power_distribution_gpio.
```
copperium, line 3: Also change the module name here

projects/power_distribution/inc/publish_data_config.h
```diff
@@ -2,7 +2,7 @@
 
 // Standard configurations for power_distribution_publish_data.
```
copperium, line 3: The module name should also be changed here.

projects/power_distribution/src/gpio.c
```diff
@@ -1,5 +1,5 @@
-#include "power_distribution_gpio.h"
-#include "power_distribution_events.h"
+#include "gpio.h"
```
copperium, line 1: I don't think it's a good idea to rename `power_distribution_gpio` to just `gpio` since we already have a `gpio` library in ms-common, and the name conflicts. Perhaps change it to `pd_gpio` instead.

projects/power_distribution/src/gpio.c
```diff
@@ -1,5 +1,5 @@
...
+#include "events.h"
```
copperium, line 2: I would like to keep some form of a prefix on `events` since just `events.h` sounds like an ms-common library. `pd_events.h` would follow the naming convention of e.g. `mci_events.h`, `solar_events.h` in other projects.

g-jessmuir:
quick comment (I haven't looked at it yet) but this ticket number is deeeefinitely higher than 45 🤔🤔🤔

PR #106, opened 2020-06-19
Soft 184 spi functions temperature

AveryChiu64:
Here is the thought process
1. Send WRCOMM command to allow me to write to the COMM register with SPI 
2. Write 3 bytes of data to the COMM register that contains the pin number for the mux (note each byte is a duplicate of each other since WRCOMM forces me to send 3 bytes at once)
3. Send STCOMM command with SPI to send 3 bytes of data to the mux from the LTC6811 using SPI
4. Read data from the mux (a specific thermistor) on GPIO1 of the LTC6811


libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -217,7 +252,8 @@ StatusCode ltc_afe_impl_trigger_aux_conv(LtcAfeStorage *afe, uint8_t device_cell
   // we use GPIO2, GPIO3, GPIO4, GPIO5 to select which input to read
```
mitchellostler, line 252: Is this now resolved? `// TODO(SOFT-9): Update GPIO usage to match updated design`

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -104,6 +103,42 @@ static StatusCode prv_trigger_aux_adc_conversion(LtcAfeStorage *afe) {
...
+}
+
+static StatusCode prv_mux_enable_spi(LtcAfeStorage *afe) {
+  LtcAfeSettings *settings = &afe->settings;
+  // Setting sending STCOMM command
+  uint16_t stcomm = LTC6811_STCOMM_RESERVED;
+  uint8_t cmd[4] = { 0 };
+  prv_build_cmd(stcomm, cmd, SIZEOF_ARRAY(cmd));
+  prv_wakeup_idle(afe);
+  return spi_exchange(settings->spi_port, cmd, 4, NULL, 0);
```
GeraldNDA, line 139: So we def need to do the write 72 clock cycles thing now that I'm looking at it again. Not sure if we can just make sure to not use spi for that time or if there's something we need to do to get that to happen ...

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -104,6 +103,42 @@ static StatusCode prv_trigger_aux_adc_conversion(LtcAfeStorage *afe) {
...
+  }
+  LtcAfeSettings *settings = &afe->settings;
+  LtcAfeWriteCommRegPacket packet = { 0 };
+  // Build WRCOMM Command
+  prv_build_cmd(LTC6811_WRCOMM_RESERVED, packet.wrcomm, SIZEOF_ARRAY(packet.wrcomm));
+  // Write 3 bytes of data to the COMM registers
+  // We send the same data 3 times since using the WRCOMM command forces us to send 3 bytes
+  // However, the mux only needs a byte of data
+  packet.reg.icom0 = LTC6811_ICOM_CSBM_LOW;
+  packet.reg.d0 = device_cell - 1;
```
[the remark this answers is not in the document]
AveryChiu64: image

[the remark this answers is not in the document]
AveryChiu64: It's based on the truth table, I took another look and the pins aren't exactly the same as their binary representation. Basically what they do is they take the pin number -1 to get the byte of data that needs to be sent.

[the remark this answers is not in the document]
AveryChiu64: image

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -217,7 +252,8 @@ StatusCode ltc_afe_impl_trigger_aux_conv(LtcAfeStorage *afe, uint8_t device_cell
   // we use GPIO2, GPIO3, GPIO4, GPIO5 to select which input to read
   // corresponding to the binary representation of the cell
   prv_write_config(afe, (device_cell << 4) | LTC6811_GPIO1_PD_OFF);
```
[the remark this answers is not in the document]
AveryChiu64: Oh yeah, we should probably turn off all the GPIOs, I just need GPIO 1 to be able to read data from the mux though.

libraries/ms-drivers/inc/ltc6811.h
```diff
@@ -7,6 +7,8 @@
 
 #define LTC6811_CELLS_IN_REG 3
 #define LTC6811_GPIOS_IN_REG 3
+#define ADS1259_SHIFT_REG_SIZE 8
+#define ADS1259_NUM_PINS 32
```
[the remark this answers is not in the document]
AveryChiu64: I mean I could make a ADS1259.h file but it might not have much in it, but yeah it's not really the best place to put it. Any suggestions?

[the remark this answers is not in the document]
GeraldNDA: I'd say either move it to its own file or name it to point out that it's for the aux. Either should be good.

libraries/ms-drivers/src/ltc_afe_impl.c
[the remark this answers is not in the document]
GeraldNDA: IMO, this is another good place to use to a struct probably to be explicit. Also I believe the `device_cell` is already 0 indexed from the caller side so you don't need to subtract 1 here.

libraries/ms-drivers/inc/ltc6811.h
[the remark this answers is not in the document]
g-jessmuir: cmon, one of you must’ve noticed it’s the adg731 not the ads1259 we’re using here... 😂

libraries/ms-drivers/inc/ltc6811.h
```diff
@@ -66,13 +68,36 @@ typedef struct {
 } _PACKED LtcAfeConfigRegisterData;
 static_assert(sizeof(LtcAfeConfigRegisterData) == 6, "LtcAfeConfigRegisterData must be 6 bytes");
 
+// COMM Register, refer to LTC6803 datasheet page 31, Table 15
+typedef struct {
+  // COMM0
+  uint8_t icom0 : 4;
+  uint8_t d0 : 8;
```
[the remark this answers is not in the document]
AveryChiu64: static_assert(sizeof(LtcAfeCommRegisterData) == 6, "LtcAfeCommRegisterData must be 6 bytes");

The above assertion will fail when I remove the 8

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -104,6 +103,42 @@ static StatusCode prv_trigger_aux_adc_conversion(LtcAfeStorage *afe) {
...
+  packet.reg.d0 = device_cell - 1;
+  packet.reg.fcom0 = LTC6811_FCOM_CSBM_LOW;
+
+  packet.reg.icom1 = LTC6811_ICOM_CSBM_LOW;
+  packet.reg.d1 = device_cell - 1;
+  packet.reg.fcom1 = LTC6811_FCOM_CSBM_LOW;
+
+  packet.reg.icom2 = LTC6811_ICOM_CSBM_LOW;
+  packet.reg.d2 = device_cell - 1;
+  packet.reg.fcom2 = LTC6811_FCOM_CSBM_HIGH;
```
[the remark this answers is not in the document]
AveryChiu64: image
Hmm I see, I think I can also just pull it HIGH and itll stop

libraries/ms-drivers/inc/ltc6811.h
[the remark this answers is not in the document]
GeraldNDA: Ah ic, my b https://stackoverflow.com/questions/25822679/packed-bit-fields-in-c-structures-gcc

libraries/ms-drivers/inc/ltc6811.h
```diff
@@ -66,13 +69,43 @@ typedef struct {
...
+  uint8_t wrcomm[4];
+  LtcAfeCommRegisterData reg;
+  uint8_t pec;
+} _PACKED LtcAfeWriteCommRegPacket;
+
+// STMCOMM + clock cycles
+typedef struct {
+  uint8_t stcomm[4];
+  uint8_t clk[3];
+} _PACKED LtcAfeSendCommRegPacket;
```
GeraldNDA, line 106: I would say make 2 defines for the amount of bytes for each of these (a define for the size of a command would be nice because we just write the magic number everywhere) so that you don't have to use SIZEOF_ARRAY

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -104,6 +103,41 @@ static StatusCode prv_trigger_aux_adc_conversion(LtcAfeStorage *afe) {
...
+  LtcAfeWriteCommRegPacket packet = { 0 };
+  // Build WRCOMM Command
+  prv_build_cmd(LTC6811_WRCOMM_RESERVED, packet.wrcomm, SIZEOF_ARRAY(packet.wrcomm));
+  // Write 3 bytes of data to the COMM registers
+  // We send the a byte and then we send CSBM_HIGH to
+  // release the SPI port
+  packet.reg.icom0 = LTC6811_ICOM_CSBM_LOW;
+  packet.reg.d0 = device_cell;
+  packet.reg.fcom0 = LTC6811_FCOM_CSBM_HIGH;
+  uint16_t comm_pec = crc15_calculate((uint8_t *)&packet.reg, sizeof(LtcAfeCommRegisterData) / 3);
```
GeraldNDA, line 120: From the datasheet:
> Similar to the I2C master, if ICOMn[3:0] specified a CSBM HIGH or a NO TRANSMIT condition, the CSBM, SCKM and SDIOM lines of the SPI master are released and the rest of the data in the word is ignored.

You def. don't want to leave icom as all zeros because that'll configure it to use i2c and then send stuff you don't want. If you don't want to send data for the following 2 bytes then either set it the icon fields to CSBM HIGH or NO TRANSMIT. I'd prefer the later because it's a bit more explicit

You also want to calculate pec on the whole register, otherwise the chip will get a different pec result ...

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -104,6 +103,41 @@ static StatusCode prv_trigger_aux_adc_conversion(LtcAfeStorage *afe) {
...
+  return spi_exchange(settings->spi_port, (uint8_t *)&packet, sizeof(LtcAfeWriteCommRegPacket),
+                      NULL, 0);
+}
+
+static StatusCode prv_mux_send_comm_register(LtcAfeStorage *afe) {
+  LtcAfeSettings *settings = &afe->settings;
+  LtcAfeSendCommRegPacket packet = { 0 };
+  // Build STCOMM command
+  prv_build_cmd(LTC6811_STCOMM_RESERVED, packet.stcomm, SIZEOF_ARRAY(packet.stcomm));
+  for (uint8_t i = 0; i < SIZEOF_ARRAY(packet.clk); i++) {
```
GeraldNDA, line 132: See comments about using defines instead of sizeof here.

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -130,7 +164,7 @@ static StatusCode prv_write_config(LtcAfeStorage *afe, uint8_t gpio_enable_pins)
     config_packet.devices[curr_device].reg.undervoltage = undervoltage;
     config_packet.devices[curr_device].reg.overvoltage = overvoltage;
 
-    // GPIO5, ..., GPIO2 are used to MUX data
+    // GPIO 1 is used to read data from the mux
```
GeraldNDA, line 167: ~might wanna remove the comment completely~

libraries/ms-drivers/src/ltc_afe_impl.c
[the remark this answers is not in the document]
GeraldNDA: why's this still here ... you probably want to do the same "all gpios off" as with in init

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -211,13 +243,10 @@ StatusCode ltc_afe_impl_trigger_cell_conv(LtcAfeStorage *afe) {
...
-  // TODO(SOFT-9): Update GPIO usage to match updated design
-
-  // configure the mux to read from cell
-  // we use GPIO2, GPIO3, GPIO4, GPIO5 to select which input to read
-  // corresponding to the binary representation of the cell
+  // We use GPIO 1 to read data from the mux
   prv_write_config(afe, (device_cell << 4) | LTC6811_GPIO1_PD_OFF);
-
+  prv_mux_write_comm_register(afe, device_cell);
+  prv_mux_send_comm_register(afe);
```
GeraldNDA, line 249: imo you might wanna prefix these with `prv_aux_` or `prv_aux_spi_` instead of `prv_mux_`. The only mux specific thing in this code is the content of the message

GeraldNDA: @g-jessmuir can prolly comment better if this is a good idea, but we may want to generalize write_comm register to take 3 bytes and a number which indicates how many bytes you want to send. That way this code is entirely decoupled from the mux beyond this line.

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -211,13 +244,11 @@ StatusCode ltc_afe_impl_trigger_cell_conv(LtcAfeStorage *afe) {
...
-  // TODO(SOFT-9): Update GPIO usage to match updated design
-
-  // configure the mux to read from cell
-  // we use GPIO2, GPIO3, GPIO4, GPIO5 to select which input to read
-  // corresponding to the binary representation of the cell
-  prv_write_config(afe, (device_cell << 4) | LTC6811_GPIO1_PD_OFF);
-
+  uint8_t gpio_bits =
+      LTC6811_GPIO1_PD_OFF | LTC6811_GPIO3_PD_OFF | LTC6811_GPIO4_PD_OFF | LTC6811_GPIO5_PD_OFF;
+  prv_write_config(afe, (device_cell << 4) | gpio_bits);
```
GeraldNDA, line 249: You should remove the `device_cell << 4` which was there to toggle GPIO2-5 based on the cell to read

libraries/ms-drivers/src/ltc_afe_impl.c
```diff
@@ -83,25 +83,60 @@ static StatusCode prv_trigger_adc_conversion(LtcAfeStorage *afe) {
...
+  LtcAfeWriteCommRegPacket packet = { 0 };
+  // Build WRCOMM Command
+  prv_build_cmd(LTC6811_WRCOMM_RESERVED, packet.wrcomm, LTC6811_CMD_SIZE);
+  // Write 3 bytes of data to the COMM registers
+  // We send the a byte and then we send CSBM_HIGH to
+  // release the SPI port
+  packet.reg.icom0 = LTC6811_ICOM_CSBM_LOW;
+  packet.reg.d0 = device_cell;
+  packet.reg.fcom0 = LTC6811_FCOM_CSBM_HIGH;
+  packet.reg.icom1 = LTC6811_ICOM_NO_TRANSMIT;
```
GeraldNDA, line 120: you also need to write icom2 as well

libraries/ms-drivers/inc/ltc6811.h
```diff
@@ -4,10 +4,18 @@
...
 #define LTC6811_GPIOS_IN_REG 3
 
+// used for the external mux (ADG731) connected to the AFE
+#define AUX_ADG731_NUM_PINS 32
+
+// Size of command code + PEC
+#define LTC6811_CMD_SIZE 4
+
+// 3 bytes are required to send 24 clock cycles with our SPI driver for the STCOMM command
+#define LTC6811_24_CLOCK_CYCLES_STCOMM_BYTES 3
```
GeraldNDA, line 17: Current name is fine but I feel like something like `_COMM_REG_SIZE` or `_NUM_COMM_REG_BYTES`  woulda been better since that's what it is (and then in the comment say what it's used for)

g-jessmuir:
> The generalization comment can probably be pushed to another issue tbh

Agreed, I think it's likely if we use these again it'll be with the mux anyways. Should just be small edits to update it.

libraries/ms-drivers/inc/ltc6811.h
AveryChiu64: This isn't the COMM reg's size though? This is just for the NULL bytes when I send the clock cycles

GeraldNDA: You use it to send 3 NULL bytes but the reason it's 3 is because the COMM register sends 3 bytes of data. the second suggestion is probably less confusing I guess, but up to you whether you want to change it.

PR #107, opened 2020-06-19
[SOFT-262] Implement charging manager and tests

g-jessmuir:
Charging manager manages the charge state to ensure that
charger can't start charging the car unless we're in park.


projects/centre_console/inc/charging_manager.h
```diff
@@ -11,6 +12,6 @@ typedef enum {
   NUM_CHARGING_STATES
 } ChargingState;
 
-StatusCode init_charging_manager(void);
+StatusCode init_charging_manager(const DriveState *drive_state);
 
 ChargingState *get_global_charging_state(void);
```
GeraldNDA, line 17: This was here before, but why is this by a pointer instead of just the charging state ...

projects/centre_console/test/test_charging_manager.c
```diff
@@ -0,0 +1,82 @@
...
+  event_queue_init();
+  can_init(&s_can_storage, &s_can_settings);
+  TEST_ASSERT_OK(init_charging_manager(&s_drive_state));
+}
+
+void teardown_test(void) {}
+
+void test_get_global_charging_state(void) {
+  ChargingState *charging_state = get_global_charging_state();
+  TEST_ASSERT(*charging_state == CHARGING_STATE_NOT_CHARGING);
```
GeraldNDA, line 37: Shouldn't these be `TEST_ASSERT_EQUAL` instead?

projects/centre_console/inc/charging_manager.h
g-jessmuir: Good question, at this point it's just for matching convention with the other centre console modules. May as well be consistent?

projects/centre_console/test/test_charging_manager.c
g-jessmuir: true

projects/centre_console/inc/charging_manager.h
GeraldNDA: idk, exporting internal state pointers is fine (although not ideal) if used in the test context. Since this is a function that other modules actually use, I feel like they shouldn't be exporting access to the internal state if you don't expect them to change it

g-jessmuir: Fair enough. Fixed, involved some refactoring in main event generator.

PR #108, opened 2020-06-20
Soft 213 mppt solar module

JarvisWeng:


projects/solar/src/mppt.c
```diff
@@ -0,0 +1,42 @@
...
+#include "mux.h"
+#include "spv1020_mppt.h"
+
+// will get the address and size of the demultiplexer
+static MuxAddress s_mux_address = {
+  .bit_width = 3,
+  .sel_pins[0] = { .port = GPIO_PORT_B, .pin = 3 },
+  .sel_pins[1] = { .port = GPIO_PORT_B, .pin = 4 },
+  .sel_pins[2] = { .port = GPIO_PORT_B, .pin = 5 },
+  .mux_enable_pin = { .port = GPIO_PORT_A, .pin = 1 },  // need to check hardward for correct pin
```
copperium, line 11: HW hasn't actually given us an enable pin - we'll need to set this to some pin that isn't being used, like PB1. (The unused pin should be #define'd so we can move it to a config file later and use the same unused pin everywhere we need one.) Also, I believe you'll also have to fill in the `mux_output_pin` parameter with an unused pin. (A better solution of course would be to refactor the `mux` driver so it doesn't need these pins, but alas that's a problem for another ticket.)

projects/solar/src/mppt.c
```diff
@@ -0,0 +1,42 @@
+#include "gpio.h"
+#include "mux.h"
+#include "spv1020_mppt.h"
+
+// will get the address and size of the demultiplexer
+static MuxAddress s_mux_address = {
+  .bit_width = 3,
+  .sel_pins[0] = { .port = GPIO_PORT_B, .pin = 3 },
+  .sel_pins[1] = { .port = GPIO_PORT_B, .pin = 4 },
+  .sel_pins[2] = { .port = GPIO_PORT_B, .pin = 5 },
```
copperium, line 10: Damn, I had no idea that this syntax works. I'd probably prefer these GPIO addresses to be #define'd so that we can move them easily to a config file with all the hardware details, but this is fine.

projects/solar/src/mppt.c
```diff
@@ -0,0 +1,42 @@
...
+  .mux_enable_pin = { .port = GPIO_PORT_A, .pin = 1 },  // need to check hardward for correct pin
+};
+
+StatusCode mppt_shut(SpiPort port, uint8_t pin) {
+  status_ok_or_return(mux_set(&s_mux_address, pin));
+  return spv1020_shut(port);
```
copperium, line 16: Like I explained on Slack, since hardware hasn't given us an enable pin, to set the CS line for SPI low and then high again for the specific SPV1020 we're talking to, we have to set the mux to the `pin` before calling the `spv1020` function, then set the mux to an output we're not using (like output 7). That way, the SPV1020 sees the CS line go low, then high again at the end.

Thus, you should add a `#define DISCONNECTED_MUX_OUTPUT 7`, then call `mux_set(&s_mux_address, DISCONNECTED_MUX_OUTPUT)` at the end of each of the `mppt_*` functions.

Also, to deal with status codes: to ensure that the final `mux_set` call you're adding is always called even if the `spv1020` call fails, you should save the output of the `spv1020_*` call in a `status` variable, wrap the final `mux_set` call in `status_ok_or_return`, then return the `status` variable you saved. That way the final `mux_set` is always executed and we never end up with the CS line low for a long period of time.

projects/solar/src/mppt.c
```diff
@@ -0,0 +1,42 @@
...
+
+// will get the address and size of the demultiplexer
+static MuxAddress s_mux_address = {
+  .bit_width = 3,
+  .sel_pins[0] = { .port = GPIO_PORT_B, .pin = 3 },
+  .sel_pins[1] = { .port = GPIO_PORT_B, .pin = 4 },
+  .sel_pins[2] = { .port = GPIO_PORT_B, .pin = 5 },
+  .mux_enable_pin = { .port = GPIO_PORT_A, .pin = 1 },  // need to check hardward for correct pin
+};
+
```
copperium, line 13: I think you're missing an `mppt_init` function that calls `mux_init`.

projects/solar/test/test_mppt.c
```diff
@@ -0,0 +1,84 @@
...
+#define TEST_SPI_PORT SPI_PORT_2
+
+void setup_test() {
+  gpio_init();
+  SpiSettings spi_settings = { .baudrate = 60000,
+                               .mode = SPI_MODE_3,
+                               .mosi = { .port = GPIO_PORT_B, 15 },
+                               .miso = { .port = GPIO_PORT_B, 14 },
+                               .sclk = { .port = GPIO_PORT_B, 13 },
+                               .cs = { .port = GPIO_PORT_B, 12 } };
```
copperium, line 17: `make format` will probably reformat this - Travis did something strange with this PR and didn't run.

projects/solar/test/test_mppt.c
```diff
@@ -0,0 +1,84 @@
...
+  spi_init(TEST_SPI_PORT, &spi_settings);
+}
+
+void teardown_test() {}
+
+// Test that we can successfully send the SHUT and Turn ON commands.
+void test_shut_and_turn_on(void) {
```
copperium, line 24: The tests for the range of the `pin` parameter are good :) To test more `mppt`-specific functionality, maybe you could mock `mux_set` (see `test_sense.c` for how to mock a function) and make sure that it's called correctly on each `mppt` function call?

projects/solar/test/test_mppt.c
```diff
@@ -0,0 +1,118 @@
...
+  TEST_ASSERT_EQUAL(s_times_mux_set_called, 12);
+  TEST_ASSERT_NOT_OK(mppt_turn_on(TEST_SPI_PORT, (uint8_t)0b1010));
+  TEST_ASSERT_EQUAL(s_times_mux_set_called, 12);
+  TEST_ASSERT_OK(mppt_turn_on(TEST_SPI_PORT, (uint8_t)0b0110));
+  TEST_ASSERT_EQUAL(s_times_mux_set_called, 14);
+}
+
+// Test that we can successfully read a current and it's in the stated range.
+void test_read_current(void) {
+  s_times_mux_set_called = 0;
```
copperium, line 64: Nit: can you move the `s_times_mux_set_called = 0` to `setup_test`?

projects/solar/src/mppt.c
```diff
@@ -0,0 +1,72 @@
+#include "gpio.h"
+#include "mux.h"
+#include "spv1020_mppt.h"
+
+#define MUX_ENABLE \
+  { .port = GPIO_PORT_B, .pin = 1 }  // need to check hardward for correct pin
+#define MUX_OUTPUT \
+  { .port = GPIO_PORT_B, .pin = 1 }  // need to check hardward for correct pin
```
copperium, line 8: Nit: maybe just have a comment saying "this pin is unused" or something like that

PR #112, opened 2020-06-24
[SOFT-235] Add DEFINE make argument, use in Travis

copperium:
This sets the Travis log level to to `LOG_LEVEL_WARN`. Also, make is voodoo.

Also adds a general `DEFINE` make argument that can be used to define arbitrary preprocessor symbols from the command line. We do some funky make witchcraft to make it so that we rebuild all dependencies when the value of `DEFINE` changes but not when it doesn't. (The witchcraft involved is writing the value of `DEFINE` to `build/dep_vars/[x86/stm32f0xx]/DEFINE`, then depending on that file and rewriting the file, triggering a rebuild, if the file's contents don't match the variable.)


Makefile
```diff
@@ -101,6 +105,18 @@ define find_in
 $(foreach folder,$(1),$(wildcard $(folder)/$(2)))
 endef
 
+# Hack to enable depending on a variable - https://stackoverflow.com/a/26147844
+# $(eval $(call dependable_var,variable))
+DEP_VARS :=
```
cathcai, line 110: Can these lines not be combined into `DEP_VARS := $(DEP_VAR_DIR)/$1` ?

PR #113, opened 2020-06-25
[soft-263] - Quick linting and format

tristanle22:
Faster lint/format:

New:

- lint_quick: run lint.py only on changed/new files
- format_quick: run clang-format only changed/new files
- Fix minor bug in make file that didn't use the input projects/libraries in lint and format
- Replace full lint/format with quick lint/format at commit

Thanks to @g-jessmuir and @GeraldNDA for the idea!


Makefile
```diff
@@ -129,29 +132,41 @@ $(foreach proj,$(VALID_PROJECTS),$(call include_proj,$(proj)))
 
 IGNORE_CLEANUP_LIBS := CMSIS FreeRTOS STM32F0xx_StdPeriph_Driver unity FatFs
 FIND_PATHS := $(addprefix -o -path $(LIB_DIR)/,$(IGNORE_CLEANUP_LIBS))
-FIND := find $(PROJ_DIR) $(LIB_DIR) \
+FIND := find $(PROJECT_DIR) $(LIBRARY_DIR) \
 			  \( $(wordlist 2,$(words $(FIND_PATHS)),$(FIND_PATHS)) \) -prune -o \
 				-iname "*.[ch]" -print
+FIND_MOD_NEW := git status --porcelain | grep -v -e "^D" -e "^ D" | grep -e ".*\.c" -e ".*\.h" | cut -c 4-
```
GeraldNDA, line 138: Any reason you chose to do this instead of `git diff --name-only --diff-filter=ACMRT` like `test_format`?

hooks/pre-commit
```diff
@@ -10,4 +10,4 @@ if [[ $GIT_AUTHOR_NAME == 'vagrant' || $GIT_AUTHOR_EMAIL == 'vagrant@midsunbox'
   echo "You didn't set up your git config! Please follow the Software 101 modules on Confluence."
   echo "https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/60260353/Module+1+Setup"
   exit 1
-fi
+fi
```
GeraldNDA, line 13: probably not on purpose?

Makefile
tristanle22: Actually I didn't know we have something like that already. The git diff command seems cleaner and already working so I'll switch to that. Thanks for pointing this out!

tristanle22:
@g-jessmuir I created a local test branch, then modify/add/delete .c .h and other files. Run lint_quick and format_quick => verified only works on .c and .h files, deleted and other files are ignored. I verified that lint_quick and format_quick works at commit too.

PR #114, opened 2020-06-26
[SOFT-208] sense_current module for solar

copperium:
`sense_current` is an "implementation" of `sense` that reads from the current sense MCP3427 and dumps it in `data_store` as `DATA_POINT_CURRENT`. Also, I added a bit to `mcp3427_adc` so that the ADC results are always scaled such that the LSB is 62.5uV, and I added a `data_store_is_set` function (and corresponding `data_store_init`) because I realized that some data points might not always be set.


projects/solar/test/test_data_store.c
```diff
@@ -6,47 +6,85 @@
...
+#define VALID_TEST_VALUE 0x1337
+
 void setup_test(void) {
   event_queue_init();
 }
 void teardown_test(void) {}
 
 // Test that we can set and get a value from the data store.
 void test_data_store_set_and_get_basic(void) {
+  TEST_ASSERT_OK(data_store_init());
```
g-jessmuir, line 19: Is there a reason you don't just add `TEST_ASSERT_OK(data_store_init())` to `setup_test()`? I dunno if that's not best practice but it workes, I'm fine with either

projects/solar/test/test_data_store.c
```diff
@@ -6,47 +6,85 @@
...
+
+// Test that |data_store_get_is_set| responds correctly on a null pointer.
+void test_data_store_get_is_set_null_pointer(void) {
+  TEST_ASSERT_OK(data_store_init());
+  TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, data_store_get_is_set(VALID_TEST_DATA_POINT, NULL));
 }
 
 // Test that |data_store_done| correctly raises DATA_READY_EVENTs.
 void test_data_store_done_raises_events(void) {
+  TEST_ASSERT_OK(data_store_init());
```
g-jessmuir, line 87: nit: technically you don't need this here, do you?

projects/solar/src/sense_current.c
```diff
@@ -0,0 +1,63 @@
...
+#include "mcp3427_adc.h"
+#include "sense.h"
+#include "solar_config.h"
+#include "solar_events.h"
+
+// we only use one channel on the MCP3427
+#define CURRENT_MCP3427_CHANNEL MCP3427_CHANNEL_1
+
+// after this many faults in a row, we raise a fault event
+#define MAX_CONSECUTIVE_MCP3427_FAULTS 3
```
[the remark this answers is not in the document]
copperium: My philosophy was originally that solar config holds the hardware settings but I think it makes more sense as a general holder for settings for the project. I think it makes sense to move this to solar config.

projects/solar/src/solar_config.c
```diff
@@ -0,0 +1,16 @@
+#include "solar_config.h"
+#include "solar_events.h"
+
+#define SOLAR_MCP3427_CONVERSION_MODE MCP3427_CONVERSION_MODE_CONTINUOUS
+#define SOLAR_MCP3427_SAMPLE_RATE MCP3427_SAMPLE_RATE_12_BIT
```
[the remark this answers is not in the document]
copperium: The philosophy was that these settings are probably going to be reused for the rest of the MCP3427s in the project, but the other settings (pins, I2C port, amp gain (maybe)) will vary by MCP3427, so they shouldn't be #define'd. The data trigger/ready events are already enum values so it'd be redundant to define them again.

projects/solar/test/test_data_store.c
copperium: Probably something weird would happen if an assertion failed in `setup_test`, that's why I didn't put it there

projects/solar/test/test_data_store.c
copperium: Technically no, but it's an implementation detail of the module that `data_store_done` doesn't depend on `data_store_init`, so I prefer not to depend on that.

PR #116, opened 2020-06-28
Soft 196 adt7476a

Max-MZ:
Dealt with comments from previous draft pr, changed functions to adapt to new driver.


libraries/ms-drivers/inc/adt_7476a_fan_controller.h
```diff
@@ -0,0 +1,43 @@
...
+
+// probably a better name for these
+typedef enum { FAN_1 = 0, FAN_2, FAN_3, FAN_4 } FAN_NUM;
+
+// Initialize the Adt7476a with the given settings; the select pin is an STM32 GPIO pin.
+StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings);
+
+// Translate and write the new speed
+StatusCode adt7476a_set_speed(I2CPort port, uint8_t speed, uint8_t FAN_PWM_ADDR,
+                              uint8_t ADR7470_I2C_ADDRESS);
```
g-jessmuir, line 43: nit: the parameters should just be in lowercase, also what the heck is an ADR 7470 🤔🤔

libraries/ms-drivers/inc/adt_7476a_fan_controller.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "soft_timer.h"
```
g-jessmuir, line 8: These headers should be included in the .c file rather than the .h file (except the ones you need like gpio and i2c) also why do you need the adc header?

libraries/ms-drivers/inc/adt_7476a_fan_controller.h
```diff
@@ -0,0 +1,43 @@
...
+} Adt7476aStorage;
+
+// probably a better name for these
+typedef enum { FAN_1 = 0, FAN_2, FAN_3, FAN_4 } FAN_NUM;
+
+// Initialize the Adt7476a with the given settings; the select pin is an STM32 GPIO pin.
+StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings);
+
+// Translate and write the new speed
+StatusCode adt7476a_set_speed(I2CPort port, uint8_t speed, uint8_t FAN_PWM_ADDR,
```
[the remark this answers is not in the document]
Max-MZ: Since we have 2 PWM's, it might make sense just to have an enum of  `FAN_GROUP_1` and `FAN_GROUP_2`

libraries/ms-drivers/inc/adt_7476a_fan_controller.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+// Requires interrupts, soft timers, GPIO, and GPIO interrupts to be initialized
+
+typedef void (*Adt7476ADataCallback)(void *context);
```
g-jessmuir, line 5: minor nitpick: This should be `Adt7476aDataCallback` rather than `Adt7476ADataCallback` to match the convention you set with the storage and settings type names

PR #117, opened 2020-06-30
[SOFT-197] Update killswitch to match new architecture

g-jessmuir:
This commit updates BPS and killswitch to match the new architecture
as well as implements tests.


projects/bms_carrier/src/killswitch.c
```diff
@@ -1,34 +1,35 @@
...
   GpioState state = NUM_GPIO_STATES;
   gpio_get_state(address, &state);
-  if (state == GPIO_STATE_LOW) {
-    // Falling edge - killswitch was hit
-    battery_heartbeat_raise_fault(storage, EE_BATTERY_HEARTBEAT_FAULT_SOURCE_KILLSWITCH);
-  } else {
-    // Rising edge - killswitch was released
-    battery_heartbeat_clear_fault(storage, EE_BATTERY_HEARTBEAT_FAULT_SOURCE_KILLSWITCH);
-  }
+  bool clear = state == GPIO_STATE_LOW ? true : false;
```
GeraldNDA, line 10: Is this pattern? Because simply `clear = (state == GPIO_STATE_LOW);` should be good. I would also maybe suggest putting it directly into the below line, but I like this verbosity.

projects/bms_carrier/test/test_bps_heartbeat.c
```diff
@@ -0,0 +1,117 @@
...
+#include "delay.h"
+#include "event_queue.h"
+#include "exported_enums.h"
+#include "interrupt.h"
+#include "log.h"
+#include "ms_test_helpers.h"
+#include "soft_timer.h"
+
+#undef BPS_HB_FREQ_MS
+#define BPS_HB_FREQ_MS 50
```
GeraldNDA, line 16: Why?

projects/bms_carrier/test/test_bps_heartbeat.c
```diff
@@ -0,0 +1,117 @@
...
+  *ack_reply = s_ack_status;
+
+  CAN_UNPACK_BPS_HEARTBEAT(msg, &s_bps_bitmask);
+
+  return s_ack_status == CAN_ACK_STATUS_OK ? STATUS_CODE_OK : STATUS_CODE_TIMEOUT;
+}
+
+static uint8_t s_fault_bps_bitmask = 0;
+static bool s_fault_bps_clear = false;
+
```
GeraldNDA, line 48: Should probz move these above `prv_hb_rx`

projects/bms_carrier/test/test_bps_heartbeat.c
```diff
@@ -0,0 +1,117 @@
...
+  s_hb_count = 0;
+  s_bps_bitmask = 0;
+  s_ack_status = NUM_CAN_ACK_STATUSES;
+  memset(&s_storage, 0, sizeof(s_storage));
+  s_storage.ack_devices = SYSTEM_CAN_DEVICE_BMS_CARRIER;
+  s_fault_bps_bitmask = 0;
+  s_fault_bps_clear = false;
+  interrupt_init();
+  soft_timer_init();
+  event_queue_init();
```
GeraldNDA, line 65: Should probably move these to the top or ... use even better maybe use `initialize_can_and_dependencies`

projects/bms_carrier/test/test_bps_heartbeat.c
```diff
@@ -0,0 +1,117 @@
...
+  TEST_ASSERT_OK(bps_heartbeat_init(&s_storage, BPS_HB_FREQ_MS));
+  MS_TEST_HELPER_CAN_TX_RX_WITH_ACK(BMS_CAN_EVENT_TX, BMS_CAN_EVENT_RX);
+  TEST_ASSERT_EQUAL(1, s_hb_count);
+}
+
+void test_hb_repeats(void) {
+  s_ack_status = CAN_ACK_STATUS_OK;
+  TEST_ASSERT_OK(bps_heartbeat_init(&s_storage, BPS_HB_FREQ_MS));
+  MS_TEST_HELPER_CAN_TX_RX_WITH_ACK(BMS_CAN_EVENT_TX, BMS_CAN_EVENT_RX);
+  delay_ms(BPS_HB_FREQ_MS + 3);
```
GeraldNDA, line 82: lowkey we should a define for this `3` but uh, should be fine

projects/bms_carrier/test/test_killswitch.c
```diff
@@ -1,6 +1,70 @@
...
+  *state = s_gpio_state_get;
+  return STATUS_CODE_OK;
+}
+
+StatusCode TEST_MOCK(debouncer_init_pin)(DebouncerStorage *debouncer, const GpioAddress *address,
+                                         GpioItCallback callback, void *context) {
+  s_registered_callback = callback;
+  s_registered_context = context;
+  return STATUS_CODE_OK;
+}
```
GeraldNDA, line 38: why do you have to mock this? If it's because an x86 is missing, we should probably add that (or add a ticket and TODO works too)

projects/bms_carrier/src/killswitch.c
```diff
@@ -1,34 +1,35 @@
...
 }
 
-StatusCode killswitch_init(KillswitchStorage *storage, const GpioAddress *killswitch,
-                           BatteryHeartbeatStorage *battery_heartbeat) {
-  // Force update
-  prv_killswitch_handler(killswitch, battery_heartbeat);
-  return debouncer_init_pin(&storage->debouncer, killswitch, prv_killswitch_handler,
-                            battery_heartbeat);
-}
+StatusCode killswitch_init(DebouncerStorage *storage) {
```
GeraldNDA, line 14: Lowkey, I think calling it `_init` is misleading. This module only does one task ... maybe `_setup` or smtg? I'm not sure what we normally do.

projects/bms_carrier/src/killswitch.c
g-jessmuir: Updated to `bool clear = (state == GPIO_STATE_LOW)`

projects/bms_carrier/test/test_bps_heartbeat.c
g-jessmuir: im dum

projects/bms_carrier/test/test_bps_heartbeat.c
g-jessmuir: moved to aggregate static declarations

projects/bms_carrier/test/test_bps_heartbeat.c
g-jessmuir: used `initialize_can_and_dependencies`

projects/bms_carrier/test/test_bps_heartbeat.c
g-jessmuir: defined `BUFFER_TIME_MS 3`

projects/bms_carrier/test/test_killswitch.c
g-jessmuir: I have no idea why I did this, just removed this mock. Debouncer is platform agnostic.

projects/bms_carrier/src/killswitch.c
g-jessmuir: I think the pattern is to just call anything init. It's initializing gpio pins, so is it really that wrong to call it the init function?

GeraldNDA: lowkey, but it's a nit so up to you. Typically w/ this pattern you init, and then it either does things in the background or you can interact with it. In this case, despite doing inits, it really only does one thing and then exists.

PR #118, opened 2020-07-01
[SOFT-209] sense_mcp3427 solar module

copperium:
This is a general sense module for all sensing from MCP3427s. It therefore supersedes `sense_current`, which I deleted. Also added a config for `sense_mcp3427` in `solar_config`. Since it depends on the number of MPPTs there are (5 or 6), there's now a function in `solar_config` for getting the correct config for the board type.


projects/solar/test/test_sense_mcp3427.c
```diff
@@ -0,0 +1,339 @@
...
+  { I2C_PORT_2, MCP3427_PIN_STATE_LOW, MCP3427_PIN_STATE_FLOAT },
+  { I2C_PORT_2, MCP3427_PIN_STATE_LOW, MCP3427_PIN_STATE_HIGH },
+  { I2C_PORT_2, MCP3427_PIN_STATE_FLOAT, MCP3427_PIN_STATE_LOW },
+  { I2C_PORT_2, MCP3427_PIN_STATE_FLOAT, MCP3427_PIN_STATE_HIGH },
+  { I2C_PORT_2, MCP3427_PIN_STATE_HIGH, MCP3427_PIN_STATE_LOW },
+  { I2C_PORT_2, MCP3427_PIN_STATE_HIGH, MCP3427_PIN_STATE_FLOAT },
+  { I2C_PORT_2, MCP3427_PIN_STATE_HIGH, MCP3427_PIN_STATE_HIGH },
+};
+
+// Helper function to get SenseMcp3427Settings that cover the max number of MCP3427s
```
GeraldNDA, line 65: I like the verbosity of this, but I think a simple nested for (and remove the wanted cases) would've been alot simpler ...

projects/solar/test/test_sense_mcp3427.c
```diff
@@ -0,0 +1,339 @@
...
+static void prv_get_max_mcp3427s_settings(SenseMcp3427Settings *settings) {
+  for (uint8_t i = 0; i < MAX_SOLAR_MCP3427; i++) {
+    settings->mcp3427s[i].mcp3427_settings = s_test_mcp3427_settings;
+    settings->mcp3427s[i].mcp3427_settings.port = s_mcp3427_address_combos[i].port;
+    settings->mcp3427s[i].mcp3427_settings.addr_pin_0 = s_mcp3427_address_combos[i].addr_pin_0;
+    settings->mcp3427s[i].mcp3427_settings.addr_pin_1 = s_mcp3427_address_combos[i].addr_pin_1;
+    settings->mcp3427s[i].data_point = TEST_DATA_POINT + i;
+  }
+  settings->num_mcp3427s = MAX_SOLAR_MCP3427;
```
GeraldNDA, line 74: Also there's nothing about that list that enforces this?

copperium:
@g-jessmuir - would like your opinion on the merits of extracting the callbacks from the MCP3427 settings in the mocked `mcp3427_start` in test_sense_mcp3427 versus mocking `mcp3427_register_callback` and `mcp3427_register_fault_callback`. Extracting the callbacks from `mcp3427_start` got rid of two mocks and the tests rely less on the implementation details of `sense_mcp3427`, but we have to rely on the structure of the internal `Mcp3427Storage` to retrieve the callbacks and contexts.

g-jessmuir:
> @g-jessmuir - would like your opinion on the merits of extracting the callbacks from the MCP3427 settings in the mocked `mcp3427_start` in test_sense_mcp3427 versus mocking `mcp3427_register_callback` and `mcp3427_register_fault_callback`. Extracting the callbacks from `mcp3427_start` got rid of two mocks and the tests rely less on the implementation details of `sense_mcp3427`, but we have to rely on the structure of the internal `Mcp3427Storage` to retrieve the callbacks and contexts.

I think pulling from `mcp3427_start` is fine. The storage is public anyways so it doesn't seem like a huge deal.

projects/solar/src/sense_mcp3427.c
```diff
@@ -0,0 +1,84 @@
...
+#define SENSE_MCP3427_CHANNEL MCP3427_CHANNEL_1
+
+static uint8_t s_num_mcp3427s;
+static Mcp3427Storage s_mcp3427_storages[MAX_SOLAR_MCP3427];
+static DataPoint s_mcp3427_data_points[MAX_SOLAR_MCP3427];
+static uint8_t s_indices[MAX_SOLAR_MCP3427];  // for contexts
+
+static int16_t s_values[MAX_SOLAR_MCP3427];
+static bool s_has_value[MAX_SOLAR_MCP3427];
+static uint8_t s_consecutive_faults[MAX_SOLAR_MCP3427];
```
GeraldNDA, line 22: This was probably here the last time I looked at it, but didn't realize. This is normally handled by having a big struct. And then pass the pointer to the correct struct instead of passing around the index.

projects/solar/src/sense_mcp3427.c
```diff
@@ -0,0 +1,84 @@
+#include "sense_mcp3427.h"
+#include "data_store.h"
```
GeraldNDA, line 2: nit: I guess the section thing isn't that strictly enforced lol, my b

projects/solar/test/test_sense_mcp3427.c
```diff
@@ -0,0 +1,327 @@
...
+  // start the MCP3427 cycles
+  TEST_ASSERT_OK(sense_mcp3427_start());
+  TEST_ASSERT_EQUAL(1, s_times_mcp3427_start_called);
+  data_store_get_is_set(TEST_DATA_POINT, &is_set);
+  TEST_ASSERT_EQUAL(false, is_set);  // not setting data yet
+
+  // make sure we received the various callbacks
+  TEST_ASSERT_NOT_NULL(s_sense_callbacks[0]);
+  TEST_ASSERT_NOT_NULL(s_mcp3427_callbacks[0]);
+  TEST_ASSERT_NOT_NULL(s_mcp3427_fault_callbacks[0]);
```
GeraldNDA, line 138: IMO Move these checks into the mocked callback. You shouldn't be setting any nulls there I believe.

projects/solar/src/sense_mcp3427.c
copperium: Ah whoops, I think the section thing was decided after this was written and I forgot to update it, my bad.

PR #121, opened 2020-07-03
[soft-224] Smoke bts7200

tristanle22:
Smoke test for BTS7200

- Periodically read and log the current readings from user selected bts7200 channels
- Had to copy some power distribution header and source filee to use the current hardware configs


projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+
+// Smoke test settings. Can be modified to fit testing purpose.
+#define CURRENT_MEASURE_INTERVAL_MS 500  // Set wait time between each set of readings
+#define IS_FRONT_POWER_DISTRO true       // Set whether to test FRONT or REAR power distro
+
+// Number of channels to be tested, MUST match the number of elements in test_channels
+#define NUM_TEST_CHANNELS 8
+// Set of channels to be tested, the number of elements MUST match NUM_TEST_CHANNELS.
+// Range of channels: 0-7
+static uint8_t test_channels[NUM_TEST_CHANNELS] = { 0, 1, 2, 3, 4, 5, 6, 7 };
```
copperium, line 24: As far as I can tell, the purpose of the `test_channels` array is to allow the smoketest user to exclude some BTS7200s from the test. In that case, you should make it clear that the "channels" are indices of BTS7200s in the front/rear HW config, and you should make the purpose of the array clear.

(Also, `test_channels` should be prefixed with `s_` since it's static, and I don't see a need for the `NUM_TEST_CHANNELS` define: can't you just use `SIZEOF_ARRAY(test_channels)` each time?)

projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+
+  PowerDistributionCurrentHardwareConfig s_hw_config =
+      IS_FRONT_POWER_DISTRO ? FRONT_POWER_DISTRIBUTION_CURRENT_HW_CONFIG
+                            : REAR_POWER_DISTRIBUTION_CURRENT_HW_CONFIG;
+
+  // Initialize the PCA9539R on every I2C address specified
+  for (uint8_t i = 0; i < s_hw_config.num_dsel_i2c_addresses; i++) {
+    status_ok_or_return(
+        pca9539r_gpio_init(s_hw_config.i2c_port, s_hw_config.dsel_i2c_addresses[i]));
+  }
```
copperium, line 55: You don't need to initialize the PCA9539Rs, `bts_7200_init_pca9539r` does that for you.

projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  for (uint8_t i = 0; i < NUM_TEST_CHANNELS; i++) {
+    // check that the currents are valid
+    if (s_hw_config.bts7200s[test_channels[i]].current_0 >= NUM_POWER_DISTRIBUTION_CURRENTS ||
+        s_hw_config.bts7200s[test_channels[i]].current_1 >= NUM_POWER_DISTRIBUTION_CURRENTS) {
+      return status_code(STATUS_CODE_INVALID_ARGS);
+    }
+
+    bts_7200_settings.select_pin = &s_hw_config.bts7200s[test_channels[i]].dsel_pin;
```
copperium, line 70: You also need to define `bts_7200_settings.i2c_port`.

projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+#define CURRENT_MEASURE_INTERVAL_MS 500  // Set wait time between each set of readings
+#define IS_FRONT_POWER_DISTRO true       // Set whether to test FRONT or REAR power distro
+
+// Number of channels to be tested, MUST match the number of elements in test_channels
+#define NUM_TEST_CHANNELS 8
+// Set of channels to be tested, the number of elements MUST match NUM_TEST_CHANNELS.
+// Range of channels: 0-7
+static uint8_t test_channels[NUM_TEST_CHANNELS] = { 0, 1, 2, 3, 4, 5, 6, 7 };
+
+static Bts7200Storage s_bts7200_storages[NUM_TEST_CHANNELS];
```
copperium, line 26: For the point about `NUM_TEST_CHANNELS` earlier: you can probably define a `MAX_TEST_CHANNELS` constant with the maximum number of distinct channels (probably 8) and use that here instead of `NUM_TEST_CHANNELS`. You'll have a few unused storages if the size of `test_channels` is less than the maximum but that's ok.

projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+  Bts7200Pca9539rSettings bts_7200_settings = {
+    .sense_pin = &s_hw_config.mux_address.mux_output_pin,
+  };
+
+  for (uint8_t i = 0; i < NUM_TEST_CHANNELS; i++) {
+    // check that the currents are valid
+    if (s_hw_config.bts7200s[test_channels[i]].current_0 >= NUM_POWER_DISTRIBUTION_CURRENTS ||
+        s_hw_config.bts7200s[test_channels[i]].current_1 >= NUM_POWER_DISTRIBUTION_CURRENTS) {
+      return status_code(STATUS_CODE_INVALID_ARGS);
+    }
```
copperium, line 68: Not sure why you're checking the validity of the current field in the HW config, it's just used in power distribution for determining what data point each BTS7200 is collecting (i.e. which board/light's current it is). You don't seem to use it in the smoke test.

projects/smoke_bts7200/src/main.c
```diff
@@ -0,0 +1,80 @@
+// A simple smoke test for BTS7200
+
+// Periodically take reading from user selected channels and log the result
+// Configurable items: wait time, FRONT or REAR power distro selection, channels to be tested
+#include "bts_7200_current_sense.h"
+#include "current_measurement.h"
+#include "current_measurement_config.h"
```
copperium, line 7: Nit: I'd prefer if you also include `pca9539r_gpio_expander.h` even though it's imported in `bts_7200_current_sense.h` just because it defines symbols used in this file.

projects/smoke_bts7200/inc/current_measurement.h
```diff
@@ -0,0 +1,68 @@
...
+  // Data of all BTS7200s from which to read
+  PowerDistributionBts7200Data *bts7200s;
+  uint8_t num_bts7200_channels;  // length of preceding array
+
+  // Data of all BTS7040/7008s from which to read
+  PowerDistributionBts7040Data *bts7040s;
+  uint8_t num_bts7040_channels;  // length of preceding array
+
+  MuxAddress mux_address;
+} PowerDistributionCurrentHardwareConfig;
```
copperium, line 46: This is the only section of this header file you actually need, all the rest isn't used in the smoke test.

projects/smoke_bts7200/inc/currents.h
```diff
@@ -0,0 +1,57 @@
...
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_4,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_5,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_6,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_7,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_8,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_9,
+  REAR_POWER_DISTRIBUTION_CURRENT_SPARE_10,
+
+  NUM_POWER_DISTRIBUTION_CURRENTS,
+} PowerDistributionCurrent;
```
copperium, line 57: Since you aren't using this enum in the smoke test, I think you could get away with removing this file and removing `.current_0` and `.current_1` from each entry in the config file. The fields would be set to garbage, but that doesn't matter since you aren't using them.

projects/smoke_bts7200/src/current_measurement_config.c
```diff
@@ -0,0 +1,212 @@
...
+          POWER_DISTRIBUTION_I2C_ADDRESS_0,  //
+          POWER_DISTRIBUTION_I2C_ADDRESS_1,  //
+      },
+  .num_dsel_i2c_addresses = 2,
+  .bts7200s =
+      (PowerDistributionBts7200Data[]){
+          {
+              .dsel_pin = FRONT_PIN_LEFT_RIGHT_CAM_DSEL,
+              .current_0 = FRONT_POWER_DISTRIBUTION_CURRENT_LEFT_CAMERA,
+              .current_1 = FRONT_POWER_DISTRIBUTION_CURRENT_RIGHT_CAMERA,
```
copperium, line 21: Like I commented above, I think you can remove these from each entry in this file. This will cut down on the unnecessary code.

projects/smoke_bts7200/src/current_measurement_config.c
```diff
@@ -0,0 +1,212 @@
...
+          {
+              .current = REAR_POWER_DISTRIBUTION_CURRENT_SPARE_6,
+              .mux_selection = 3,
+          },
+          {
+              .current = REAR_POWER_DISTRIBUTION_CURRENT_SPARE_7,
+              .mux_selection = 5,
+          },
+      },
+  .num_bts7040_channels = 6,
```
copperium, line 198: Since this isn't a BTS7040 smoke test, you can remove this since you aren't using it.

projects/smoke_bts7200/inc/current_measurement.h
copperium: You could also probably just move this struct to `current_measurement_config.h` and delete this whole file.

projects/smoke_bts7200/src/main.c
tristanle22: True I definitely should make it clearer what `test_channels` does. I always thought the size of array is required when declaring it, but I guess I could do `static uint_8 test_channels_s[] = {0,1,2,3,4,5,6,7}`

PR #122, opened 2020-07-06
[SOFT-211] sense_mppt

copperium:
Every sense cycle, `sense_mppt` reads the current, voltage, PWM, and status from each MPPT. It stores the current/voltage/PWM and the CR bit of the status in the data store, then checks the status byte for faults and raises a fault event if it finds any.

Note that this is *slightly* incompatible with `sense_mcp3427`, in that I moved `SolarMpptCount` to a new module `solar_boards.h` to avoid a circular dependency between `sense_mppt` and `solar_config`, so whenever one gets merged the other will have to be updated.


projects/solar/src/sense_mppt.c
```diff
@@ -0,0 +1,76 @@
...
+  if (spv1020_is_overvoltage(status)) {
+    event_raise(SOLAR_FAULT_EVENT_MPPT_OVERVOLTAGE, mppt);
+  }
+  if (spv1020_is_overtemperature(status)) {
+    event_raise(SOLAR_FAULT_EVENT_MPPT_OVERTEMPERATURE, mppt);
+  }
+}
+
+static void prv_sense_cycle_callback(void *context) {
+  uint8_t mppt = *(uint8_t *)context;
```
g-jessmuir, line 31: do you think it makes sense to `typedef uint8_t Mppt`?

projects/solar/inc/sense_mppt.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+// Implementation of sense for reading from the MPPTs.
+// Requires the event queue, GPIO, SPI, and the data store to be initialized.
```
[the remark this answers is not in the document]
copperium: We'll use the `mppt` module (which requires SPI) later for turning off/on MPPTs for recovering from faults in `solar_fsm`, so all SPI usage isn't actually owned by this module. Come to think of it, `sense_mppt` shouldn't own `mppt` either for the same reason. I'll move the `mppt` initialization out.

projects/solar/src/sense_mppt.c
copperium: I think that makes sense.

PR #123, opened 2020-07-06
[SOFT-189] DRV120 relay driver

hewittmcg:


libraries/ms-drivers/inc/drv120_relay.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+// Driver for the DRV120 relay
+// Requires GPIO and status to be initialized
```
copperium, line 4: `status` doesn't need to be initialized, so we shouldn't say it has to be initialized.

libraries/ms-drivers/inc/drv120_relay.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "gpio.h"
+#include "status.h"
+
+// Storage for GPIO pin to enable relay
+GpioAddress relay_pin;
```
copperium, line 10: This should be declared in the `.c` file and should be static (and be prefixed with `s_`). It's the module's internal storage, so it doesn't need to be exported.

libraries/ms-drivers/inc/drv120_relay.h
```diff
@@ -0,0 +1,16 @@
...
+
+StatusCode drv120_relay_init(GpioAddress *pin);
```
copperium, line 12: For this function it'd be nice to have a brief explanatory comment about what it does and what `pin` is - it's the enable pin on the DRV120. You should also say that the relay defaults to closed as soon as this function is called.

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -0,0 +1,38 @@
+// Brief test sequence for DRV120
+
+#include "drv120_relay.h"
+#include "log.h"
+#include "ms_test_helpers.h"
+
+// Pin used to enable/disable relay
+const GpioAddress TEST_DRV120_PIN = { GPIO_PORT_A, 8 };
```
copperium, line 8: This should be `static`, lowercase, and prefixed with `s_`.

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -0,0 +1,38 @@
...
+void teardown_test(void) {}
+
+void test_drv120_relay(void) {
+  // Variable used later to get GPIO state
+  GpioState state = GPIO_STATE_LOW;
+
+  LOG_DEBUG("Initializing relay \n");
+  TEST_ASSERT_OK(drv120_relay_init(&TEST_DRV120_PIN));
+
+  // Port should initialize to low: relay closed
```
copperium, line 23: Don't you mean it should initialize to **high**?

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -0,0 +1,38 @@
...
+  gpio_init();
+}
+
+void teardown_test(void) {}
+
+void test_drv120_relay(void) {
+  // Variable used later to get GPIO state
+  GpioState state = GPIO_STATE_LOW;
+
+  LOG_DEBUG("Initializing relay \n");
```
copperium, line 20: Minor nit: you don't need the space in front of the \n here and in the other logs.

libraries/ms-drivers/src/drv120_relay.c
```diff
@@ -1,24 +1,30 @@
 #include "drv120_relay.h"
 
-// Initialize pin for output
+
+// Storage for EN pin on DRV120 to open/close relay
+static GpioAddress s_relay_pin;
+
+// Initialize pin for output to open/close relay
+// Pin taken in is the EN pin on the DRV120: high = closed, low = open
+// Relay defaults closed on initialization
```
copperium, line 9: Nit: this should go in the header file since it's useful for users of the module - you shouldn't have to look in the implementation file to see the documentation.

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -5,10 +5,11 @@
 #include "ms_test_helpers.h"
 
 // Pin used to enable/disable relay
-const GpioAddress TEST_DRV120_PIN = { GPIO_PORT_A, 8 };
+static GpioAddress s_test_drv120_pin = { GPIO_PORT_A, 8 };
```
copperium, line 8: Minor nit: I think it's ok for this to be `static const` since it's still constant.

PR #125, opened 2020-07-07
Soft 175 smoke ads1259

mitchellostler:
Smoke test for ads1259 adc. 
Reads voltage into a buffer (ten readings) and then dumps values every second.


projects/smoke_ads1259/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  for (int reading = 0; reading < READING_QUEUE_LENGTH; reading++) {
+    LOG_DEBUG("VOLTAGE READING: %lf\n", queue[reading]);
+    queue[reading] = 0;
+  }
+  s_index = 0;
+}
+
+static void prv_periodic_read(SoftTimerId id, void *context) {
+  double *queue = context;
+  if (s_index < 10) {
```
g-jessmuir, line 33: should use `READING_QUEUE_LENGTH` here

PR #126, opened 2020-07-07
Soft 239 mcp23008 smoke test

shahhet14:
smoke test for mcp23008 gpio expander.
Initializes all 8 pins, checks for correct initialization and checks for correct toggling of states


projects/smoke_mcp23008/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+    if (get_state == *state) {
+      LOG_DEBUG("State for pin %d set and read correctly\n", pin);
+    } else {
+      LOG_DEBUG("State for pin %d incorrectly\n", pin);
+    }
+  }
+  *state =
+      (*state == MCP23008_GPIO_STATE_HIGH ? MCP23008_GPIO_STATE_LOW : MCP23008_GPIO_STATE_HIGH);
+
+  LOG_DEBUG("GPIO state = %d\n", *state);
```
g-jessmuir, line 87: I think this log should go before you flip the state, since this way you'll toggle gpio, then flip `*state`, then print `*state`, which would then be inaccurate

g-jessmuir: I also thing you should use a string format here to be more clear, something like
```
"GPIO state = %s\n", *state == STATE_HIGH ? "high" : "low"
```
Or something similar

projects/smoke_mcp23008/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+  soft_timer_start_millis(WAIT_TIME_MILLIS, prv_soft_timer_callback_output, state, NULL);
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+  setup();
```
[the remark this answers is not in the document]
shahhet14: its calling a function that initializes i2c and mcp23008 gpio, i'll make the function name more descriptive

PR #127, opened 2020-07-08
[SOFT-229] Power selection smoke test

jacobpiirsalu:
smoke test for power selection


GeraldNDA:
What's the point of this? We don't put implementations in headers so it's gonna need to be restructured. 

If the goal is to export some duplicated logic, then generalize it and add it to a test helper module. Otherwise, I think it's fine to define it twice across the two files.

g-jessmuir:
My bad on the super vague requirements, I didn't really think about the current implementation when I thought about this part.

I think it's a good move to just incorporate the 'smoke test' into the project, to make it most usable for someone on hardware I think the move is to:

- move the ADC readings and logging to a separate function
- leave a note stating this section could be used as a standalone smoke test, and what the readings are
- add a confluence page for a 'power selection smoke test' as a sub page here: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1263927334/Smoke+Tests that outlines where the smoke test is

Thanks for your work, sorry again for my mistake on this.

jacobpiirsalu:
I've made the changes according to the instructions (I believe this is what is required), separating logging and reading into separate functions in power_selection.c, with a note above the logging function that it can be used as a smoke test.

If this is not what is required I am happy to make more changes.

projects/power_selection/inc/power_selection.h
```diff
@@ -15,3 +15,6 @@
 #define temp_to_res(r) 33000.0 / (double)((r) / 1000.0) - 10000
 
 StatusCode aux_dcdc_monitor_init();
+
+// void smoke_test(uint16_t s_aux_volt, uint16_t s_aux_temp, double resistance, uint16_t
```
g-jessmuir, line 19: seems like we don't need these? please remove if unnecessary

projects/power_selection/src/power_selection.c
```diff
@@ -34,15 +34,32 @@ AdcChannel aux_channels[AUX_ADC_CURRENT_CHANNEL] = { [AUX_ADC_VOLT_CHANNEL] = AD
...
 static uint16_t s_aux_volt = 0;
 static uint16_t s_aux_temp = 0;
+
+// note: this section may be used as a smoke test, logging has been separated into this function
+/*The typical readings are:
+[0] projects/power_selection/src/power_selection.c:40: AUX Volatge Data: 2500
+[0] projects/power_selection/src/power_selection.c:41: AUX Temp Voltage Data: 2500
+[0] projects/power_selection/src/power_selection.c:42: AUX Temp Resistance Value: 3200.000000
+[0] projects/power_selection/src/power_selection.c:43: AUX Temp Data in C: 53
+*/
```
g-jessmuir, line 44: nit: we don't use multi line comments in our codebase, don't really know why, just a legacy thing, but please use single line comments

PR #129, opened 2020-07-11
[SOFT-269] Align precharge control with hardware

GeraldNDA:
Were some changes in hardware that didn't align. So these should hopefully fix. 

Instead of using a bitset, I just added an intermediate state. External users shouldn't worry about it, just that we're not in charged state.


projects/mci/src/precharge_control.c
```diff
@@ -17,16 +17,22 @@ static PrechargeControlStorage s_precharge_storage = { 0 };
...
-  gpio_set_state(&storage->precharge_control2, state);
   return STATUS_CODE_OK;
 }
 
-void prv_precharge_completed(const GpioAddress *address, void *context) {
+void prv_precharge_monitor(const GpioAddress *address, void *context) {
   PrechargeControlStorage *storage = context;
-  storage->state = MCI_PRECHARGE_CHARGED;
-  CAN_TRANSMIT_PRECHARGE_COMPLETED();
+  if (storage->state == MCI_PRECHARGE_INCONSISTENT) {
```
g-jessmuir, line 28: do you think it would be more clear if this was in chronological order? something like
```
if (storage->state == MCI_PRECHARGE_DISCHARGED) {
  ...
else {
  ...
  CAN_TRANSMIT_...
}
```

PR #131, opened 2020-07-11
[SOFT-9] BMS Carrier: Cell Sense Implementation

GeraldNDA:
Mostly ported from firmware:plutus/src/fault_monitor.c. 

Things that need special attention:
* new "pattern" for multiple event blocks. Not sure if it's a good idea, or should go back to normal
* adding to `current_sense` interface
* separation storage from readings and reformatting how afe readings are stored


projects/bms_carrier/src/cell_sense.c
```diff
@@ -0,0 +1,87 @@
...
+        s_storage.readings->voltages[i] > s_storage.settings.overvoltage_dmv) {
+      fault = true;
+    }
+  }
+
+  fault_bps(EE_BPS_STATE_FAULT_CURRENT_SENSE_AFE_CELL, !fault);
+}
+
+static void prv_extract_aux_result(uint16_t *result_arr, size_t len, void *context) {
+  ltc_afe_request_cell_conversion(s_storage.afe);
```
g-jessmuir, line 36: Do you think constantly switching between conversions would eat up too much compute? would BMS have enough time to work on other things?

Honestly no idea if that's a reasonable concern or not, there's just a lot of other things going on, mainly current sense reads a lot. Otherwise in my head there'd be a small soft timer between voltage/temp readings since they're unlikely to change a lot that rapidly

projects/bms_carrier/test/test_cell_sense.c
```diff
@@ -0,0 +1,265 @@
...
+  TEST_ASSERT_NOT_NULL(cb);
+  if (s_afe_should_fault)
+    return status_ok(event_raise_priority(EVENT_PRIORITY_HIGHEST, afe_events->fault_event, 0));
+  cb(result_arr, afe->settings.num_cells, afe->settings.result_context);
+  return status_ok(event_raise_priority(EVENT_PRIORITY_HIGHEST, afe_events->callback_run_event, 0));
+}
+
+static LtcAfeStorage s_afe;
+static StatusCode prv_init_ltc(void) {
+  status_ok_or_return(gpio_init());
```
g-jessmuir, line 64: why? isn't this guaranteed by other tests?

projects/bms_carrier/test/test_cell_sense.c
```diff
@@ -0,0 +1,265 @@
...
+  MS_TEST_HELPER_ASSERT_NEXT_EVENT_ID(e, BMS_AFE_EVENT_TRIGGER_CELL_CONV);
+  ltc_afe_process_event(&s_afe, &e);
+  prv_check_cell_results(true);
+
+  LOG_DEBUG("Verifying FSM result fault\n");
+  MS_TEST_HELPER_ASSERT_NEXT_EVENT_ID(e, BMS_AFE_EVENT_FAULT);
+  cell_sense_process_event(&e);
+}
+
+void prv_test_single_loop() {
```
g-jessmuir, line 133: is there a reason these `prv` functions are below setup/teardown? we don't really have solid convention

projects/bms_carrier/inc/bms_events.h
```diff
@@ -4,5 +4,17 @@ typedef enum {
   BMS_CAN_EVENT_RX = 0,
   BMS_CAN_EVENT_TX,
   BMS_CAN_EVENT_FAULT,
-  NUM_BMS_CAN_EVENTS
+  INVALID_BMS_CAN_EVENT
 } BmsCanEvent;
+#define NUM_BMS_CAN_EVENTS (INVALID_BMS_CAN_EVENT - BMS_CAN_EVENT_RX + 1)
```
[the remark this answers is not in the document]
GeraldNDA: Like this pattern. We'll should probably use this pattern for other modules I think.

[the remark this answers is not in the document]
GeraldNDA: Reason I went with this was because usage would be semi-similar to what it already is. Lowkey, I think just naming it `<event>_END` is sufficient to match the current pattern tho

projects/bms_carrier/inc/cell_sense.h
```diff
@@ -2,17 +2,41 @@
...
+#define NUM_TOTAL_CELLS (NUM_AFES * NUM_CELL_MODULES_PER_AFE)
+#define NUM_THERMISTORS (NUM_TOTAL_CELLS)
+#define MAX_AFE_FAULTS 5
 
-typedef struct AfeStorage {
-  uint16_t voltages[NUM_AFES][NUM_CELL_MODULES_PER_AFE];
-  uint16_t temps[NUM_AFES][NUM_THERMISTORS_PER_AFE];
-} AfeStorage;
+typedef struct CellSenseSettings {
+  // Units are 100 uV (or DeciMilliVolts)
```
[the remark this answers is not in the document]
GeraldNDA: This is what they were before so I assume it's good

projects/bms_carrier/src/cell_sense.c
GeraldNDA: There are delays between various parts of conversion so it should leave a decent amount of time to do other operations. We can add extra delays if these results aren't critical. I was thinking the same thing but it wasn't in the last car so I opted to use the same model.

projects/bms_carrier/test/test_cell_sense.c
GeraldNDA: prv functions above are mocking and checks. The ones below are exported test functionality. I'll move it up since I think that's a bit unexpected but these tests are supposed to act as a type of documentation of how to use the module/what it can do. So I figured I wanted to keep "what the module is doing" close to the actual tests

projects/bms_carrier/test/test_cell_sense.c
GeraldNDA: It is guaranteed but I use same pattern as used in the smoketest.
It's not a TEST_ASSERT because it should be guaranteed but I think it's fair to leave it if someone were to run test on h/w.

projects/bms_carrier/src/cell_sense.c
```diff
@@ -0,0 +1,87 @@
...
+  memset(afe_readings, 0, sizeof(AfeReadings));
+  memcpy(&s_storage.settings, settings, sizeof(CellSenseSettings));
+  ltc_afe_set_result_cbs(afe, prv_extract_cell_result, prv_extract_aux_result, NULL);
+  return ltc_afe_request_cell_conversion(afe);
+}
+
+StatusCode cell_sense_process_event(const Event *e) {
+  switch (e->id) {
+    case BMS_AFE_EVENT_FAULT:
+      // TODO(SOFT-9): Logic about when to trigger a fault could be exported to fault_bps
```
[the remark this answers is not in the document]
GeraldNDA: I'd prefer to leave it here myself. I just wanted to see if that was what you were planning on doing.

PR #132, opened 2020-07-11
[SOFT-223] Quick Lint/Format Bugfix

GeraldNDA:
Quick lint/format wouldn't find any files before this ...

Also included all files that changed from head, so less surprises in travis build 🤞

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ FIND_PATHS := $(addprefix -o -path $(LIB_DIR)/,$(IGNORE_CLEANUP_LIBS))
 FIND := find $(PROJECT_DIR) $(LIBRARY_DIR) \
 			  \( $(wordlist 2,$(words $(FIND_PATHS)),$(FIND_PATHS)) \) -prune -o \
 				-iname "*.[ch]" -print
-FIND_MOD_NEW := git diff --name-only --diff-filter=ACMRT -- '*.c' '*.h'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -157,7 +157,7 @@ FIND_PATHS := $(addprefix -o -path $(LIB_DIR)/,$(IGNORE_CLEANUP_LIBS))
 FIND := find $(PROJECT_DIR) $(LIBRARY_DIR) \
 			  \( $(wordlist 2,$(words $(FIND_PATHS)),$(FIND_PATHS)) \) -prune -o \
 				-iname "*.[ch]" -print
-FIND_MOD_NEW := git diff --name-only --diff-filter=ACMRT -- *.{c,h}
+FIND_MOD_NEW := git diff origin/master --name-only --diff-filter=ACMRT -- '*.c' '*.h'
```
AveryChiu64, line 160: Is this the part that checks for anything that's different from the HEAD?

AveryChiu64: Just curious how this works

GeraldNDA: Yea adding `origin/master` should compare to the HEAD of that branch. By default this would only find uncommitted/unstaged files. So this should hopefully be better assuming you've fetched.

The `*.c *.h` change lets git filter only C files. I think before it was trying to shell glob, which doesn't work here

PR #133, opened 2020-07-11
Soft 179 smoketest spv1020

JarvisWeng:


projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,81 @@
...
+
+  mppt_read_status(SPI_PORT, &status, SPV1020);
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, status);
+  mppt_read_pwm(SPI_PORT, &pwm, SPV1020);
+  LOG_DEBUG("SPV1020 # %d pwm is: %x\r\n", SPV1020, pwm);
+  mppt_read_voltage_in(SPI_PORT, &vin, SPV1020);
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, vin);
+  mppt_read_current(SPI_PORT, &current, SPV1020);
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, current);
+
```
copperium, line 49: Idea to make the smoke test even more useful: since you're using values that should never occur for status/pwm/vin/current, after reading everything you could check that all of the values have been changed and output a warning like "Reading PWM failed" if any of the variables have the same value they had originally.

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,81 @@
...
+  mppt_init();
+
+  LOG_DEBUG("Initializing spv1020 smoke test\n");
+
+  Event e = { 0 };
+  soft_timer_start_millis(SMOKETEST_WAIT_TIME_MS, prv_spv1020_check, NULL, NULL);
+  while (true) {
+    while (event_process(&e) != STATUS_CODE_OK) {
+    }
```
copperium, line 78: Why are you using the event queue here if there's no events to process? I think you could just replace all of this with a call to `wait()` from `wait.h`.

(I know we standardized the main event loop format but I don't think that applies to smoke tests without any events!)

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,81 @@
...
+    .baudrate = 60000,
+    .mode = SPI_MODE_3,
+    .mosi = MOSI_PIN,
+    .miso = MISO_PIN,
+    .sclk = SCLK_PIN,
+    .cs = CS_PIN,
+  };
+  spi_init(SPI_PORT, &spi_settings);
+
+  mppt_init();
```
copperium, line 70: I think you could get around having to copy/paste `mppt` into this project by moving all the mux config defines and the mux address from `mppt.c` into this file, documenting them, and initializing `mux` here. Then you can `mux_set(&s_mux_address, SPV1020)` at the beginning of `prv_spv1020_check` and `mux_set(&s_mux_address, DISCONNECTED_MUX_OUTPUT)` at the end of `prv_spv1020_check`. Then you can just call the raw `spv1020` functions instead of the `mppt` functions, and you don't have the copy/paste the `mppt` module.

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,81 @@
...
+  { .port = GPIO_PORT_B, 14 }
+#define SCLK_PIN \
+  { .port = GPIO_PORT_B, 13 }
+#define CS_PIN \
+  { .port = GPIO_PORT_B, 12 }
+
+uint8_t status = 0xFF;      // 8th bit is set
+uint16_t pwm = 0xFFFF;      // over 900
+uint16_t vin = 0xFFFF;      // over 10 bits
+uint16_t current = 0xFFFF;  // over 10 bits
```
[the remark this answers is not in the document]
JarvisWeng: Actually I don't remember what I was thinking when I did that. I think it was thinking something like saving memory, but idk

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,81 @@
...
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, status);
+  mppt_read_pwm(SPI_PORT, &pwm, SPV1020);
+  LOG_DEBUG("SPV1020 # %d pwm is: %x\r\n", SPV1020, pwm);
+  mppt_read_voltage_in(SPI_PORT, &vin, SPV1020);
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, vin);
+  mppt_read_current(SPI_PORT, &current, SPV1020);
+  LOG_DEBUG("SPV1020 # %d status is: %x\r\n", SPV1020, current);
+
+  mppt_shut(SPI_PORT, SPV1020);
+}
```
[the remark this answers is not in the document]
JarvisWeng: ah right. Thank you

projects/smoke_spv1020/src/main.c
JarvisWeng: ah ok. Sorry I'll change that

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+  uint16_t vin = 0xFFFF;      // over 10 bits
+  uint16_t current = 0xFFFF;  // over 10 bits
+
+  mux_set(&s_mux_address, SPV1020);
+
+  spv1020_turn_on(SPI_PORT);
+
+  spv1020_read_status(SPI_PORT, &status);
+  if (status != 0) {
+    LOG_DEBUG("Reading Status failed");
```
copperium, line 87: Reading the status didn't fail if `status == 0`! If `status != 0` then we successfully read the status but there's some fault in the SPV1020 chip (overcurrent/overtemperature/etc). This condition should check `status == 0xFF` since that's the initial, invalid value you gave it.

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    LOG_DEBUG("SPV1020 #%d status is: 0x%x\r\n", SPV1020, status);
+  }
+
+  spv1020_read_pwm(SPI_PORT, &pwm);
+  if (pwm == 0xFFFF) {
+    LOG_DEBUG("Reading PWM failed");
```
copperium, line 94: Your fault messages need `\n` at the end.

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+
+  mux_set(&s_mux_address, DISCONNECTED_MUX_OUTPUT);
+
+  soft_timer_start_millis(SMOKETEST_WAIT_TIME_MS, prv_spv1020_check, NULL, NULL);
+}
+
+int main(void) {
+  event_queue_init();
+  gpio_init();
+  gpio_it_init();
```
copperium, line 123: Why are you including + initializing `gpio_it`?

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+    LOG_DEBUG("Reading PWM failed");
+  } else {
+    LOG_DEBUG("SPV1020 #%d pwm is: 0x%x\r\n", SPV1020, pwm);
+  }
+
+  spv1020_read_voltage_in(SPI_PORT, &vin);
+  if (vin == 0xFFFF) {
+    LOG_DEBUG("Reading Voltage failed");
+  } else {
+    LOG_DEBUG("SPV1020 #%d voltage_in is: 0x%x\r\n", SPV1020, vin);
```
copperium, line 103: Nit: why are you using `\r\n` instead of just `\n`? (I believe minicom can be set to auto-insert carriage returns before newlines.)

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+    .sclk = SCLK_PIN,
+    .cs = CS_PIN,
+  };
+  spi_init(SPI_PORT, &spi_settings);
+
+  mux_init(&s_mux_address);
+
+  LOG_DEBUG("Initializing spv1020 smoke test\n");
+
+  Event e = { 0 };
```
copperium, line 141: Forgot to remove this?

projects/smoke_spv1020/src/main.c
```diff
@@ -0,0 +1,147 @@
...
+
+  spv1020_shut(SPI_PORT);
+
+  mux_set(&s_mux_address, DISCONNECTED_MUX_OUTPUT);
+
+  soft_timer_start_millis(SMOKETEST_WAIT_TIME_MS, prv_spv1020_check, NULL, NULL);
+}
+
+int main(void) {
+  event_queue_init();
```
copperium, line 121: I don't think you need to include + initialize `event_queue` since nothing in the smoke test uses it.

projects/smoke_spv1020/src/main.c
JarvisWeng: ah right thanks.

projects/smoke_spv1020/src/main.c
JarvisWeng: it needed interrupts to run on platform x86, but i guess I don't need it.

projects/smoke_spv1020/src/main.c
JarvisWeng: ok i'll remove that

projects/smoke_spv1020/src/main.c
JarvisWeng: Ah right, I'll get rid of the includes too.

projects/smoke_spv1020/src/main.c
JarvisWeng: yup

projects/smoke_spv1020/src/main.c
copperium: It seems to run fine for me on x86 without `gpio_it`

PR #134, opened 2020-07-16
Soft 221 smoke adt7476a

Max-MZ:


projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
+#include "adt7476a_fan_controller.h"
+#include "adt7476a_fan_controller_defs.h"
+#include "gpio.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define SET_SPEED_INTERVAL 5
```
g-jessmuir, line 10: please add a unit suffix at the end, e.g. `_S`. Also minor nit, the default interval should probably be 1 second so it's more immediately obvious if it's working or not

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
+#include "adt7476a_fan_controller.h"
+#include "adt7476a_fan_controller_defs.h"
+#include "gpio.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
```
g-jessmuir, line 9: add a comment here (or anywhere near the top) describing the expected behaviour, and document any values easily configurable (e.g. the interval and the increment), also the I2C address since we might be testing on a board with only one fan controller

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+#define SET_SPEED_INTERVAL 5
+#define FAN_SPEED_INCREMENT 10
+#define I2C_WRITE_ADDR_1 0x5E
+#define I2C_WRITE_ADDR_2 0x58
+#define I2C_READ_ADDR_1 0x5F
+#define I2C_READ_ADDR_2 0x59
+
+// 2 storages for 2 components
```
g-jessmuir, line 17: I think we should actually only test one fan controller at a time, because there are some boards (or at least there will be some boards) such as power distribution that will only use one fan controller

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+
+  I2CSettings i2c_settings = {
+    .speed = I2C_SPEED_FAST,
+    .scl = { .port = GPIO_PORT_B, .pin = 10 },
+    .sda = { .port = GPIO_PORT_B, .pin = 11 },
+  };
+
+  const Adt7476aSettings settings_1 = {
+    .smbalert_pin = smbalert,
+    .callback = NULL,  // set to NULL for no callback
```
g-jessmuir, line 46: please add a callback that prints something when the alert is triggered (ideally grabs status and logs which fan has bad status)

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  interrupt_init();
+  gpio_init();
+  soft_timer_init();
+  adt7476a_init(&s_storage_1, &settings_1);
+  adt7476a_init(&s_storage_2, &settings_2);
+  s_current_speed = 0;
+
+  soft_timer_start_seconds(SET_SPEED_INTERVAL, prv_periodic_set_speed, NULL, NULL);
```
g-jessmuir, line 71: rather than start a timer here, I think it makes sense to call `prv_periodic_set_speed` directly here, so that it logs something immediately and the periodic timer should happen anyways.

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+#define I2C_READ_ADDR_2 0x59
+
+// 2 storages for 2 components
+static Adt7476aStorage s_storage_1;
+static Adt7476aStorage s_storage_2;
+int s_current_speed;
+
+static void prv_periodic_set_speed(SoftTimerId id, void *context) {
+  s_current_speed += FAN_SPEED_INCREMENT;
+  s_current_speed = s_current_speed % 101;
```
g-jessmuir, line 24: why not just 100? I feel like I'm missing something here (which may be my intro programming memories)

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+  LOG_DEBUG("SETTING SPEED: %d PERCENT\n", s_current_speed);
+  adt7476a_set_speed(I2C_PORT_2, s_current_speed, ADT_PWM_PORT_1, I2C_WRITE_ADDR_1);
+  adt7476a_set_speed(I2C_PORT_2, s_current_speed, ADT_PWM_PORT_2, I2C_WRITE_ADDR_1);
+
+  soft_timer_start_seconds(SET_SPEED_INTERVAL, prv_periodic_set_speed, NULL, NULL);
+}
+
+int main() {
+  GpioAddress smbalert = {
```
g-jessmuir, line 33: probably good to add this as a #define up top since it may change for other boards that use the fan controllers

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+    .port = GPIO_PORT_A,
+    .pin = 9,
+  };
+
+  I2CSettings i2c_settings = {
+    .speed = I2C_SPEED_FAST,
+    .scl = { .port = GPIO_PORT_B, .pin = 10 },
```
g-jessmuir, line 40: same thing here - add gpio settings as #defines, including I2C port

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+    .callback = NULL,  // set to NULL for no callback
+    .callback_context = NULL,
+    .i2c_write_addr = I2C_WRITE_ADDR_2,
+    .i2c_read_addr = I2C_READ_ADDR_2,
+    .i2c = I2C_PORT_2,
+    .i2c_settings = i2c_settings,
+  };
+
+  interrupt_init();
+  gpio_init();
```
g-jessmuir, line 65: don't you need `gpio_it_init` as well to capture the smbalert?

projects/smoke_adt7476a/src/main.c
Max-MZ: wouldn't we not be able to hit full speed if it was `% 100`?

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,97 @@
...
+
+  this teset will increment the fan speed by 10% every second until at full speed,
+  and then loop back to 0% speed and continue.
+
+  the speed interval and increment defines can be changed for further testing
+
+  the i2c port, i2c write and read address, and i2c settings should be changed as needed
+  depending on the board.
+
+*/
```
g-jessmuir, line 20: other styling stuff: we only use single line comments, not sure why but at this point it's legacy

projects/smoke_adt7476a/src/main.c
```diff
@@ -0,0 +1,97 @@
...
+  interrupt_init();
+  gpio_init();
+  gpio_it_init();
+  soft_timer_init();
+  adt7476a_init(&s_storage, &settings);
+
+  s_current_speed = 0;
+
+  soft_timer_start_seconds(SET_SPEED_INTERVAL_S, prv_periodic_set_speed, NULL, NULL);
+  prv_periodic_set_speed(SOFT_TIMER_INVALID_TIMER, NULL);
```
g-jessmuir, line 93: this will start two recurring timers, won't it? I think if you just call `prv_periodic_set_speed` it'll automatically start a recurring timer, you don't have to set a timer for it separately

projects/smoke_adt7476a/src/main.c
g-jessmuir: im dumb

projects/smoke_adt7476a/src/main.c
Max-MZ: oop my bad I completely missed taking this out

PR #135, opened 2020-07-18
Soft 276 smoketest can

JarvisWeng:


projects/smoke_can/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+
+  soft_timer_start_millis(SMOKETEST_SEND_TIME_MS, prv_can_transmit, NULL, NULL);
+}
+
+static StatusCode prv_rx_callback(const CanMessage *msg, void *context, CanAckStatus *ack_reply) {
+  LOG_DEBUG("Received a message!\n");
+  printf("Data:\n\t");
+  for (uint8_t i = 0; i < msg->dlc; i++) {
+    uint8_t byte = 0;
+    byte = msg->data >> (i * 8);
```
copperium, line 51: Can `byte`'s declaration and setting its value be combined into one line? Also, I'd do `(msg->data >> (i * 8)) & 0xFF` just to be safe/explicit.

projects/smoke_can/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+    printf("%x ", byte);
```
copperium, line 52: You should put an `0x` before the `%x` just so it's clear that it's hex.

projects/smoke_can/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+#define TEST_CAN_MSG_ID 0x1
+#define TEST_CAN_BITRATE CAN_HW_BITRATE_500KBPS
+
+// this is how often the message will send
+// modify if you want to send more or less
+#define SMOKETEST_SEND_TIME_MS 1000
+// this is the data it'll send
+// change if you wish
+#define DATA \
+  { 0xfe, 0xdc, 0xba, 0x98, 0x76, 0x54, 0x32, 0x10 }
```
copperium, line 25: Make sure to say that it's an array of exactly 8 uint8s. In fact, it might be better to replace it with a `static const uint8_t s_data[8] = {...}` just to make it explicit.

projects/smoke_can/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+
+  LOG_DEBUG("Initializing smoke test can\n");
+
+  soft_timer_start_millis(SMOKETEST_SEND_TIME_MS, prv_can_transmit, NULL, NULL);
+
+  Event e = { 0 };
+  while (true) {
+    while (event_process(&e) != STATUS_CODE_OK) {
+    }
+    can_process_event(&e);
```
copperium, line 86: Nit: we standardized the event loop format so it should look like
```c
while (event_process(&e) == STATUS_CODE_OK) {
  can_process_event(&e);
}
```

projects/smoke_can/src/main.c
JarvisWeng: yup i'll change that

projects/smoke_can/src/main.c
JarvisWeng: ok

projects/smoke_can/src/main.c
JarvisWeng: will do

projects/smoke_can/src/main.c
JarvisWeng: ah sorry. Didn't check that

projects/smoke_can/src/main.c
JarvisWeng: it keeps throwing a 
initialization makes integer from pointer without a cast
so I'll just keep it as a #define

projects/smoke_can/src/main.c
JarvisWeng: It doesn't seem to work if I change it.
It seems that there's some time that the event_process will not return STATUS_CODE_OK so it'll just exit the loop.

This is only on x86 so it might be fine on hardware.

copperium: Ah whoops I meant
```c
while (true) {
  while (event_process(&e) == STATUS_CODE_OK) {
    can_process_event(&e);
  }
}
```

PR #136, opened 2020-07-23
[SOFT-277] data_store restructure and scaling factors

copperium:
The big gross enum in `data_store` is gone, replaced with `DataPointType`s and a bunch of macros that generate the data points. This makes it easier to change the number of MPPTs. It uses slightly more space but that's not a big deal.

We also now use `uint32_t`s in `data_store` for the purpose of using scaling factors in the `sense_*` modules. This way, we can define what our units are in `data_store`, and that way we can declare that we're working with values with units rather than meaningless ADC values.


projects/solar/test/test_sense_mcp3427.c
```diff
@@ -319,6 +335,42 @@ void test_sense_mcp3427_fault(void) {
...
+    } },
+    .num_mcp3427s = 1,
+  };
+  settings.mcp3427s[0].mcp3427_settings = s_test_mcp3427_settings;
+  TEST_ASSERT_OK(sense_mcp3427_init(&settings));
+  TEST_ASSERT_OK(sense_mcp3427_start());
+
+  prv_test_value_transform(&settings, 20, 5, "Scaling 20->5 (factor 0.25) failed.");
+  prv_test_value_transform(&settings, 10, 2, "Truncating 10->2 (factor 0.25) failed.");
+  prv_test_value_transform(&settings, -1025, 0xFFFFFF00, "-1025->0xFFFFFF00 (factor 0.25) failed.");
```
g-jessmuir, line 371: should probably have a note saying `0xFFFFFF0 == -256.25` for clarity

PR #138, opened 2020-07-31
[SOFT-212] Solar logger module

copperium:
It didn't take very long so I went ahead and did the logger module.

If you can think of a good way to test the output let me know.


g-jessmuir:
Not sure if you were stuck or just couldn't be bothered to fix it - the build issue was caused by print formatting mismatching int size when building for STM32F0XX and x86. Feel free to remove my commit for a different solution, just hoping to point you in the right direction.

g-jessmuir:
> If you can think of a good way to test the output let me know.

Tried mocking printf but it seemed kind of hard.

copperium:
Changed it to `long int` and `unsigned long int` since I think on STM32 `long int` is 32 bits and `int` is 16 bits, so casting to `int` would cause narrowing but casting to `long int` would not.

PR #139, opened 2020-08-01
[SOFT 214] data_tx solar module

shahhet14:
Created module that will get data from data_store and tx it into CAN.

Also includes codegen files.


projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+static CanMessage s_can_msg;    // variable used to hold can msgs and verify values
+
+bool TEST_MOCK(data_store_get_is_set)(DataPoint data_point, bool *is_set) {
+  return changeable_is_set;
+}
+
+StatusCode TEST_MOCK(data_store_get)(DataPoint data_point, uint32_t *value) {
+  *value = data_point;
+  return STATUS_CODE_OK;
+}
```
copperium, line 27: I'd prefer if you didn't mock the `data_store` functions and instead used the actual data store in the test. That way we keep mocks to a minimum and the test is less flimsy (e.g. the current implementation of `data_store_get_is_set` won't work because you have to assign to `is_set` and the function should return a `StatusCode`). Then to set all the data points you can just loop through and set them all (you may want a helper function for doing this).

Also for future reference, if you use mocks in a test you have to add the test to the project's `rules.mk` file, there are examples in `solar/rules.mk`.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "ms_test_helpers.h"
+#include "soft_timer.h"
+#include "solar_events.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+int count = -1;                 // count of can msgs sent, incremented in rx_handler
+bool changeable_is_set = true;  // value that tells data_tx if data points are set
```
copperium, line 17: These should be static and prefixed with `s_`. I'm also not too big a fan of the comments, I'd prefer more descriptive names like e.g. `can_msg_count` instead of the comment describing what the variable does (unless it's more complex).

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+bool TEST_MOCK(data_store_get_is_set)(DataPoint data_point, bool *is_set) {
+  return changeable_is_set;
+}
+
+StatusCode TEST_MOCK(data_store_get)(DataPoint data_point, uint32_t *value) {
+  *value = data_point;
+  return STATUS_CODE_OK;
+}
+
+typedef enum { TEST_DATA_TX_EVENT_CAN_TX = 0, TEST_DATA_TX_EVENT_CAN_RX } TestDataTxEvent;
```
copperium, line 29: To prevent clang-format from putting your enums (or structs/unions) on one line, you can put `//` after each line, like
```c
typedef enum {
  TEST_DATA_TX_EVENT_CAN_TX = 0,  //
  TEST_DATA_TX_EVENT_CAN_RX,      //
} TestDataTxEvent;
```
`make format` will automatically align the comments for you.

Also, this should have a `TEST_DATA_TX_EVENT_CAN_FAULT` member for the fault CAN event that the settings needs, and it should end with a `NUM_TEST_DATA_TX_CAN_EVENTS` member.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+static CanStorage s_can_storage = { 0 };
+static CanSettings s_can_settings = {
+  .device_id = SYSTEM_CAN_DEVICE_SOLAR,
+  .bitrate = CAN_HW_BITRATE_500KBPS,
+  .rx_event = TEST_DATA_TX_EVENT_CAN_RX,
+  .tx_event = TEST_DATA_TX_EVENT_CAN_TX,
+  .tx = { GPIO_PORT_A, 12 },
+  .rx = { GPIO_PORT_A, 11 },
+  .loopback = true,
+};
```
copperium, line 40: This needs a `.fault_event = TEST_DATA_TX_EVENT_CAN_FAULT` after you've added that event. (Also, it could be a local variable in `setup_test`, but that's your call.)

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+void teardown_test(void) {}
+
+// raise DataReadyEvent and data_tx will tx all data points into CAN, verify can message values and
+// count
+void test_data_tx(void) {
+  event_raise(DATA_READY_EVENT, 0);
+  Event e = { 0 };
+  MS_TEST_HELPER_AWAIT_EVENT(e);
```
copperium, line 68: You don't have to raise and capture the event, you can just do `Event e = { .id = DATA_READY_EVENT };` and pass that to `data_tx_process_event`.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+// ensure no can messages are sent when data points are not set
+void test_data_tx_no_values_set(void) {
+  count = 0;
+  changeable_is_set = false;
+  event_raise(DATA_READY_EVENT, 0);
+  Event e = { 0 };
+  MS_TEST_HELPER_AWAIT_EVENT(e);
```
copperium, line 86: Same as previous.

projects/solar/src/test_data_tx.c
copperium: Also, `count` should be a fixed-width type, probably `uint16_t`.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+  .device_id = SYSTEM_CAN_DEVICE_SOLAR,
+  .bitrate = CAN_HW_BITRATE_500KBPS,
+  .rx_event = TEST_DATA_TX_EVENT_CAN_RX,
+  .tx_event = TEST_DATA_TX_EVENT_CAN_TX,
+  .tx = { GPIO_PORT_A, 12 },
+  .rx = { GPIO_PORT_A, 11 },
+  .loopback = true,
+};
+
+StatusCode prv_test_data_tx_callback_handler(const CanMessage *msg, void *context,
```
copperium, line 42: Missing `static`.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+
+void teardown_test(void) {}
+
+// raise DataReadyEvent and data_tx will tx all data points into CAN, verify can message values and
+// count
+void test_data_tx(void) {
+  event_raise(DATA_READY_EVENT, 0);
+  Event e = { 0 };
+  MS_TEST_HELPER_AWAIT_EVENT(e);
+  TEST_ASSERT_EQUAL(changeable_is_set, true);
```
copperium, line 69: Like I said before I'd prefer if you used the data store module itself instead of the mocks, so you won't have to fiddle around with `changeable_is_set`. Instead, you should give a value to all the data points here.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+  data_tx_process_event(&e);
+
+  for (int can_msg = 0; can_msg < NUM_DATA_POINTS; can_msg++) {
```
copperium, line 72: Use `uint16_t` or another fixed-width type instead of `int` (`int` and friends change size between STM32 and x86)

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+void setup_test(void) {
+  gpio_init();
+  interrupt_init();
+  soft_timer_init();
+  event_queue_init();
+  TEST_ASSERT_OK(can_init(&s_can_storage, &s_can_settings));
+
+  TEST_ASSERT_OK(can_register_rx_handler(SYSTEM_CAN_MESSAGE_SOLAR_DATA,
+                                         prv_test_data_tx_callback_handler, NULL));
+}
```
copperium, line 59: You should reset `count` in `setup_test` instead of relying on the order of test execution.

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+
+// ensure no can messages are sent when data points are not set
+void test_data_tx_no_values_set(void) {
+  count = 0;
+  changeable_is_set = false;
+  event_raise(DATA_READY_EVENT, 0);
+  Event e = { 0 };
+  MS_TEST_HELPER_AWAIT_EVENT(e);
+  TEST_ASSERT_EQUAL(changeable_is_set, false);
+  TEST_ASSERT_EQUAL(count, 0);
```
copperium, line 88: I'm not sure why you're checking this, didn't you just set `count` and `changeable_is_set`?

projects/solar/src/data_tx.c
```diff
@@ -0,0 +1,28 @@
...
+    if (data_is_set) {
+      data_store_get(data_point, &data_value);
+      CAN_TRANSMIT_SOLAR_DATA((uint32_t)data_point, (uint32_t)data_value);
+    }
+  }
+}
+
+// if DataReady event is recieved, will call function to tx solar data
+void data_tx_process_event(Event *e) {
+  if (e->id == DATA_READY_EVENT) {
```
copperium, line 25: I might add a null check just so that we don't segfault if someone passes in `NULL`

projects/solar/src/test_data_tx.c
```diff
@@ -0,0 +1,93 @@
...
+  data_tx_process_event(&e);
+
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+  TEST_ASSERT_EQUAL(count, 0);
+}
```
copperium, line 93: Other things you could test:
* passing in `NULL` for the event
* passing in an event other than `DATA_READY_EVENT`, should do nothing
* if you change `data_tx_process_event` to return `bool`, add tests (or add to existing tests) to make sure the return value is working properly

shahhet14:
Redesigned the data_tx module to periodically tx data using soft timers to prevent the exhaustion of the CAN queue

shahhet14:
Redesigned the data_tx module to make the wait time between tx iterations and messages per tx iteration customizable. Data_tx now keeps tracks of messages txed using a local variable that is passed between iterations.

projects/solar/test/test_data_tx.c
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+static void prv_process_can_events(SoftTimerId timer_id, void *context) {
+  for (uint16_t msg = 0; msg < s_data_tx_settings.msgs_per_tx_iteration; msg++) {
+    MS_TEST_HELPER_CAN_TX(SOLAR_CAN_EVENT_TX);
+  }
+  for (uint16_t msg = 0; msg < s_data_tx_settings.msgs_per_tx_iteration; msg++) {
+    MS_TEST_HELPER_CAN_RX(SOLAR_CAN_EVENT_RX);
+  }
```
copperium, line 62: This will break if `NUM_DATA_POINTS` isn't a multiple of `msgs_per_tx_iteration`. If this ever occurs, the last `prv_process_can_events` call will try to process more TX events than there were TXes, and this will fail because `MS_TEST_HELPER_CAN_TX` will choke on the first RX event after the TX events.

To fix this, I'd set a variable at the top of the function to `MIN(s_data_tx_settings.msgs_per_tx_iteration, NUM_DATA_POINTS - s_can_msg_count)` and use that instead so that it won't try to process more TX events than are raised. (Note that the `MIN` macro is defined in misc.h.)

Even though we're keeping `msgs_per_tx_iteration` at 1 for now pending the x86 CAN bug, we'll likely increase it in the future.

projects/solar/test/test_data_tx.c
```diff
@@ -0,0 +1,120 @@
...
+  s_num_callbacks++;
+  if (s_can_msg_count < NUM_DATA_POINTS)
+    TEST_ASSERT_OK(soft_timer_start_millis(s_data_tx_settings.wait_between_tx_in_millis,
+                                           prv_process_can_events, NULL, NULL));
```
copperium, line 66: Please use curly brackets even around single line statements!

projects/solar/test/test_data_tx.c
```diff
@@ -0,0 +1,120 @@
...
+// expected order
+void test_data_tx(void) {
+  prv_test_data_tx_set_data_values();
+  Event e = { .id = DATA_READY_EVENT };
+  TEST_ASSERT(data_tx_process_event(&e));
+
+  prv_process_can_events(SOFT_TIMER_INVALID_TIMER, NULL);
+  int32_t delay_time = (NUM_DATA_POINTS / (int32_t)s_data_tx_settings.msgs_per_tx_iteration + 1) *
+                       (int32_t)s_data_tx_settings.wait_between_tx_in_millis;
+  delay_ms((uint32_t)delay_time);
```
copperium, line 111: Why all the casts with `int32_t`? If it's because `NUM_DATA_POINTS` is signed, I'd prefer you cast it instead, and if it's to widen everything to 32 bits I'd prefer you cast to `uint32_t` instead. Not a super big deal though.

projects/solar/inc/data_tx.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+// Recieves a DataReadyEvent
+// Takes data from data_store and tx each data point in a CAN message
+// Requires CAN, event_queue, soft_timers, interrupts and the data store to be initiallized
```
copperium, line 5: Couple of minor typos: `Recieves => Receives`, `initiallized => initialized`.

copperium (changes requested):
The only thing really blocking merging this is the point about the possible case where the test could fail. Everything else is super minor, if you address the couple of points we can get this merged!

shahhet14:
Fixed `prv_process_can_events ` so it doesnt fail when `NUM_DATA_POINTS` is not a multiple of `msgs_per_tx_iteration`. Also fixed the other minor errors.

copperium (approved):

PR #141, opened 2020-08-05
[SOFT 281] added loopback mode smoketest

AveryChiu64:


projects/smoke_mcp2515/src/main.c
```diff
@@ -62,15 +99,43 @@ static void prv_periodic_send(SoftTimerId timer_id, void *context) {
...
+  mcp2515_tx(&s_mcp2515, 0x19999999, true, 0xBEEFDEADBEEFDEAD, 8);
+  delay_ms(50);
+
+  if ((s_id_loopback == 0x19999999) && (s_extended == true) && (s_data == 0xBEEFDEADBEEFDEAD) &&
+      (s_dlc == 8)) {
+    LOG_DEBUG("Extended Message Received\n");
+  } else {
+    LOG_DEBUG("Extended message not received\n");
+  }
+}
```
g-jessmuir, line 124: Please add a broad "all loopback checks ok" kind of message to indicate whether or not the loopback was successful. With just this, it isn't 100% obvious whether or not things are working unless you look through the code.

projects/smoke_mcp2515/src/main.c
```diff
@@ -62,15 +99,43 @@ static void prv_periodic_send(SoftTimerId timer_id, void *context) {
...
+  } else {
+    LOG_DEBUG("Standard message not received\n");
+  }
+
+  LOG_DEBUG("Testing send extended id\n");
+  mcp2515_tx(&s_mcp2515, 0x19999999, true, 0xBEEFDEADBEEFDEAD, 8);
+  delay_ms(50);
+
+  if ((s_id_loopback == 0x19999999) && (s_extended == true) && (s_data == 0xBEEFDEADBEEFDEAD) &&
+      (s_dlc == 8)) {
```
g-jessmuir, line 119: you can't make this check because the first test set this to 8, you should probably zero out the static variables between tests

projects/smoke_mcp2515/src/main.c
```diff
@@ -21,8 +24,22 @@
 
 static SystemCanDevice s_id = 123;
 
+static Mcp2515Storage s_mcp2515;
 static GenericCanMcp2515 s_can_mcp2515;
 
+static uint32_t s_id_loopback = 0;
+static bool s_extended = false;
+static uint64_t s_data = 0;
+static size_t s_dlc = 0;
```
g-jessmuir, line 33: why not use a can message struct rather than all these?

PR #142, opened 2020-08-08
[Soft-272] passive balancing

hewittmcg:


GeraldNDA:
Instead of doing this on a soft_timer we should lowkey call `passive_balance` from cell_sense, the passive balance code is synchronous so it should fit well there

projects/bms_carrier/src/passive_balance.c
```diff
@@ -0,0 +1,38 @@
...
+#include "log.h"
+
+// Callback for soft timer
+static void prv_passive_balance_cb(SoftTimerId timer_id, void *context) {
+  passive_balance(context);
+
+  soft_timer_start_millis(PASSIVE_BALANCE_INTERVAL_MS, prv_passive_balance_cb, context, NULL);
+}
+
+StatusCode passive_balance_init(LtcAfeStorage *storage) {
```
GeraldNDA, line 11: This should be the AfeStorage, no?

If you end up moving it to being called in cell_sense then you might just be able to take in the result array and the length of it.

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,190 @@
...
+#define TEST_LTC_AFE_SPI_SCLK \
+  { .port = GPIO_PORT_A, .pin = 5 }
+#define TEST_LTC_AFE_SPI_CS \
+  { .port = GPIO_PORT_A, .pin = 4 }
+
+// Number of cell voltages to be directly modified during testing
+static const uint8_t NUM_TEST_VOLTAGES = 4;
+
+static LtcAfeSettings s_test_afe_settings = {
+  .num_cells = LTC_AFE_MAX_CELLS,
```
GeraldNDA, line 31: Probably should use the number of cells that we'd actually use. Internally we have an oversized array so using the max could mask invalid access bugs technically

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,190 @@
...
+  soft_timer_init();
+
+  // Configure all cells for output
+  for (uint8_t i = 0; i < LTC_AFE_MAX_DEVICES; i++) {
+    s_test_afe_settings.cell_bitset[i] = 1;
+    for (uint8_t j = 1; j < LTC_AFE_MAX_CELLS_PER_DEVICE; j++) {
+      s_test_afe_settings.cell_bitset[i] <<= 1;
+      s_test_afe_settings.cell_bitset[i] |= 0x1;
+    }
+  }
```
GeraldNDA, line 70: Might wanna change the bitset somewhat to make sure that you end up balancing the correct one. i.e.  or 0x0F0 vs. 0xF0F

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,190 @@
...
+  s_test_afe_storage.cell_voltages[2] = 1030;
+  s_test_afe_storage.cell_voltages[3] = 1010;
+
+  // Calculate device_cell in same way as ltc_afe_impl_toggle_cell_discharge would
+  // to check that bitset is as expected for the given cell
+  actual_cell = s_test_afe_storage.discharge_cell_lookup[balance_cell];
+  device_cell = actual_cell % LTC_AFE_MAX_CELLS_PER_DEVICE;
+  passive_balance(&s_test_afe_storage);
+  TEST_ASSERT_EQUAL(s_test_afe_storage.discharge_bitset[balance_device],
+                    AFE_DEFAULT_DISCHARGE_BITSET | (1 << device_cell));
```
GeraldNDA, line 110: Also I feel like you shouldn't be looking at the internals of other modules like this ... might be better to just mock and verify it was called with the correct values. That or add a function for checking discharge bitset to the ltc_afe interface

projects/bms_carrier/test/test_passive_balance.c
hewittmcg: You know off the top of your head how many we're using?  I can ask hardware if not

GeraldNDA: Checkout `cell_sense.h` I believe that's aligned with hardware

projects/bms_carrier/inc/passive_balance.h
```diff
@@ -1,7 +1,24 @@
 #pragma once
 
+// Passive balancing module for AFE cells.  Periodically checks for differences in voltages between
+// cells and raises an event in the AFE FSM to set the balance_control pin for the cell with the
```
GeraldNDA, line 4: this isn't how it works internally, so might wanna change that

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,118 @@
...
+// Balance values within allowed range, ensure they don't change
+void test_balance_range(void) {
+  LOG_DEBUG("Testing balancing around edge of range\n");
+  // Same general procedures as above
+
+  // Fill cells
+  uint16_t test_voltages[NUM_TOTAL_CELLS];
+  for (uint8_t i = NUM_TEST_VOLTAGES; i < NUM_TOTAL_CELLS; i++) {
+    test_voltages[i] = 1010;
+  }
```
GeraldNDA, line 94: might wanna make this static, and do this loop in setup_test, but lowkey this is fine

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Number of cell voltages to be directly modified during testing
+static const uint8_t NUM_TEST_VOLTAGES = 4;
+
+static LtcAfeStorage s_test_afe_storage;
+
+static const uint16_t AFE_DEFAULT_DISCHARGE_BITSET = 0;
+
+void setup_test(void) {}
```
GeraldNDA, line 27: should reset the static variables here

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,118 @@
...
+  // Only last cell should be balanced
+  test_voltages[0] = 1000;
+  test_voltages[1] = 1025;
+  test_voltages[2] = 1030;
+  test_voltages[balance_cell] = 1040;
+
+  // Verify
+  passive_balance(test_voltages, NUM_TOTAL_CELLS, &s_test_afe_storage);
+  TEST_ASSERT_EQUAL(s_cell_balanced, balance_cell);
+  TEST_ASSERT_EQUAL(s_cell_discharge, true);
```
GeraldNDA, line 82: actually these are technically 3 different tests? You should probably split them out.

projects/bms_carrier/src/cell_sense.c
```diff
@@ -29,6 +31,9 @@ static void prv_extract_cell_result(uint16_t *result_arr, size_t len, void *cont
     }
   }
 
+  // Balance cells if needed
+  passive_balance(result_arr, len, s_storage.afe);
```
GeraldNDA, line 35: `s_storage.readings->voltages` should be used here I think, the assumption is the prior array could be modified during execution (hence the critical section)

projects/bms_carrier/src/passive_balance.c
```diff
@@ -12,7 +11,6 @@ StatusCode passive_balance(uint16_t *result_arr, size_t len, LtcAfeStorage *afe)
 
   // Iterate through all cells in storage, updating values.
   for (uint8_t i = 0; i < len; i++) {
-    LOG_DEBUG("Value %d: %d\n", i, *result_arr);
     if (*result_arr > cell_voltage_max) {
```
GeraldNDA, line 14: Oh lol, I skipped over it, but now that you removed the comment ....
`result_arr` is an array, you can access the elements by doing `result_arr[i]` (`result_arr[0]` in the first case)
incrementing the pointer works but isn't how we would do things typically.

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -20,99 +20,103 @@ StatusCode TEST_MOCK(ltc_afe_toggle_cell_discharge)(LtcAfeStorage *afe, uint16_t
...
+static uint16_t s_test_voltages[NUM_TOTAL_CELLS];
 
-static const uint16_t AFE_DEFAULT_DISCHARGE_BITSET = 0;
+static LtcAfeStorage s_test_afe_storage;
 
-void setup_test(void) {}
+void setup_test(void) {
+  for (uint8_t i = NUM_TEST_VOLTAGES; i < NUM_TOTAL_CELLS; i++) {
+    s_test_voltages[i] = 1010;
+  }
```
GeraldNDA, line 31: Should probably initialize the whole array? Maybe set the defaults for the first 4 here, and then change the value of the balance cell in each test

projects/bms_carrier/test/test_passive_balance.c
```diff
@@ -0,0 +1,120 @@
...
+
+StatusCode TEST_MOCK(ltc_afe_toggle_cell_discharge)(LtcAfeStorage *afe, uint16_t cell,
+                                                    bool discharge) {
+  s_cell_balanced = cell;
+  s_cell_discharge = discharge;
+  return STATUS_CODE_OK;
+}
+
+// To store voltages for test
+static uint16_t s_test_voltages[NUM_TOTAL_CELLS];
```
GeraldNDA, line 21: Might wanna comment that this number is expected to be at least 4.

or ... you can set a local max cell define that is set to be the max of TOTAL_CELLS and 4.

PR #143, opened 2020-08-11
[SOFT-216] solar_fsm solar module

copperium:
This mainly manages the relay state. Based on https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1460011090/Fault+behaviour.

Behaviour:
* relay initializes to **closed**
* relay changes to **open** upon encountering the following fault conditions: output current over 9A, output current negative, sum of MPPT output voltages over 160V.
* relay will change between **closed** and **open** depending on battery relay CAN messages so that we mirror the battery relay (currently implemented as `SOLAR_COMMAND_EVENT_CLOSE/OPEN_RELAY`)


projects/solar/inc/solar_fsm.h
```diff
@@ -0,0 +1,31 @@
...
+#include "fsm.h"
+#include "status.h"
+
+#define MAX_RELAY_OPEN_EVENTS 8
+
+typedef struct SolarFsmSettings {
+  // Fault events which cause the relay to open.
+  EventId relay_open_events[MAX_RELAY_OPEN_EVENTS];
+  uint8_t num_relay_open_events;  // length of preceding array
+} SolarFsmSettings;
```
GeraldNDA, line 19: Do we ever expect the number of relay open events to change?

copperium: I think it could change if hardware thinks of or finds more faults that require isolation. It shouldn't change during runtime though, certainly.

copperium:
@GeraldNDA - I think you're referring to the external command event enum. I decided to use enums for the open/close relay commands since they're going to be translated from CAN messages, and I'd like to keep the (future) CAN handler loosely coupled from `solar_fsm`. (It would be a bit weird if the events were only raised internally.)

PR #144, opened 2020-08-12
[SOFT-175] debug current sense

g-jessmuir:


libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -118,14 +116,8 @@ StatusCode ads1259_init(Ads1259Settings *settings, Ads1259Storage *storage) {
...
   };
   status_ok_or_return(spi_init(settings->spi_port, &spi_settings));
-  delay_us(100);
-  // first command that must be sent on power-up before registers can be read
-  prv_send_command(storage, ADS1259_STOP_READ_DATA_CONTINUOUS);
   status_ok_or_return(prv_configure_registers(storage));
-  prv_send_command(storage, ADS1259_OFFSET_CALIBRATION);
-  delay_ms(s_calibration_time_ms_lookup[ADS1259_DATA_RATE_SPS]);
-  prv_send_command(storage, ADS1259_GAIN_CALIBRATION);
-  delay_ms(s_calibration_time_ms_lookup[ADS1259_DATA_RATE_SPS]);
```
copperium, line 128: Are the calibration commands not necessary? Also, removing these left `s_calibration_time_ms_lookup` unused, it should be removed as well.

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -1,4 +1,5 @@
 #include "ads1259_adc.h"
```
copperium, line 1: A few things I noticed about this driver:
* `Ads1259ErrorHandlerCb` has a `context` parameter, but there's no `context` field in the settings, and we just pass `NULL` for the context always, so the parameter is useless.
* We initialize SPI inside `ads1259_init`; I'd prefer to initialize it externally to the module like normal for better interoperability.
* The parameters are in the wrong order in `ads1259_init`, it's conventionally `storage, settings` to put storage first always.
* There's no callback function for when a conversion is ready, so a client (like the smoke test) has to wait with its own soft timer for the conversion time, which is really an implementation detail of the ADS1259 driver.

You can fix these if you'd like, though you don't have to if there's any danger it would break hardware validation.

projects/smoke_ads1259/src/main.c
```diff
@@ -1,15 +1,23 @@
+#include <stdlib.h>
+
 #include "ads1259_adc.h"
```
copperium, line 3: Should have `#include "adc1259_adc.h"` at the top, but not a big deal.

projects/smoke_ads1259/src/main.c
```diff
@@ -1,15 +1,23 @@
...
+// change this number to read x times in a cycle
+#define READ_CYCLE_SIZE 3
+
 // slightly larger than conversion time of adc
 #define CONVERSION_TIME_MS 18
-#define READING_QUEUE_LENGTH 10
+#define READING_QUEUE_LENGTH READ_CYCLE_SIZE
 static Ads1259Storage s_storage;
 static int s_index;
+static int s_count = 1;
```
copperium, line 20: Minor nit: these should probably be `uint8_t`s.

projects/smoke_ads1259/src/main.c
```diff
@@ -19,30 +27,29 @@ static void prv_rx_error_handler_cb(Ads1259StatusCode code, void *context) {
...
   double *queue = context;
   if (s_index < READING_QUEUE_LENGTH) {
+    LOG_DEBUG("=========READING # %i========= vref mv: %d\n", s_count++,
+              (int)(EXTERNAL_VREF_V * 1000));
     ads1259_get_conversion_data(&s_storage);
     queue[s_index] = s_storage.reading;
     s_index++;
     soft_timer_start_millis(CONVERSION_TIME_MS, prv_periodic_read, queue, NULL);
   } else {
-    prv_dump_queue(queue);
```
copperium, line 39: Maybe it's moot since the smoke test has already served its purpose, but how does the smoke test user know what's been read now?

libraries/ms-drivers/src/ads1259_adc.c
g-jessmuir: I'm taking hardware's word that the values were correct without calibration, so it's on them if they're not :) Removed `s_calibration_time_ms_lookup`

libraries/ms-drivers/src/ads1259_adc.c
g-jessmuir: - Added context field in settings
- There's actually not much convention on whether to init spi inside or outside the driver, see AFE and mcp2515 code. I'll leave it this way for now. (I notice we tend to init spi inside drivers, but i2c outside, may make sense since we rarely [except spv1020] use multiple spi ICs on the same bus but tend to use multiple i2c ICs on the same bus)
- Switched order for init args
- The lack of callback function was a (not necessarily good) design decision since it was a concern that the IC would hog processing time if it interrupted every cycle. It's much higher performance than we're using, so this way we can more easily tune it to whatever rate is necessary.

projects/smoke_ads1259/src/main.c
g-jessmuir: Not necessarily since that's not the respective header file for the module, right? just following clang format

projects/smoke_ads1259/src/main.c
g-jessmuir: changed to `uint8_t`

projects/smoke_ads1259/src/main.c
g-jessmuir: It's logged a few lines above. It looks good in the logs, not so much in code

g-jessmuir:
Added a few more changes to comment out some out of date testcases. Are you willing to let this go in? Honestly not feeling super up to figuring out the math on the conversions, but it works on hardware

projects/smoke_ads1259/src/main.c
copperium: Oh yeah 🤦 , thought this was a test for some reason.

PR #145, opened 2020-08-17
[SOFT-290] Add solar fault can msg

copperium:
See https://github.com/uw-midsun/codegen-tooling-msxiv/pull/22.


copperium:
The CAN message we're using for solar data is being added by Het's data_tx PR, the `SOLAR_DATA_FRONT` message was unused and confusing.

PR #146, opened 2020-08-17
[SOFT-290] Solar main module, fault_handler, command_rx

copperium:
This PR includes the changes from https://github.com/uw-midsun/firmware_xiv/pull/145 and https://github.com/uw-midsun/firmware_xiv/pull/143 (so those should probably be merged first). It will also need to be updated when `data_tx` gets merged.


projects/solar/src/fault_tx.c
```diff
@@ -0,0 +1,17 @@
...
+#include "solar_events.h"
+
+bool fault_tx_process_event(Event *e) {
+  if (e == NULL || e->id != SOLAR_FAULT_EVENT) return false;
+  uint8_t fault = GET_FAULT_FROM_EVENT(*e);
+  uint8_t data = GET_DATA_FROM_EVENT(*e);
+  if (fault >= NUM_EE_SOLAR_FAULTS) return false;
+  CAN_TRANSMIT_SOLAR_FAULT(fault, data);
+  return true;
+}
```
GeraldNDA, line 17: is it on purpose that you're doing this via the event queue as well? transmits are going to be put on the queue anyways so I feel like you should just transmit right away ...

projects/solar/src/solar_fsm.c
```diff
@@ -0,0 +1,92 @@
...
+#include "solar_events.h"
+#include "status.h"
+
+FSM_DECLARE_STATE(state_relay_open);
+FSM_DECLARE_STATE(state_relay_closed);
+
+static bool prv_relay_open_fault_guard(const Fsm *fsm, const Event *e, void *context) {
+  SolarFsmStorage *storage = context;
+  EESolarFault fault = GET_FAULT_FROM_EVENT(*e);
+  for (uint8_t i = 0; i < storage->num_relay_open_faults; i++) {
```
GeraldNDA, line 22: Seems pretty possible that non-fault event could be parsed as a fault event, no?

Also, I feel like this logic should go in "fault_tx" (with it being renamed probably) because it already catches these events. As I mentioned before ... you can probably just call a function i.e. `raise_fault` which does both the can transmit and if necessary calls a function here to advance the FSM.

projects/solar/src/solar_fsm.c
```diff
@@ -0,0 +1,92 @@
...
+  if (relay_mask & (1 << EE_RELAY_ID_BATTERY)) {
+    if (relay_state & (EE_RELAY_STATE_CLOSE << EE_RELAY_ID_BATTERY)) {
+      return event_raise_no_data(SOLAR_COMMAND_EVENT_CLOSE_RELAY);
+    } else {
+      return event_raise_no_data(SOLAR_COMMAND_EVENT_OPEN_RELAY);
+    }
+  }
+
+  return STATUS_CODE_OK;
+}
```
GeraldNDA, line 62: now that I'm seeing it, I feel like this is sufficiently different from the rest of this code for it to be separate. I don't think I'd expect to see this here based on filenames ...

projects/solar/test/test_solar_fsm.c
```diff
@@ -0,0 +1,235 @@
...
+  drv120_relay_init(&s_test_relay_pin);
+}
+void teardown_test(void) {}
+
+static void prv_process_all_events(SolarFsmStorage *storage) {
+  Event e = { 0 };
+  while (event_process(&e) == STATUS_CODE_OK) {
+    solar_fsm_process_event(storage, &e);
+    can_process_event(&e);  // for acks
+  }
```
GeraldNDA, line 47: might wanna use MS_HELPERS for these where the events are expected

projects/solar/src/solar_config.c
```diff
@@ -43,6 +48,79 @@
...
+  .sense_period_us = SENSE_CYCLE_PERIOD_US,
+};
+
+const SolarFsmSettings solar_fsm_settings = {
+  .relay_open_faults =
+      {
+          EE_SOLAR_FAULT_OVERCURRENT,
+          EE_SOLAR_FAULT_NEGATIVE_CURRENT,
+          EE_SOLAR_FAULT_OVERVOLTAGE,
+      },
```
GeraldNDA, line 120: my point was to try to avoid this ... there's no other consumers of these events it seems, and they could easily be replaced by a function that raises the internal event (similar to the can handler)

projects/solar/src/solar_fsm.c
```diff
@@ -0,0 +1,92 @@
...
+
+static StatusCode prv_handle_relay_state_can(const CanMessage *msg, void *context,
+                                             CanAckStatus *ack) {
+  // mirror the battery relay state
+  uint16_t relay_mask, relay_state;
+  CAN_UNPACK_SET_RELAY_STATES(msg, &relay_mask, &relay_state);
+
+  if (relay_mask & (1 << EE_RELAY_ID_BATTERY)) {
+    if (relay_state & (EE_RELAY_STATE_CLOSE << EE_RELAY_ID_BATTERY)) {
```
GeraldNDA, line 54: why not just 1 like the previous line ... if you want to check if the value matches EE_RELAY_STATE_CLOSE then you should do something like:
```c
(relay_state & (1 << EE_RELAY_ID_BATTERY)) >> EE_RELAY_ID_BATTERY  ==  EE_RELAY_STATE_CLOSE
```
(since changing to EE_RELAY_STATE_OPEN breaks the check)
Also technically [code style](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) for both require the `!= 0`. See "Comparison with NULL / 0" heading

GeraldNDA:
Is this PR still live?

projects/solar/test/test_relay_fsm.c
```diff
@@ -0,0 +1,147 @@
...
+  // open -> closed
+  TEST_ASSERT_OK(relay_fsm_close());
+  PROCESS_RELAY_FSM_EVENT(s_storage, e);
+  drv120_relay_get_is_closed(&closed);
+  TEST_ASSERT_TRUE(closed);
+
+  // closed still (should be no-op)
+  TEST_ASSERT_OK(relay_fsm_close());
+  PROCESS_RELAY_FSM_EVENT(s_storage, e);
+  drv120_relay_get_is_closed(&closed);
```
GeraldNDA, line 82: I feel like verifying no-ops would've been better off as a separate test while doing the normal operation would work well here.

projects/solar/test/test_relay_fsm.c
```diff
@@ -0,0 +1,147 @@
...
+// Test that |relay_fsm_process_event| returns whether there was a state change.
+void test_relay_fsm_process_event_return_value(void) {
+  // force to closed at start - valid in case we don't initialize to closed in the future
+  Event e = { 0 };
+  TEST_ASSERT_OK(relay_fsm_close());
+  PROCESS_RELAY_FSM_EVENT(s_storage, e);
+
+  e.id = IRRELEVANT_EVENT;
+  TEST_ASSERT_FALSE(relay_fsm_process_event(&s_storage, &e));
+
```
GeraldNDA, line 112: should probably verify that it is indeed closed here using `drv120_relay_get_is_closed`

projects/solar/src/main.c
```diff
@@ -1,2 +1,123 @@
...
+}
+
+static StatusCode prv_initialize_sense_modules(SolarMpptCount mppt_count) {
+  status_ok_or_return(mppt_init());
+
+  status_ok_or_return(sense_init(config_get_sense_settings()));
+
+  SenseMcp3427Settings sense_mcp3427_settings;
+  status_ok_or_return(config_get_sense_mcp3427_settings(mppt_count, &sense_mcp3427_settings));
+  status_ok_or_return(sense_mcp3427_init(&sense_mcp3427_settings));
```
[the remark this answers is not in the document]
copperium: The idea with the status_ok is that if the `mppt_count` is wrong, it'll return a status and not touch the settings pointer, so the status_ok returns gracefully instead of segfaulting trying to initialize garbage.

projects/solar/test/test_fault_monitor.c
```diff
@@ -3,17 +3,41 @@
...
 
+#define TEST_ASSERT_NO_FAULT() TEST_ASSERT_EQUAL(0, s_num_faults_raised)
+
+#define TEST_ASSERT_SINGLE_FAULT(fault, data)       \
+  ({                                                \
+    TEST_ASSERT_EQUAL(1, s_num_faults_raised);      \
+    TEST_ASSERT_EQUAL((fault), s_faults_raised[0]); \
+    TEST_ASSERT_EQUAL((data), s_fault_data[0]);     \
+    s_num_faults_raised = 0;                        \
+  })
```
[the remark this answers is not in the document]
copperium: This pattern is used in many other places (e.g. `ms_test_helpers.h`, `can_transmit.h`, `misc.h`) and ~8 times more frequently than `do {} while (0)`.

projects/solar/src/main.c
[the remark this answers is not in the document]
GeraldNDA: You already check for nulls in the inits tho. It's a straight forward debug if you just return NULL for a bad count imo

projects/solar/test/test_relay_fsm.c
copperium: I feel like this is fine since this test is just supposed to test `relay_fsm_process_event`'s return value, the other tests already test the relay state (including with respect to irrelevant events).

projects/solar/test/test_fault_monitor.c
```diff
@@ -3,17 +3,41 @@
...
 #define TEST_TEMPERATURE_THRESHOLD 1000000000uL
 
+#define TEST_ASSERT_NO_FAULT() TEST_ASSERT_EQUAL(0, s_num_faults_raised)
+
+#define TEST_ASSERT_SINGLE_FAULT(fault, data)       \
+  ({                                                \
+    TEST_ASSERT_EQUAL(1, s_num_faults_raised);      \
+    TEST_ASSERT_EQUAL((fault), s_faults_raised[0]); \
+    TEST_ASSERT_EQUAL((data), s_fault_data[0]);     \
+    s_num_faults_raised = 0;                        \
```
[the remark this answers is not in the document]
copperium: Renamed to `TEST_FAULT_MONITOR_ASSERT_[SINGLE/NO]_FAULT` so it's more clear that it's specific to this module.

projects/solar/src/mppt.c
```diff
@@ -5,16 +5,11 @@
...
-  }  // need to check hardward for correct pin
-     // this pin is unused
+#define UNUSED_PIN \
+  { GPIO_PORT_B, 1 }
+
+#define MUX_ENABLE UNUSED_PIN  // we don't have an accessible enable pin but the driver needs one
+#define MUX_OUTPUT UNUSED_PIN
 #define SEL_PIN_0 \
   { .port = GPIO_PORT_B, .pin = 3 }
 #define SEL_PIN_1 \
```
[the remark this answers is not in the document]
copperium: Moved these and the `solar_config` pin defines to a `pin_defs.h`.

projects/solar/src/main.c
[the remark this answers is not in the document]
copperium: Hm fair enough, I'll change them to return const pointers

projects/solar/test/test_fault_monitor.c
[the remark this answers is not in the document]
GeraldNDA: Did some research and apparently it's not listed on the style guide. do/while is nice because this pattern is a gcc-only thing but we should probz stick with what's in the can_transmit autogen code lol. 

@g-jessmuir we might want to cleanup the do/while macros if any and add to the style guide I guess.

projects/solar/test/test_relay_fsm.c
GeraldNDA: Fair enough, you only need this test because you don't check the return value in the other tests ... maybe extending the macro so you assert it might be worth?

copperium: Hm, extending the macro would probably make the other tests less clear since we'd have a seemingly unrelated bool parameter. We'd also need two macros, one checking and one not, since we don't want to check the return value when resetting the relay to open, since that would introduce a dependency on the initial state of the relay which I only want in the `test_starts_closed` test. It's probably fine having a test just to check the return value.

GeraldNDA: The assumption is already there, but it's up to you

PR #147, opened 2020-08-19
[SOFT-172] Implement current sense

g-jessmuir:
Current sense stores the last 20 readings, and faults
bps in case of overcurrent.

(also I'm dumb and broke the smoke test)

### DIFF
--- a/projects/bms_carrier/src/current_sense.c
+++ b/projects/bms_carrier/src/current_sense.c
@@ -6,10 +6,8 @@
 #include "log.h"
 #include "soft_timer.h"
 
-static CurrentReadings *s_readings;
 static Ads1259Storage s_storage;
-static uint16_t s_ring_idx = 0;
-static uint32_t s_conv_delay_ms = 0;
+static bool s_is_charging;
 
 static void prv_ads_error_cb(Ads1259StatusCode code, void *context) {
   if (code == ADS1259_STATUS_CODE_OUT_OF_RANGE) {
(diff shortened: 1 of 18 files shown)


projects/bms_carrier/test/test_current_sense.c
```diff
@@ -0,0 +1,113 @@
...
+void print_reads() {
+  for (int i = 0; i < NUM_STORED_CURRENT_READINGS; i++) {
+    printf("%d: %d\n", i, s_reads.readings[i]);
+  }
+  printf("avg: %d\n", s_reads.average);
+}
+
+void test_is_charging(void) {
+  s_ads_read = 0.5;
+  TEST_ASSERT_OK(current_sense_init(&s_reads, &s_spi_settings, TEST_CS_CONV_DELAY));
```
copperium, line 67: You may want a way to cancel the soft timer at the end of the tests since I'm pretty sure currently the timer will just keep running throughout all of the test cases.

projects/smoke_ads1259/src/main.c
```diff
@@ -34,6 +34,7 @@ static void prv_periodic_read(SoftTimerId id, void *context) {
               (int)(EXTERNAL_VREF_V * 1000));
     ads1259_get_conversion_data(&s_storage);
     queue[s_index] = s_storage.reading;
+    LOG_DEBUG("%d mV", (uint16_t)(s_storage.reading * 1000));
```
copperium, line 37: `\n`?

projects/bms_carrier/test/test_current_sense.c
```diff
@@ -0,0 +1,113 @@
...
+  s_fault_bps_bitmask = 0;
+  s_fault_bps_clear = false;
+
+  s_ads_read = 0.0;
+  s_ads_cb = NULL;
+}
+
+void teardown_test(void) {}
+
+void print_reads() {
```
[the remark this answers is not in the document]
g-jessmuir: whoops, this was just a function I was using for debugging. Removed it.

projects/bms_carrier/test/test_current_sense.c
g-jessmuir: I'm fairly certain timers are zeroed out upon `soft_timer_init()` in `setup_test`, I don't think this is an issue

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,80 @@
...
+  for (uint16_t i = 0; i < NUM_STORED_CURRENT_READINGS; i++) {
+    sum += s_readings->readings[i];
+  }
+  s_readings->average = sum / NUM_STORED_CURRENT_READINGS;
+}
+
+// returns value in centiamps
+int16_t prv_voltage_to_current(double reading) {
+  // current = voltage * 100, see confluence
+  return (int16_t)(100 * 100 * reading);
```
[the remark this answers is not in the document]
g-jessmuir: Liam gave me the value `100`, I'm inclined to trust him but I'll follow up to be sure

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,80 @@
 #include "current_sense.h"
 
-// TODO(SOFT-172): Actually implement this method
+#include "ads1259_adc.h"
+#include "bms.h"
```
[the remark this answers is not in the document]
g-jessmuir: Happened in a different PR, rebased to include this change

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,80 @@
...
+#include "exported_enums.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static CurrentReadings *s_readings;
```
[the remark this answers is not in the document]
g-jessmuir: I think it's worth having a no-argument `current_sense_is_charging`, but switched to just save a static bool rather than hold onto the main pointer. Also switched readings to a `CurrentStorage` with associated settings included.

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,82 @@
...
+  storage->ring_idx = (storage->ring_idx + 1) % NUM_STORED_CURRENT_READINGS;
+  ads1259_get_conversion_data(&s_storage);
+  soft_timer_start_millis(storage->conv_period_ms, prv_periodic_ads_read, context, NULL);
+
+  // update average
+  int32_t sum = 0;
+  for (uint16_t i = 0; i < NUM_STORED_CURRENT_READINGS; i++) {
+    sum += storage->readings_ring[i];
+  }
+  storage->average = sum / NUM_STORED_CURRENT_READINGS;
```
copperium, line 47: Assuming the storage gets zero-initialized at the start, the average calculation will be weighted towards 0 until we've read `NUM_STORED_CURRENT_READINGS` values, so the average won't be accurate for the first ~360ms. This is probably fine, though if we're overcurrent right at the start we won't know for ~360ms (also probably fine, but wanted to make sure you're aware).

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,82 @@
...
+    .spi_port = CURRENT_SENSE_SPI_PORT,
+    .spi_baudrate = settings->baudrate,
+    .mosi = settings->mosi,
+    .miso = settings->miso,
+    .sclk = settings->sclk,
+    .cs = settings->cs,
+    .handler = prv_ads_error_cb,
+    .error_context = NULL,
+  };
+  storage->conv_period_ms = conv_period_ms;
```
copperium, line 77: might want to do a memset on `storage` or else zero-initialize it manually so we don't get garbage readings (etc) causing false positive faults and/or segfaults if someone forgets to zero-initialize the storage.

projects/bms_carrier/src/current_sense.c
```diff
@@ -1,6 +1,82 @@
...
 
-// TODO(SOFT-172): Actually implement this method
+#include "ads1259_adc.h"
+#include "bms.h"
+#include "exported_enums.h"
+#include "fault_bps.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static Ads1259Storage s_storage;
```
copperium, line 10: Nit: calling this `s_storage` is kind of confusing since we also have a `storage`, might want to call it `s_ads1259_storage` or something

copperium (approved):
A couple nits, neither a big deal

PR #148, opened 2020-08-25
Soft 174 fan control module

Max-MZ:
build fails on 'test_solar'


projects/bms_carrier/inc/fan_control.h
```diff
@@ -2,18 +2,34 @@
...
 
 #include "cell_sense.h"
 
-#define BMS_FAN_CTRL_I2C_PORT_1 TBD
+#define BMS_FAN_CTRL_I2C_PORT_1 BMS_PERIPH_I2C_PORT
 #define BMS_FAN_CTRL_I2C_PORT_2 TBD
-#define NUM_BMS_FANS 4
+#define NUM_BMS_PORTS 2
+#define ADT_7476A_INTERRUPT_MASK_OFFSET 2
+#define FAN_TEMP_POLL_INTERVAL_US 500000
```
g-jessmuir, line 17: just by convention we tend to use milliseconds rather than microseconds, tends to be easier to read (along with `soft_timer_start_millis()`)

projects/bms_carrier/src/fan_control.c
```diff
@@ -0,0 +1,50 @@
+#include "bms.h"
+#include "math.h"
+
+static Adt7476aStorage adt7476a_storage;
+static SoftTimerId s_timer_id;
```
g-jessmuir, line 5: I don't think you use this, can probably remove it

projects/bms_carrier/inc/fan_control.h
```diff
@@ -2,18 +2,34 @@
...
+  uint16_t speed;
+  AfeReadings *readings;
   StatusCode status;
+  uint8_t i2c_write_addr;
+  uint8_t i2c_read_addr;
 } FanStorage;
 
-StatusCode fan_control_init(I2CSettings *settings, FanStorage *storage);
+typedef struct FanControlSettings {
+  I2CSettings i2c_settings;
```
g-jessmuir, line 30: I think this should include the poll interval, i.e. `uint32_t poll_interval_ms;`

projects/bms_carrier/src/fan_control.c
```diff
@@ -0,0 +1,50 @@
+#include "bms.h"
```
g-jessmuir, line 1: header convention is:
1. module header (i.e. `"fan_control.h"`)
2. blank line
3. system headers in angle brackets (i.e. `<string.h> or <math.h>`)
4. blank line
5. any other library headers like `soft_timer.h` or whatever else you might need

g-jessmuir: I know `fan_control.h` is redundant but eh convention is convention

projects/bms_carrier/src/fan_control.c
```diff
@@ -0,0 +1,50 @@
...
+  const Adt7476aSettings adt7476a_settings = {
+    .smbalert_pin = BMS_FAN_ALERT_PIN,
+    .callback = (GpioItCallback)settings->callback,  // set to NULL for no callback
+    .callback_context = settings->callback_context,
+    .i2c_write_addr = storage->i2c_write_addr,
+    .i2c_read_addr = storage->i2c_read_addr,
+    .i2c = BMS_FAN_CTRL_I2C_PORT_1,
+    .i2c_settings = settings->i2c_settings,
+  };
+
```
g-jessmuir, line 44: I think `s_interval_ms/us` should be set here

projects/bms_carrier/src/fan_control.c
Max-MZ: instead of passing `s_timer_id` in `soft_timer_start_millis`, should it be replaced with something like `SOFT_TIMER_INVALID_TIMER` there too?

copperium: You can pass `NULL` there, generally you only need to pass a non-null pointer there if you're going to cancel the timer later.

g-jessmuir (approved):
LGTM

PR #150, opened 2020-08-26
[SOFT-268] Added longer delay to 50ms

AveryChiu64:
Originally I was going to change the loopback mode to write to memory but it seems a bit complicated, changing the time delay may be a better solution for more consistency with Travis


g-jessmuir:
Travis still seems to be failing in spots, let's close this off and revisit later with a more thorough solution

PR #156, opened 2020-09-10
[SOFT-303] Integrate data_tx into solar main module

copperium:
This makes solar code complete. 🎉 

I also put `relay_fsm_process_event` at the top of the list of process_event calls so that it gets called first for a miniscule speed boost in opening the relay, though it probably doesn't matter.


GeraldNDA (approved):
LGTM, one suggestion you might wanna consider, but it's not blocking

PR #157, opened 2020-09-11
Soft 299 adc channels to gpio

mitchellostler:
Converted all functions in adc.c to use gpio pins instead of channels. Also rewrote x86 implementation and test_adc.c to deal with new changes. When tests are run on x86 all pass. Not to sure which programs use adc.h, but if codebase will need to be reworked to deal with this change, let me know and I can create a ticket, and add it to this pull request. Thanks!

**Overview of changes**
STM32F0xx:
- Changed all function stubs to use GpioAddresses instead of channels
- Reduced AdcChannel enum to only channels which do not have a corresponding pin
- Made adc_get_channel() private, and it is now used in all other functions at the very beginning to convert from gpio to channels (same logic in each function is preserved)
- Added prv_channel_to_gpio() -> this function is only used in ADC1_COMP_IRQHandler(), as channels are gotten directly from the system and need to be passed as gpioAddresses to the callback. If anyone has a better idea pls let me know

x86:
- Made similar changes to those above

Test_adc:
- Created static Gpio Array, and switched tests to gpioaddresses
- switched repetitive tests used on pins 1, 2, and 3 to for loops

### DIFF
--- a/libraries/ms-common/inc/adc.h
+++ b/libraries/ms-common/inc/adc.h
@@ -20,6 +20,7 @@ typedef enum {
   NUM_ADC_CHANNELS,
 } AdcChannel;
 
+// Allows use of either channel or GpioAddress, if not using channel write null
 typedef void (*AdcCallback)(GpioAddress address, void *context);
 
 // Initialize the ADC to the desired conversion mode
--- a/libraries/ms-common/inc/adc.h
+++ b/libraries/ms-common/inc/adc.h
@@ -14,47 +14,29 @@ typedef enum {
 } AdcMode;
 
 typedef enum {
-  ADC_CHANNEL_0 = 0,
-  ADC_CHANNEL_1,
-  ADC_CHANNEL_2,
-  ADC_CHANNEL_3,
-  ADC_CHANNEL_4,
-  ADC_CHANNEL_5,
-  ADC_CHANNEL_6,
-  ADC_CHANNEL_7,
-  ADC_CHANNEL_8,
-  ADC_CHANNEL_9,
-  ADC_CHANNEL_10,
-  ADC_CHANNEL_11,
-  ADC_CHANNEL_12,
-  ADC_CHANNEL_13,
-  ADC_CHANNEL_14,
-  ADC_CHANNEL_15,
-  ADC_CHANNEL_TEMP,
+  ADC_CHANNEL_TEMP = 16,
   ADC_CHANNEL_REF,
   ADC_CHANNEL_BAT,
   NUM_ADC_CHANNELS,
 } AdcChannel;
 
-typedef void (*AdcCallback)(AdcChannel adc_channel, void *context);
+typedef void (*AdcCallback)(GpioAddress address, void *context);
 
 // Initialize the ADC to the desired conversion mode
 void adc_init(AdcMode adc_mode);
+// To access ADC_CHANNEL_TEMP, ADC_CHANNEL_REF, ADC_CHANNEL_BAT, initialize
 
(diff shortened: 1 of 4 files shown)


libraries/ms-common/inc/adc.h
```diff
@@ -14,47 +14,31 @@ typedef enum {
...
-  ADC_CHANNEL_15,
-  ADC_CHANNEL_TEMP,
+  ADC_CHANNEL_TEMP = 16,
   ADC_CHANNEL_REF,
   ADC_CHANNEL_BAT,
   NUM_ADC_CHANNELS,
 } AdcChannel;
 
-typedef void (*AdcCallback)(AdcChannel adc_channel, void *context);
+// Allows use of either channel or GpioAddress, if not using channel write null
```
AveryChiu64, line 23: Is this NULL GpioAddress of NULL context?

libraries/ms-common/inc/adc.h
```diff
@@ -14,47 +14,31 @@ typedef enum {
...
-StatusCode adc_set_channel(AdcChannel adc_channel, bool new_state);
-
-// Return a channel corresponding to the given GPIO address
-StatusCode adc_get_channel(GpioAddress address, AdcChannel *adc_channel);
+StatusCode adc_set_channel(GpioAddress address, bool new_state);
 
 // Register a callback function to be called when the specified channel
 // completes a conversion
-StatusCode adc_register_callback(AdcChannel adc_channel, AdcCallback callback, void *context);
+StatusCode adc_register_callback(GpioAddress address, AdcCallback callback, void *context);
```
AveryChiu64, line 38: Since the ADC API is changed, I think it would be good to go through each project we currently have and see if they included "adc.h" (eg steering) in their code/tests.  If so, then change the ADC code for the project so it matches the new API. Then run on x86 to ensure that the test pass.

copperium (changes requested):
There are two issues with making a breaking change to the adc.h API like this:
* like @AveryChiu64 pointed out, the rest of the codebase won't compile without modification to conform to the new API, possibly introducing more bugs
* there are occasional times outside of adc.c where we need to access special channels like `ADC_CHANNEL_REF` and `ADC_CHANNEL_BAT`, e.g. in thermistor.c or tutorial_board_adc/src/main.c.

Because of this, I think it might be a better idea to maintain the original API but add secondary wrapper versions of each adc function that take GpioAddresses instead of AdcChannels and call the original function with the corresponding channel. This would make it not a breaking change, and would also avoid the hacky workarounds with setting a GPIO pin address to a channel to access a specific channel. You might want to rename the existing AdcChannel functions (with a global find+replace on the codebase) so that the wrapper functions look like the "defaults" externally.

libraries/ms-common/inc/adc.h
mitchellostler: Sorry, deprecated comment, will change

g-jessmuir:
> Because of this, I think it might be a better idea to maintain the original API but add secondary wrapper versions of each adc function that take GpioAddresses instead of AdcChannels and call the original function with the corresponding channel. This would make it not a breaking change, and would also avoid the hacky workarounds with setting a GPIO pin address to a channel to access a specific channel. You might want to rename the existing AdcChannel functions (with a global find+replace on the codebase) so that the wrapper functions look like the "defaults" externally.

I think I'm personally in favour of just marking the old functions with a 'deprecated' comment and adding new functions to save the refactor work. We can gradually port things if need be.

copperium:
Closing this in favour of https://github.com/uw-midsun/firmware_xiv/pull/163.

PR #158, opened 2020-09-11
[NO-MERGE] fw_102 onboarding exercise

anthony-chang:


projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#include <stdint.h>
+
+typedef struct Counters {
+  uint8_t counter_a, counter_b;
+} Counters_t;
```
g-jessmuir, line 10: for legacy consistency sake we just stick to PascalCase and ommit the `_t`

projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+
+void run(SoftTimerId timer_id, void *ctx) {
```
g-jessmuir, line 12: for module internal functions we prefix with `prv_` and generally make them static, also more descriptive names are appreciated e.g. `prv_timer_callback` or `prv_periodic_count` or similar

projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+  LOG_DEBUG("Counter A: %u\n", counter->counter_a);
+  if (!(counter->counter_a % 2)) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %u\n", counter->counter_b);
+  }
+
+  soft_timer_start_millis(500, run, counter, NULL);
+}
+
+int main(int argc, const char **argv) {
```
g-jessmuir, line 25: for better or for worse we don't make use of `main` params, so just `void` is fine here

projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+  Counters_t counter = { 0 };
+
+  interrupt_init();
+  soft_timer_init();
+
+  run(0, &counter);
```
g-jessmuir, line 31: generally if you're calling a soft timer callback directly it's best to use `SOFT_TIMER_INVALID_TIMER` as the parameter, since `0` is actually a valid timer id

projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+  Counters_t *counter = (Counters_t *)ctx;
+
+  counter->counter_a++;
+  LOG_DEBUG("Counter A: %u\n", counter->counter_a);
+  if (!(counter->counter_a % 2)) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %u\n", counter->counter_b);
+  }
+
+  soft_timer_start_millis(500, run, counter, NULL);
```
g-jessmuir, line 22: best to use a `#define COUNT_PERIOD_MS` with the `_MS` denoting milliseconds rather than hardcoded

g-jessmuir (changes requested):
Just nitpicks, I'll call it done post changes

projects/testing/src/main.c
```diff
@@ -0,0 +1,38 @@
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#include <stdint.h>
```
g-jessmuir, line 6: system header should be above library / module headers

g-jessmuir:
LGTM, closing as complete.

PR #165, opened 2020-09-16
[NO-MERGE] firmware 102 homework done, implemented simple soft timer counter

DoctaRay:
Screen Shot 2020-09-16 at 5 21 19 PM

### DIFF
--- a/projects/hello_firmware/src/main.c
+++ b/projects/hello_firmware/src/main.c
@@ -1,4 +1,4 @@
-#include <stdint.h>      // for integer types
+#include <stdint.h>  // for integer types
 
 #include "interrupt.h"   // interrupts are required for soft timers
 #include "log.h"         // for logging
@@ -18,7 +18,7 @@ typedef struct TimerData {
 
 static void prv_increment_and_print_counter(int counter_num, Counter *main_counter) {
   main_counter->counters[counter_num]++;
-  LOG_DEBUG("Counter %c: %d\n", counter_num+65, main_counter->counters[counter_num]);
+  LOG_DEBUG("Counter %c: %d\n", counter_num + 65, main_counter->counters[counter_num]);
 }
 
 static void prv_counter_callback(SoftTimerId timer_id, void *context) {
--- a/projects/hello_firmware/src/main.c
+++ b/projects/hello_firmware/src/main.c
@@ -1,0 +1,48 @@
+#include <stdint.h>  // for integer types
+
+#include "interrupt.h"  // interrupts are required for soft timers
+#include "log.h"
+#include "soft_timer.h"  // Software timers for scheduling future events.
+#include "wait.h"        // for wait function
+
+#define COUNTER_A_TIMER 500
+#define COUNTER_B_TIMER 1000
+
+typedef struct Counter {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counter;
+
+void counter_a_callback(SoftTimerId timer_id, void *context) {
+  Counter *mainCounter = context;
+  mainCounter->counter_a++;
+
+  LOG_DEBUG("Counter A: %d\n", mainCounter->counter_a);
+
+  soft_timer_start_millis(COUNTER_A_TIMER, counter_a_callback, mainCounter, NULL);
+}
+
+void counter_b_callback(SoftTimerId timer_id, void *context) {
+  Counter *mainCounter = context;
+  mainCounter->counter_b++;
+
+  LOG_DEBUG("Counter B: %d\n", mainCounter->counter_b);
+
+  soft_timer_start_millis(COUNTER_B_TIMER, counter_b_callback, mainCounter, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  Counter mainCounter = { .counter_a = 0, .counter_b = 0 };
+
+  soft_timer_start_millis(COUNTER_A_TIMER, counter_a_callback, &mainCounter, NULL);
+
+  soft_timer_start_millis(COUNTER_B_TIMER, counter_b_callback, &mainCounter, NULL);
+
+  while (true) {
+    wait();
+  }
+
+  return 0;
+}
--- a/projects/hello_firmware/src/main.c
+++ b/projects/hello_firmware/src/main.c
@@ -1,47 +1,45 @@
-#include <stdint.h>  // for integer types
+#include <stdint.h>      // for integer types
 
-#include "interrupt.h"  // interrupts are required for soft timers
-#include "log.h"
+#include "interrupt.h"   // interrupts are required for soft timers
+#include "log.h"         // for logging
 #include "soft_timer.h"  // Software timers for scheduling future events.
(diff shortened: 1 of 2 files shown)


projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,48 @@
...
+
+#define COUNTER_A_TIMER 500
+#define COUNTER_B_TIMER 1000
+
+typedef struct Counter {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counter;
+
+void counter_a_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 16: For private functions like this we mark them `static` (which makes them not visible from other files) and prefix them with `prv_` (so we know it's a private function).

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,48 @@
...
+
+  LOG_DEBUG("Counter B: %d\n", mainCounter->counter_b);
+
+  soft_timer_start_millis(COUNTER_B_TIMER, counter_b_callback, mainCounter, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  Counter mainCounter = { .counter_a = 0, .counter_b = 0 };
```
g-jessmuir, line 37: for variable names we use `snake_case` rather than `camelCase`

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,48 @@
...
+#define COUNTER_B_TIMER 1000
+
+typedef struct Counter {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counter;
+
+void counter_a_callback(SoftTimerId timer_id, void *context) {
+  Counter *mainCounter = context;
+  mainCounter->counter_a++;
```
copperium, line 18: We use `snake_case` for variable names, so this should be `main_counter` along with all the other occurrences.

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,48 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  Counter mainCounter = { .counter_a = 0, .counter_b = 0 };
+
+  soft_timer_start_millis(COUNTER_A_TIMER, counter_a_callback, &mainCounter, NULL);
+
+  soft_timer_start_millis(COUNTER_B_TIMER, counter_b_callback, &mainCounter, NULL);
```
g-jessmuir, line 41: in the instructions the requirement was
> Uses a single soft timer to increment counter_a every half a second, and counter_b every second.

It's trivial here, but the broader purpose is that we want to minimize resource usage on the microcontrollers, so grouping timer calls is good

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,48 @@
+#include <stdint.h>  // for integer types
+
+#include "interrupt.h"  // interrupts are required for soft timers
+#include "log.h"
+#include "soft_timer.h"  // Software timers for scheduling future events.
+#include "wait.h"        // for wait function
+
+#define COUNTER_A_TIMER 500
```
g-jessmuir, line 8: Please be a bit more descriptive with naming here, e.g. `COUNTER_A_PERIOD_MS` where the `_MS` suffix is to denote milliseconds. We usually use a units suffix for `#define`s that denote real values.

copperium (changes requested):
Awesome! Just a couple stylistic nitpicks, you can see the full style guide [here](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards).

projects/hello_firmware/src/main.c
g-jessmuir, line 18: please use `snake_case`

g-jessmuir (changes requested):
Aside from using two timers the rest of the comments are just nitpicks, please address before moving on to fw 103

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+static void prv_counter_b_callback(SoftTimerId timer_id, void *context) {
+  Counter *main_counter = context;
+  main_counter->counter_a++;
+  main_counter->counter_b++;
+
+  LOG_DEBUG("Counter A: %d\n", main_counter->counter_a);
+  LOG_DEBUG("Counter B: %d\n", main_counter->counter_b);
+
+  soft_timer_start_millis(COUNTER_B_DELAY_MS, prv_counter_a_callback, main_counter, NULL);
+}
```
copperium, line 37: This will increment `counter_b` every 1500ms and `counter_b` every 500ms, then 1000ms, then 500ms, etc. You probably only want one length of delay.

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+} Counter;
+
+static void prv_counter_a_callback(SoftTimerId timer_id, void *context);
+static void prv_counter_b_callback(SoftTimerId timer_id, void *context);
+
+static void prv_counter_a_callback(SoftTimerId timer_id, void *context) {
+  Counter *main_counter = context;
+  main_counter->counter_a++;
+
+  LOG_DEBUG("Counter A: %d\n", main_counter->counter_a);
```
copperium, line 23: Nitpick: since this code is repeated between this function and `prv_counter_b_callback`, I'd *prefer* it to be broken out into its own function. It's trivial for code of this size, but for larger projects significant duplicate code is a big no-no.

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+  soft_timer_start_millis(COUNTER_A_DELAY_MS, prv_counter_b_callback, main_counter, NULL);
+}
+
+static void prv_counter_b_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 28: Another nitpick: this function's name doesn't really reflect its purpose since it isn't only `counter_b` that's incremented here anymore. Perhaps `prv_both_counters_callback` or something like that could be a less misleading name?

copperium (changes requested):
Everything other than the delay length issue is a bit of a nitpick, I'll close this out after that's fixed.

projects/hello_firmware/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+} Counter;
+
+typedef struct TimerData {
+  Counter main_counter;
+  bool incrementBoth;
+} TimerData;
+
+static void prv_increment_and_print_counter(int counter_num, Counter *main_counter) {
+  main_counter->counters[counter_num]++;
+  LOG_DEBUG("Counter %c: %d\n", counter_num + 65, main_counter->counters[counter_num]);
```
copperium, line 21: I would prefer you use `'A'` instead of `65` to convert 0/1 to A/B, since in C `'A' == 65` and the character is clearer. I think it's okay for this though.

copperium (commented):
LGTM other than the one nit, closing this as homework is complete.

PR #166, opened 2020-09-17
Soft 307 babydriver setup

JarvisWeng:


projects/baby_driver/scripts/can_message.py
```diff
@@ -0,0 +1,21 @@
...
+
+import can
+
+def send_message(msg):
+    """sends can message with 8 bytes. You can change the 2nd one"""
+    #change the channel later to can0 or vcan0
+    bus = can.interface.Bus(bustype='socketcan', channel='can0', bitrate=250000)
+
+    msg = can.Message(
+        arbitration_id=0xC0FFEE, data=[0, msg, 0, 0, 0, 0, 0, 0], is_extended_id=False
```
copperium, line 14: arbitration ID is the device ID, this should at least be a constant and probably shouldn't be 0xC0FFEE but who knows. Also if we're going to use this generally you should definitely be able to change all the bytes not just the 2nd one!

projects/baby_driver/src/main.c
```diff
@@ -0,0 +1,56 @@
+#include "can.h"
+#include "event_queue.h"
+#include "gpio.h"
+#include "interrupt.h"
+#include "log.h"
+
+#define CAN_DEVICE_ID 0x1
```
copperium, line 7: We might want to define a new device ID for babydriver, we're not about to run out of them. Alternatively we could just use something large and silly à la 0xC0FFEE (but not exactly that since the python script is using it)

projects/baby_driver/scripts/can_message.py
```diff
@@ -0,0 +1,21 @@
...
+# coding: utf-8
+"""python can message with 8 bytes"""
+from __future__ import print_function
+
+import can
+
+def send_message(msg):
+    """sends can message with 8 bytes. You can change the 2nd one"""
+    #change the channel later to can0 or vcan0
+    bus = can.interface.Bus(bustype='socketcan', channel='can0', bitrate=250000)
```
copperium, line 11: The bitrate here needs to agree with the bitrate in the CAN settings, so change it to 500000. Also did you test this with vcan0 and the CAN message showed up on candump?

projects/baby_driver/scripts/can_message.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+# coding: utf-8
+"""python can message with 8 bytes"""
+from __future__ import print_function
```
copperium, line 4: don't think this is necessary since it's python 3

projects/baby_driver/scripts/first.py
```diff
@@ -0,0 +1,4 @@
+"""thin wrapper over can_message"""
+from scripts import can_message
+
+can_message.send_message(1)
```
copperium, line 4: This file is just for testing right? not sure if necessary

projects/baby_driver/scripts/can_message.py
```diff
@@ -0,0 +1,21 @@
...
+
+import can
+
+def send_message(msg):
+    """sends can message with 8 bytes. You can change the 2nd one"""
+    #change the channel later to can0 or vcan0
```
copperium, line 10: stylistic nit: please put a space after the #, though I'm not sure the comment is necessary. You should probably make the channel a keyword argument to `send_message` (like `channel="can0"`).

also some docstring formatting suggestions that were used in Arshan's can_dump project are here: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

Makefile
```diff
@@ -246,3 +246,13 @@ update_codegen:
 
 # Dummy force target for pre-build steps
 .PHONY: .FORCE
+
+#currently only for stm32
+#will need different bash window to run the python
+#to send CAN message write
+#can_message.send_message(1) to send can message
```
copperium, line 253: not really sure if this is the best place for this documentation

projects/baby_driver/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+
+  can_init(&s_can_storage, &s_can_settings);
+  can_register_rx_default_handler(prv_default_rx_callback, NULL);
+
+  Event e = { 0 };
+  while (true) {
+    while (event_process(&e) != STATUS_CODE_OK) {
+    }
+    can_process_event(&e);
+  }
```
copperium, line 53: The event loop standardization in the style guide would prescribe
```c
while (true) {
  while (event_process(&e) == STATUS_CODE_OK) {
    can_process_event(&e);
  }
}
```
as the format for this, though it's equivalent

projects/baby_driver/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+int main() {
+  LOG_DEBUG("Welcome to BabyDriver!\n");
+  gpio_init();
+  event_queue_init();
+  interrupt_init();
+
+  can_init(&s_can_storage, &s_can_settings);
+  can_register_rx_default_handler(prv_default_rx_callback, NULL);
```
copperium, line 46: I'm pretty sure the thought was that individual babydriver modules (like gpio_set etc) would have their own RX handlers that would just get registered here, I don't think we need a default callback

copperium (changes requested):

Makefile
```diff
@@ -135,7 +135,7 @@ ROOT := $(shell pwd)
 
 # Actually calls the make
 .PHONY: all
-all: build lint pylint
+all: build lint pylint babydriver
```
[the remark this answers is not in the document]
JarvisWeng: Ah right forgot to remove that when i was learning about make

projects/baby_driver/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+  .fault_event = CAN_EVENT_FAULT,
+  .tx = { GPIO_PORT_A, 12 },
+  .rx = { GPIO_PORT_A, 11 },
+  .loopback = false,
+};
+
+static GpioSettings s_led_settings = { .direction = GPIO_DIR_OUT,
+                                       .state = GPIO_STATE_HIGH,
+                                       .alt_function = GPIO_ALTFN_NONE,
+                                       .resistor = GPIO_RES_NONE };
```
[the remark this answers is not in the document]
JarvisWeng: Ah, I actually wrote the set_gpio callback before I removed it. Guess I forgot to remove this too

projects/baby_driver/scripts/can_message.py
JarvisWeng: Yup, it was just an id. I can change it to like 0x2. Yeah, I didn't complete it because I wasn't sure what the format would be.
Like which byte would represent what. (Ex: byte 1 would be the function)
But I'll edit it to just do msg1, msg2, msg3.... and we can change the names later if necessary

projects/baby_driver/src/main.c
JarvisWeng: Yeah 0x1 was just the default. Also does it matter? I thought these were for hardware and mainly for testing so the id shouldn't matter since the board wouldn't connect to other pieces. I might be wrong though.

projects/baby_driver/scripts/first.py
JarvisWeng: Oh yeah. It's just a sample in case we wanted to change from just using can_message.py and instead have wrappers that just call the specific function callback we want. I can remove if you want.

projects/baby_driver/scripts/can_message.py
JarvisWeng: We could but right now the make babydriver is for the stm32. I could try to setup it with x86 though, I'll need to figure out a way to prompt another command window.

Cool, I'll check it out. Thanks!

Makefile
JarvisWeng: Ah hm. I could put it in the python script or the c project.

projects/baby_driver/src/main.c
JarvisWeng: Right keep forgetting.

projects/baby_driver/src/main.c
JarvisWeng: Ah, that was just for the new members so that they can have an example if they were working on it. I can just remove it though.

Makefile
```diff
@@ -246,3 +246,13 @@ update_codegen:
...
+
+#currently only for stm32
+#will need different bash window to run the python
+#to send CAN message write
+#can_message.send_message(1) to send can message
+.PHONY: babydriver
+babydriver:
+	@make program PROJECT=baby_driver
+	@python3
+	@from projects.baby_driver.scripts import can_message
```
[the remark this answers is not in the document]
JarvisWeng: Ah I see. Thanks I'll look into it

projects/baby_driver/scripts/can_message.py
JarvisWeng: Ah I did. Weird, it still received it.

projects/baby_driver/src/main.c
copperium: It probably won't connect to other boards, but I don't see a reason to inherently limit babydriver to not being able to connect to other boards.

projects/baby_driver/scripts/can_message.py
copperium: I think this method is just going to be a generic low-level "send a CAN message" function, so the data would just be a list of bytes and we'll give it an interpretation somewhere else.

Also, there is a native data type in python for an array of bytes: `bytearray`. I think instead of msg1, msg2, ... you can take a list called `msg`, convert it to a `bytearray`, and check that its length is <=8.

projects/baby_driver/src/main.c
copperium: Yeah, there should be examples in the ticket or the firmware 10X notes. Don't think it's necessary to clutter the code.

projects/baby_driver/scripts/can_message.py
JarvisWeng: Yup. Actually it can be a byte array or a single number. If the byte array is >8 then the can library just won't send it

projects/baby_driver/src/main.c
JarvisWeng: Sure. I'll leave it as 0x1 for now and we can define what ID it should have. Same with the python script

Makefile
GeraldNDA: Once this solidifies, you might wanna look into not forcing 2 windows. You can run multiple things in parallel in the same command using [this](https://stackoverflow.com/a/9220818), you can still have it as multiple lines in the make file using [this](https://stackoverflow.com/a/10121575). 

The second command should still have stdin readable from the commandline, so you'd want to use 2 commands if there's stdin for both. If not and you're worried about stdout being cluttered by both, you can prefix the output using something like [this](https://unix.stackexchange.com/a/440438) to make things a bit easier on the eyes

Makefile
GeraldNDA: I'd probably suggest both, and at the top of this file as well

projects/baby_driver/src/main.c
GeraldNDA: (we should probably set the standard to use the `status_ok` helper function :eyes:)

Makefile
JarvisWeng: Oh right now, it doesn't need 2 windows. I'm saying we need 2 windows on x86 because 1 will show the output from the C project, and the other will be needed to run the python scripts. I'm not sure it's a parallel process issue, but I could be misinterpreting what you mean. If you want, we can set a time to talk about it to clarify things.

projects/baby_driver/scripts/can_message.py
```diff
@@ -0,0 +1,20 @@
...
+
+import can
+
+def send_message(message):
+    """sends can message with 8 bytes. You can change the 2nd one"""
+    # change the channel later to can0 or vcan0
+    bus = can.interface.Bus(bustype='socketcan', channel='can0', bitrate=500000)
+
+    msg = can.Message(
+        arbitration_id=0x2,
```
copperium, line 11: It looks like I was wrong, `arbitration_id` is 11 bits: 4 bits for the source ID, 1 bit for type (0=data, 1=ack), 6 bits for the message ID. `send_message` should probably take this into account or at least allow setting the arbitration ID.

projects/baby_driver/src/main.c
```diff
@@ -0,0 +1,57 @@
+
+// This project will receive a CAN message
+// on the stm32 andperform specific callback
+// based on the message
+
+// To run this project, simply run
+// make babydriver
+// and that should program to the stm32
+// To send a CAN message write
+// can_message.send_message(<data>) to send can message
```
copperium, line 10: Should probably clarify that this is from python

copperium (approved):

PR #168, opened 2020-09-20
[SOFT-301] race condition x86 can

mitchellostler:
added flag to check if X86 is defined and not execute body of prv_tx_callback in can.c if it is. 
Wrote requisite test as well.

### DIFF
--- a/libraries/ms-common/src/can.c
+++ b/libraries/ms-common/src/can.c
@@ -21,12 +21,13 @@
 #include "log.h"
 #include "soft_timer.h"
 
-#define CAN_BUS_OFF_RECOVERY_TIME_MS 500  
+#define CAN_BUS_OFF_RECOVERY_TIME_MS 500
 
+//flag enabled if x86 passed as platform
 #ifdef X86
-  #define TX_CALLBACK_ENABLE 0
-#else 
-  #define TX_CALLBACK_ENABLE 1
+#define TX_CALLBACK_ENABLE 0
+#else
+#define TX_CALLBACK_ENABLE 1
 #endif
 
 // Attempts to transmit the specified message using the HW TX, overwriting the
@@ -109,7 +110,7 @@ StatusCode can_register_rx_default_handler(CanRxHandlerCb handler, void *context
   return can_rx_register_default_handler(&s_can_storage->rx_handlers, handler, context);
 }
 
-StatusCode can_register_rx_handler(CanMessageId msg_id, CanRxHandlerCb handler, void *context) {
+StatusCode  can_register_rx_handler(CanMessageId msg_id, CanRxHandlerCb handler, void *context) {
   if (s_can_storage == NULL) {
     return status_code(STATUS_CODE_UNINITIALIZED);
   }
@@ -151,7 +152,7 @@ bool can_process_event(const Event *e) {
 }
 
 void prv_tx_handler(void *context) {
-  if(TX_CALLBACK_ENABLE) {
+  if (TX_CALLBACK_ENABLE) {
     CanStorage *can_storage = context;
     CanMessage tx_msg;
     // If we failed to TX some messages or aren't transmitting fast enough, those
--- a/libraries/ms-common/src/can.c
+++ b/libraries/ms-common/src/can.c
@@ -21,7 +21,13 @@
 #include "log.h"
 #include "soft_timer.h"
 
-#define CAN_BUS_OFF_RECOVERY_TIME_MS 500
+#define CAN_BUS_OFF_RECOVERY_TIME_MS 500  
+
+#ifdef X86
+  #define TX_CALLBACK_ENABLE 0
+#else 
+  #define TX_CALLBACK_ENABLE 1
+#endif
 
 // Attempts to transmit the specified message using the HW TX, overwriting the
 // source device.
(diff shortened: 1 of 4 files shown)


libraries/ms-common/src/can.c
```diff
@@ -23,6 +23,13 @@
 
 #define CAN_BUS_OFF_RECOVERY_TIME_MS 500
 
+// flag enabled if x86 passed as platform
+#ifdef X86
+#define TX_CALLBACK_ENABLE 0
+#else
+#define TX_CALLBACK_ENABLE 1
+#endif
```
copperium, line 31: I'd prefer you use `true`/`false` instead of `1`/`0` for booleans (use `#include <stdbool.h>` if you need to). Also explain why you're doing this! Here and/or right above `prv_tx_handler`, put the ticket number (SOFT-301) and briefly explain what the bug is and why this fixes it.

libraries/ms-common/src/can.c
```diff
@@ -145,14 +152,15 @@ bool can_process_event(const Event *e) {
...
 void prv_tx_handler(void *context) {
-  CanStorage *can_storage = context;
-  CanMessage tx_msg;
-
-  // If we failed to TX some messages or aren't transmitting fast enough, those
-  // events were discarded. Raise a TX event to trigger a transmit attempt. We
-  // only raise one event since TX ready interrupts are 1-to-1.
-  if (can_fifo_size(&can_storage->tx_fifo) > 0) {
-    event_raise(can_storage->tx_event, 0);
+  if (TX_CALLBACK_ENABLE) {
```
copperium, line 155: Like previous comment, please explain here what the bug is, including the ticket number, and why this fixes it.

libraries/ms-common/test/test_can.c
```diff
@@ -275,3 +278,49 @@ void test_can_default(void) {
   TEST_ASSERT_EQUAL(msg.msg_id, rx_msg.msg_id);
   TEST_ASSERT_EQUAL(msg.data, rx_msg.data);
 }
+
+#ifdef X86
+void test_can_x86_tx(void) {
+  volatile CanMessage rx_msg = { 0 };
+  CanMessage msg = {
+    .msg_id = 0xF,              //
```
copperium, line 286: Any CAN message with ID below 14 (0xE) is critical, and 0xF=15 is a bit too close for comfort for me - if we ever increase the number of critical messages, this test will break. I assume you didn't use 0xFF because no ID can be greater than 64, but I'd prefer you use 63=0x3F or similar to make the test a bit less brittle. I'd also recommend you factor out this CAN message ID as a #define like `NON_CRITICAL_MSG_ID`.

libraries/ms-common/test/test_can.c
```diff
@@ -275,3 +278,49 @@ void test_can_default(void) {
...
+    .type = CAN_MSG_TYPE_DATA,  //
+    .data = 0x1,                //
+    .dlc = 1,                   //
+  };
+
+  can_register_rx_handler(0xF, prv_rx_callback, &rx_msg);
+  s_rx_cb_count = 0;
```
copperium, line 293: Resetting `s_rx_cb_count` should be done in `setup_test`, even if it's only used here.

libraries/ms-common/test/test_can.c
```diff
@@ -275,3 +278,49 @@ void test_can_default(void) {
...
+
+  for (uint8_t i = 0x0; i < 0xA; i++) {
```
copperium, line 295: Why 0xA instead of 10? I'd also prefer you put the number of messages into a `const int` and use it instead of hardcoding 10 everywhere.

libraries/ms-common/test/test_can.c
```diff
@@ -275,3 +278,49 @@ void test_can_default(void) {
...
+      tx_msg_count++;
+    }
+    delay_ms(10);
+    if (e.id == TEST_CAN_EVENT_RX) {
+      rx_msg_count++;
+    }
+    loop_count++;
+    if (loop_count > 30) {
+      TEST_FAIL();
+    }
```
copperium, line 318: You can just do `TEST_ASSERT(loop_count <= 30)` (or 3 * the constant you define)

libraries/ms-common/test/test_can.c
```diff
@@ -275,3 +278,49 @@ void test_can_default(void) {
...
+  }
+  TEST_ASSERT_EQUAL(10, rx_msg_count);
+  TEST_ASSERT_EQUAL(10, tx_msg_count);
+  TEST_ASSERT_EQUAL(10, s_rx_cb_count);
+  delay_ms(300);
```
copperium, line 323: Maybe lower this to 100ms or so so it doesn't increase CI times too much

copperium (changes requested):
Looks good! Just a few nitpicks and documentation requests, great job Mitchell!

libraries/ms-common/src/can.c
```diff
@@ -145,14 +152,17 @@ bool can_process_event(const Event *e) {
...
-  CanStorage *can_storage = context;
-  CanMessage tx_msg;
-
-  // If we failed to TX some messages or aren't transmitting fast enough, those
-  // events were discarded. Raise a TX event to trigger a transmit attempt. We
-  // only raise one event since TX ready interrupts are 1-to-1.
-  if (can_fifo_size(&can_storage->tx_fifo) > 0) {
-    event_raise(can_storage->tx_event, 0);
+  // following condition used to disable tx events being re-raised on x86
+  // as this causes a race condition: see soft_301_race_condition_x86_can
```
copperium, line 156: By the ticket I did mean the Jira ticket designation like `SOFT-301` (`soft_301_race_condition_x86_can` is the branch, which will be deleted from github after the PR is merged, so it won't be useful to future devs)

PR #171, opened 2020-09-26
[SOFT-317] Setup work for Babydriver

copperium:
This gets Babydriver ready for specific modules to be implemented. We've got:
* a `dispatcher` C module which allows registering handlers for babydriver messages like the `can_register_rx_handler` API, and also automates responding with a status message
* a Python function in `can_util` for waiting for the next babydriver CAN message with a timeout
* changing the channel between can0 and vcan0 automatically: can0 for `make babydriver`, vcan0 for `make babydriver PLATFORM=x86`
* and implementing Gerald's suggestion on how to run the firmware project in the same terminal as the Python shell on x86

I had to rename `can_message.py` to `can_util.py` to appease pylint, which was confusing it with `can_message.py` from the `can_dump` project.


projects/baby_driver/scripts/can_util.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Raises:
+        can.CanError: If there was an error in transmitting the message.
+    """
+
+    if len(data) > 7:
+        raise ValueError("Only 7 bytes of data may be sent")
+    if len(data) < 7:
+        # pad to 7 bytes so that the firmware project will accept it
+        data += [0] * (7 - len(data))
```
g-jessmuir, line 58: should data=None be handled here? I dunno python, does None + list = list? I guess the function could just require passing an empty list, but seems conveniently descriptive to pass None for data

projects/baby_driver/scripts/can_util.py
```diff
@@ -0,0 +1,129 @@
...
+        data,
+        channel=None,
+        msg_id=BABYDRIVER_CAN_MESSAGE_ID,
+        device_id=BABYDRIVER_DEVICE_ID,
+):
+    """Sends a CAN message.
+
+    Args:
+        babydriver_id: The babydriver ID (first byte of message data) of the message to send.
+        data: Up to 7 bytes (0-255) of data to send in the CAN message.
```
g-jessmuir, line 45: if you can set the message id to be not the bay driver message id, why is len(data) capped at 7?

projects/baby_driver/scripts/can_util.py
```diff
@@ -0,0 +1,129 @@
...
+    if msg is None:
+        raise TimeoutError()
+
+    if babydriver_id is not None and (not msg.data or msg.data[0] != babydriver_id):
+        raise ValueError("next_message expected babydriver ID {} but got {}".format(
+            babydriver_id,
+            msg.data[0] if msg.data else "empty message",
+        ))
+
+    return msg.data
```
g-jessmuir, line 129: do we want to return more than just message data? you can wait on multiple messages, right? so I think you might want to know the message id that you got

projects/baby_driver/scripts/message_defs.py
```diff
@@ -0,0 +1,19 @@
+"""Constants for the babydriver CAN protocol."""
```
g-jessmuir, line 1: do we want to generate this file via codegen? protobufs are pretty cool (could also leave this as an improvement to make later (the more tickets the merrier right))

projects/baby_driver/inc/babydriver_msg_defs.h
```diff
@@ -0,0 +1,11 @@
...
+
+// Definitions of the IDs that can go in the first slot of a babydriver CAN message.
+
+typedef enum {
+  // The status message, sent whenever an operation is finished to indicate its status.
+  // Message data: uint8 status.
+  BABYDRIVER_MESSAGE_STATUS = 0,
+
+  NUM_BABYDRIVER_MESSAGES,
+} BabydriverMessageId;
```
g-jessmuir, line 11: we could codegen this file with protobufs? I would love it if you somehow got protobufs-c into our infra :))) otherwise I don’t actually know how our C codegen works lol

projects/baby_driver/scripts/repl_setup.py
```diff
@@ -0,0 +1,29 @@
+"""
+This module is run just before the Python REPL opens when you run `make babydriver`.
+It imports everything that's visible to the REPL by default and sets up the default CAN channel.
+"""
+
+# pylint: disable=unused-import
+from can_util import get_bus, send_message, next_message
+from message_defs import BABYDRIVER_DEVICE_ID, BABYDRIVER_CAN_MESSAGE_ID, BabydriverMessageId
```
g-jessmuir, line 8: shouldn’t these be abstracted into the driver APIs? when would you need to use them directly? not super sure but haven’t thought about it much

projects/baby_driver/scripts/repl_setup.py
```diff
@@ -0,0 +1,29 @@
...
+
+def setup_default_channel():
+    """Set up the default CAN channel."""
+
+    # We import these here so that they aren't imported for the REPL
+    # pylint: disable=import-outside-toplevel
+    import argparse
+    import can_util
+
+    parser = argparse.ArgumentParser(description="Setup the Babydriver REPL")
```
g-jessmuir, line 19: I dunno where it should go but a future improvement would be to link the confluence docs upon startup

projects/baby_driver/test/test_dispatcher.c
```diff
@@ -0,0 +1,126 @@
...
+
+  TEST_ASSERT_EQUAL(1, s_times_callback_called);
+  TEST_ASSERT_EQUAL(TEST_BABYDRIVER_MESSAGE_ID, s_received_data[0]);
+  TEST_ASSERT_EQUAL(&arbitrary_context, s_received_context);
+
+  // use a pointer to the second element to compare |data| with the non-ID part of |s_received_data|
+  uint8_t *received_data = &s_received_data[1];
+  TEST_ASSERT_EQUAL_UINT8_ARRAY(data, received_data, 7);
+
+  // should not tx result, so no CAN events
```
g-jessmuir, line 75: nit: should explicitly set s_should_tx_result for more clarity (I had to scroll back up to understand why it shouldn’t to result) but also I’m just lazy

g-jessmuir (approved):
TIL I can review PRs from my phone in bed

I think I just don’t understand some stuff, would appreciate clarification but nothing seems blocking

libraries/codegen-tooling/inc/can_pack.h
```diff
@@ -258,6 +258,8 @@
                    CAN_PACK_IMPL_EMPTY, CAN_PACK_IMPL_EMPTY, CAN_PACK_IMPL_EMPTY,         \
                    CAN_PACK_IMPL_EMPTY)
 
-#define CAN_PACK_SOLAR_DATA(msg_ptr, data_point_type_u32, data_value_u32)                 \
```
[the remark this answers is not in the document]
copperium: Huh, I guess the last PR updating the codegen files had solar data below solar fault for some reason when it's defined oppositely in the asciipb file. This order should be correct per can_messages.asciipb.

projects/baby_driver/scripts/can_util.py
copperium: Hm, Python-CAN provides a `CanMessage` class we can use, perhaps I'll return an object of that class with extra `message_id` and `device_id` properties set.

projects/baby_driver/inc/babydriver_msg_defs.h
copperium: it would be super cool to codegen the babydriver messages for python and C! probably something for another ticket though

platform/stm32f0xx/platform.mk
```diff
@@ -44,7 +44,11 @@ OPENOCD_CFG := -s $(OPENOCD_SCRIPT_DIR) \
                -c 'stm32f0x.cpu configure -rtos FreeRTOS'
 
 # Platform targets
-.PHONY: program gdb target
+.PHONY: program gdb target babydriver
+
+babydriver:
+	@make program PROJECT=baby_driver
+	@python3 -i projects/baby_driver/scripts/repl_setup.py --channel can0
```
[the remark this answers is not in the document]
copperium: added an optional `CHANNEL` argument to override the default CAN channel

projects/baby_driver/scripts/can_util.py
copperium: Added a `Message` class to wrap `can.Message` since it turns out we can't set properties on instances of that class

g-jessmuir (approved):

PR #173, opened 2020-09-26
Soft 999 Completed timer_challenge project

danielzyy:


projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,42 @@
+#include "interrupt.h"   // interrupts are required for soft timers
+#include "log.h"         // for printing
+#include "soft_timer.h"  // for soft timers
+#include "wait.h"        // for wait function
+
+#define COUNTER_PERIOD 500
```
g-jessmuir, line 6: normally we add the suffix `_MS` to denote the units of the value (milliseconds here)

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+
+  // start the timer again, so it keeps periodically incrementing every half a second
+  soft_timer_start_millis(COUNTER_PERIOD, prv_timer_callback, storage, NULL);
+}
+
+int main(void) {
+  interrupt_init();   // interrupts must be initialized for soft timers to work
+  soft_timer_init();  // soft timers must be initialized before using them
+
+  Counters storage = { 0, 0 };  // we use this to initialize a struct to be all 0
```
g-jessmuir, line 31: nit: The `{ 0 }` is actually the convention, we don't need to set all values to zero here

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+#include "wait.h"        // for wait function
+
+#define COUNTER_PERIOD 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
g-jessmuir, line 13: nit: private functions should actually be marked `static` to ensure they're not called elsewhere

g-jessmuir (changes requested):
Looks really good, just a few tiny things before calling it done 😃

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+  storage->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  if (storage->counter_a % 2 == 0) {
+    storage->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b);
+  }
+
+  // start the timer again, so it keeps periodically incrementing every half a second
+  soft_timer_start_millis(COUNTER_PERIOD, prv_timer_callback, storage, NULL);
```
g-jessmuir, line 24: you missed updating these values, so the automated build is failing since it doesn't compile

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+}
+
+int main(void) {
+  interrupt_init();   // interrupts must be initialized for soft timers to work
+  soft_timer_init();  // soft timers must be initialized before using them
+
+  Counters storage = { 0 };  // we use this to initialize a struct to be all 0
+
+  soft_timer_start_millis(COUNTER_PERIOD,      // timer duration
```
g-jessmuir, line 33: here too. It's always good to compile locally before pushing 🙂

g-jessmuir (changes requested):

g-jessmuir:
looks good, closing as complete 👍

PR #174, opened 2020-09-27
Soft 999 - Using Soft Timers

sudhish3120:


projects/usingSoftTimers/src/main.c
```diff
@@ -0,0 +1,51 @@
+
```
g-jessmuir, line 1: #pragma once is only used in header files; this is to prevent headers from being included multiple times. It's unnecessary for .c files.

projects/usingSoftTimers/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+#pragma once
+
+// Created by Sudhish M on 2020-09-25.
+// usingSoftTimers
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
```
g-jessmuir, line 9: we generally don't include these headers because they're included in other library headers (sometimes `stdint.h` is used but not the others)

projects/usingSoftTimers/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+#define COUNTER_PERIOD_MS 500
+
+typedef struct counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  // &counterTest=counterPtr
```
g-jessmuir, line 24: please be either more clear with comments or remove unnecessary comments

projects/usingSoftTimers/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+  Counters *counterPtr = context;
```
g-jessmuir, line 25: please use `snake_case` for variable names

g-jessmuir (changes requested):
Looks good! just a few nitpicks to change, but the logic is correct. Once the changes are made we'll mark it as done.

g-jessmuir:
Looks good! closing as complete. Please move on to fw103

PR #175, opened 2020-09-30
[SOFT-313] Implement bms main

g-jessmuir:
Missing a few values from hardware.


projects/bms_carrier/src/main.c
```diff
@@ -1,3 +1,89 @@
...
+  bps_heartbeat_init(&s_bms_storage.bps_storage, BPS_HB_FREQ_MS);
+  can_handler_init(&s_bms_storage, TIME_BETWEEN_TX_IN_MILLIS);
+  // TODO(SOFT-61): fill in these values from hardware
+  CellSenseSettings cell_settings = {
+    .undervoltage_dmv = 0,
+    .overvoltage_dmv = 0,
+    .charge_overtemp_dmv = 0,
+    .discharge_overtemp_dmv = 0,
+  };
+  cell_sense_init(&cell_settings, &s_bms_storage.afe_readings, &s_bms_storage.ltc_afe_storage);
```
copperium, line 55: `ltc_afe` needs to be initialized before this right?

projects/bms_carrier/src/main.c
```diff
@@ -1,3 +1,89 @@
...
+    .i2c_read_addr = BMS_FAN_CTRL_1_I2C_ADDR,
+  };
+  // We have two fan controllers
+  fan_control_init(&fan_settings, &s_bms_storage.fan_storage_1);
+  fan_settings.i2c_write_addr = BMS_FAN_CTRL_2_I2C_ADDR;
+  fan_settings.i2c_read_addr = BMS_FAN_CTRL_2_I2C_ADDR;
+  fan_control_init(&fan_settings, &s_bms_storage.fan_storage_2);
+  fault_bps_init(&s_bms_storage);
+  killswitch_init(&s_bms_storage.killswitch_storage);
+  relay_sequence_init(&s_bms_storage.relay_storage);
```
copperium, line 78: This needs the MCP23008 initialized right?

projects/bms_carrier/src/main.c
```diff
@@ -1,3 +1,89 @@
...
+
+  Event e = { 0 };
+  while (true) {
+    while (event_process(&e) == STATUS_CODE_OK) {
+      can_process_event(&e);
+      cell_sense_process_event(&e);
+    }
```
copperium, line 85: might want a `wait()` in there to save a few CPU cycles (maybe it should be part of the standard?)

projects/bms_carrier/src/main.c
```diff
@@ -1,3 +1,89 @@
...
+    .callback_context = NULL,
+    .i2c_settings = i2c_settings,
+    .i2c_write_addr = BMS_FAN_CTRL_1_I2C_ADDR,
+    .i2c_read_addr = BMS_FAN_CTRL_1_I2C_ADDR,
+  };
+  // We have two fan controllers
+  fan_control_init(&fan_settings, &s_bms_storage.fan_storage_1);
+  fan_settings.i2c_write_addr = BMS_FAN_CTRL_2_I2C_ADDR;
+  fan_settings.i2c_read_addr = BMS_FAN_CTRL_2_I2C_ADDR;
+  fan_control_init(&fan_settings, &s_bms_storage.fan_storage_2);
```
copperium, line 75: Something to think about: the ADT7476A driver (initialized by fan_control) reinitializes I2C, so I2C will be initialized a total of 3 times in this main. I don't *think* this will cause any impossible-to-debug stm32-only bugs, but it's possible. In any case ADT7476A shouldn't be doing that and should just require I2C to be initialized.

projects/bms_carrier/src/main.c
```diff
@@ -1,3 +1,89 @@
...
+#include "can_handler.h"
+#include "can_msg_defs.h"
+#include "event_queue.h"
+#include "fault_bps.h"
+#include "interrupt.h"
+#include "killswitch.h"
+#include "soft_timer.h"
+
+static CanStorage s_can_storage = { 0 };
+static BmsStorage s_bms_storage = { 0 };
```
copperium, line 13: I **love** the architecture of having all module-level state concentrated in a single `BmsStorage` struct! Reminds me of the Redux JS framework, it could be really useful for debugging.

copperium (changes requested):
I think you forgot to initialize a couple things, otherwise LGTM

copperium:
Looks like it needs a `make format` then it should be good.

copperium:
A couple `cell_sense` tests seem to be failing

copperium (approved):

PR #176, opened 2020-10-03
[Soft-308] Implement can pack function

mitchellostler:
Created a function that converts a list of int tuples (val, len_in_bytes) to a bytearray. 
Wrote a test suite for this function, using unittest. 
I couldn't really figure out parameterized so just wrote test using the basic format, but everything looks like it's working right.


projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
         ))
 
     return msg
+
+def can_pack(data_list):
+    """
+    Converts list of tuples in form ((int) val, (int) len_in_bytes)
+            and combines them into a bytearray rendition
```
copperium, line 190: Make sure you specify the endianness of the output bytes representing each `val`; after these changes it'll be little-endian order. You also might want an example here, and explain more that each `val` is broken into `len_in_bytes` bytes in little-endian order, then the bytes are concatenated in a `bytearray`.

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+    Returns:
+        A bytearray object representing message components passed in
+    Raises:
+        ValueError: if tuple has more than 2 items
```
copperium, line 194: Generally it's not recommended to document the behaviour of a function under conditions breaking its public API, because then paradoxically that becomes part of the public API. This is a rather philosophical point and doesn't matter much for this case, but you might consider removing this.

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+        ValueError: if tuple value exceeds allotted length in bytes
```
copperium, line 195: You might want to explain what this means a bit more

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+    """
+    to_bytes = ""
+    for item in data_list:
+        if len(item) > 2:
+            raise ValueError("Tuple length {}: expected 2".format(len(item)))
+        #store value and number of bytes allotted
+        val = item[0]
+        len_in_bytes = item[1]
```
copperium, line 203: Python tip: you can just do `for val, len_in_bytes in data_list:`, it'll unpack the tuples automatically.

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+        #check that values are valid
+        if len_in_bytes < 1 or val < 0:
+            raise ValueError("len in bytes must be > 0; val must be non-negative")
```
copperium, line 206: You should document that val can't be negative in the docstring (probably just say it must be in `[0, 2**len_in_bytes - 1]`).

(It could be a possible improvement to support negative numbers in 2's complement as well!)

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+        ValueError: if tuple value exceeds allotted length in bytes
+    """
+    to_bytes = ""
+    for item in data_list:
+        if len(item) > 2:
+            raise ValueError("Tuple length {}: expected 2".format(len(item)))
+        #store value and number of bytes allotted
+        val = item[0]
+        len_in_bytes = item[1]
+        #check that values are valid
```
copperium, line 204: Please put a space after the `#` in your comments!

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+        if val >= pow(2, len_in_bytes * 8):
+            raise ValueError("Value {} exceeds allotted {} bytes. Max Val: {}".format(
+                val, len_in_bytes, pow(2, len_in_bytes * 8) - 1))
+        hex_len = 2 * len_in_bytes
+        hex_str = '{0:x}'.format(val)
+        #prepend zeros to match number of bytes allotted
+        while len(hex_str) < hex_len:
+            hex_str = '0' + hex_str
+        to_bytes += hex_str
+    return bytearray.fromhex(to_bytes)
```
copperium, line 216: Woah, why are you converting to a hex string in order to convert to a bytearray? I think a better way to do it would be:
* keep a result bytearray, initialized to `bytearray()` (the empty bytearray) at the start
* for each `val, len_in_bytes` pair, do this `len_in_bytes` times: extract the lowest byte (you can do this with `& 0xFF`), push it onto the end of the result bytearray (there's an `append` method), and shift `val` rightwards a byte with `>>`. This will push the bytes in little-endian order.
* then return the result bytearray

This puts the results in little-endian format automatically and avoids the intermediary of a hex string.

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+
```
copperium, line 217: Pylint didn't catch the trailing whitespace at the end of this line for some reason, please remove it.

projects/baby_driver/scripts/test_can_util.py
```diff
@@ -0,0 +1,53 @@
+""" This Module Tests methods in can_util.c """
+import unittest
+
+from can_util import can_pack
+
+class TestCanPack(unittest.TestCase):
+    """ Test Can Util data_pack function """
+
+    def test_low_bound(self):
+        """ Tests Maximum values """
```
copperium, line 10: You mean tests *minimum* values? Also the standard is to not use spaces around the `"""`, like `"""Tests minimum values"""`

projects/baby_driver/scripts/test_can_util.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual(b'\x00', can_pack(data_list=[(0, 1)]))
```
copperium, line 11: Pro tip: you don't need to specify the argument name like `data_list=`, especially since it doesn't really add anything here.

projects/baby_driver/scripts/test_can_util.py
```diff
@@ -0,0 +1,53 @@
+""" This Module Tests methods in can_util.c """
```
copperium, line 1: can_util.py not can_util.c

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,35 @@ def next_message(
...
+            raise ValueError("Tuple length {}: expected 2".format(len(item)))
+        #store value and number of bytes allotted
+        val = item[0]
+        len_in_bytes = item[1]
+        #check that values are valid
+        if len_in_bytes < 1 or val < 0:
+            raise ValueError("len in bytes must be > 0; val must be non-negative")
+        if val >= pow(2, len_in_bytes * 8):
+            raise ValueError("Value {} exceeds allotted {} bytes. Max Val: {}".format(
+                val, len_in_bytes, pow(2, len_in_bytes * 8) - 1))
```
copperium, line 209: Python tip: there's an exponentiation operator `**`, though in this case I might even prefer using `1 <<` instead of `2**` since it's bitwise stuff.

copperium (changes requested):
Nice job Mitchell, great job setting up the tests!

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,33 @@ def next_message(
...
+        in range [0, 2**len_in_bytes - 1]
+    Returns:
+        A bytearray object in little endian format, representing
+        message components input
+    Raises:
+        ValueError: if invalid values for val, len_in_bytes input
+    """
+    bytearr = bytearray()
+    # Traverse list in reverse order
+    for val, len_in_bytes in reversed(data_list):
```
copperium, line 203: Wait, perhaps I misexplained. By "little-endian order" I meant that each val should be encoded in little-endian order but the vals should still be in the order given in data_list. So `can_pack([(0xABCDEF, 4), (0x010203, 3)])` should be equal to `bytearray([0xEF, 0xCD, 0xAB, 0x00, 0x03, 0x02, 0x01])`. So there's no need to traverse the list in reverse order. Sorry for the confusion.

copperium (changes requested):
LGTM other than the one correctness thing, fix that and we should be good to go

projects/baby_driver/scripts/can_util.py
```diff
@@ -183,3 +183,33 @@ def next_message(
...
+    Args:
+        List of tuples in form ((int) val, (int) len_in_bytes). val must be
+        in range [0, 2**len_in_bytes - 1]
+    Returns:
+        A bytearray object in little endian format, representing
+        message components input
+    Raises:
+        ValueError: if invalid values for val, len_in_bytes input
+    """
+    bytearr = bytearray()
```
copperium, line 201: One more thing, sorry: could you change `can_pack` to return a list rather than a bytearray? Shouldn't be much of a change other than changing this line to use `[]` rather than `bytearray()` and then changing the documentation. The reason is that `send_message` is actually expecting a list rather than a bytearray, sorry.

copperium (changes requested):

copperium (approved):

PR #177, opened 2020-10-03
[NO-MERGE] Button and ADCs project for fw_103

danielzyy:
Not sure if I implemented interrupts correctly, but please let me know.

### DIFF
--- a/projects/button_adc/src/main.c
+++ b/projects/button_adc/src/main.c
@@ -7,11 +7,11 @@
 #include "soft_timer.h"
 #include "wait.h"
 
-static void prv_adc_callback(const GpioAddress *address, void *context) {
-  AdcChannel *analogChannel = context;
-  uint16_t analogPinData = 0;
-  adc_read_raw(*analogChannel, &analogPinData);
-  LOG_DEBUG("Analog Pin data: %d\n", analogPinData);
+static void prv_button_press_callback(const GpioAddress *address, void *context) {
+  GpioAddress *analog_pin = context;
+  uint16_t analog_pin_data = 0;
+  adc_read_converted_pin(*analog_pin, &analog_pin_data);
+  LOG_DEBUG("Analog Pin data: %d\n", analog_pin_data);
 }
 
 int main(void) {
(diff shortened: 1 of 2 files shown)


projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+static void prv_adc_callback(const GpioAddress *address, void *context) {
+  AdcChannel *analogChannel = context;
+  uint16_t analogPinData = 0;
+  adc_read_raw(*analogChannel, &analogPinData);
```
copperium, line 13: The homework asks for an ADC *converted* read, not a *raw* read. Look in adc.h for the difference.

projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  LOG_DEBUG("Analog Pin data: %d\n", analogPinData);
```
copperium, line 14: Please use `snake_case`, not `camelCase`. This goes for the rest of the camelCase occurrences too.

projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  gpio_it_init();
+
+  GpioAddress analogPin = { .port = GPIO_PORT_A, .pin = 6 };
+  GpioAddress button = { .port = GPIO_PORT_B, .pin = 2 };
+
+  GpioSettings pinSettings = {
+    GPIO_DIR_IN,        //
+    GPIO_STATE_LOW,     //
+    GPIO_RES_NONE,      //
+    GPIO_ALTFN_ANALOG,  //
```
copperium, line 30: Please use the GNU-style `.direction =`, `.state =`, etc when initializing settings structs like this. It makes it much easier for humans to parse and lets you move around the lines without changing the meaning. This also goes for the `InterruptSettings` struct a couple lines down.

projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+static void prv_adc_callback(const GpioAddress *address, void *context) {
```
copperium, line 10: The name of this function is a little misleading - it's not triggered by the `adc` library, it interacts with it. (The `adc` library actually can call callbacks - look in adc.h if you're interested.) Usually an "ADC callback" is triggered by the `adc` library, a "soft timer callback" is triggered by the `soft_timer` library, etc. So I might name this callback after the circumstances where it's called, i.e. `prv_button_press_callback` or something.

projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  };
+  gpio_init_pin(&analogPin, &pinSettings);
+  gpio_init_pin(&button, &button_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  AdcChannel analogChannel = NUM_ADC_CHANNELS;
+
+  adc_get_channel(analogPin, &analogChannel);
+  adc_set_channel(analogChannel, true);
```
copperium, line 48: Using ADC channels is alright since the tutorial probably hasn't been updated, but for future reference we do have brand new versions of the `adc` functions that take `GpioAddress` parameters instead of `AdcChannel` parameters so that you don't have to pass around an `AdcChannel`. (Once again, look in adc.h if you're interested.)

projects/button_adc/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+
+  LOG_DEBUG("ADC Channel: %d\n", analogChannel);
+
+  gpio_it_register_interrupt(&button, &itSetting, INTERRUPT_EDGE_FALLING, prv_adc_callback,
+                             &analogChannel);
```
copperium, line 53: It's alright to pass around an address to a local variable (`analogChannel`, which should be in snake_case) in this case since it's declared in main, but in the future never ever pass around pointers to local variables if there's ever the slightest chance that the function declaring the local variable might return before the pointer is used! C may or may not reclaim the stack space your variable is stored in and you'll get super strange bugs. You can make that kind of static-state variable global and `static` to alleviate the issue.

copperium (changes requested):
Nice job, a few stylistic nitpicks and one point about the specification. GPIO interrupts look to be used correctly.

projects/button_adc/src/main.c
danielzyy: An thanks! Just wondering, why does making the variable `static` alleviate the issue of unreclaimed stack space (does C just handle memory of static variables better than local ones)? Then is setting `GpioAddress *analog_pin = context;` ok in the case of the callback function just because the function is `void` or should this be avoided too?

danielzyy:
Does Travis CI require the code to be `make lint` and `make format` in order to pass the checks?

projects/button_adc/src/main.c
copperium: It's not the `static` keyword per se that makes it safe to pass around pointers to static globals, it's that they're global - `static` just makes the variable not visible from other modules, which is good practice for module-local variables and functions. The issue with pointers to local variables is that when a function returns, the memory that its local variables are stored in is marked as overwritable, so if you have a pointer to a local variable of a function that has returned, when you dereference the pointer, C might have overwritten the variable. Global variables are stored in a different area of memory and their space is only reclaimed when the program exits, so they're safe to pass around pointers to.

It's okay in this case to pass around the pointer to `analog_pin` because it's declared in `main`, so you know that its space will never be reclaimed until `main` returns and the program exits. Setting `GpioAddress *analog_pin = context` is ok in the callback function, that's exactly what the context is for!

copperium:
> Does Travis CI require the code to be `make lint` and `make format` in order to pass the checks?

Yes, you can at the end of `.travis.yml` (or by reading the Travis logs) that Travis will run `make test_format` and `make lint` and will fail if those discover any issues.

copperium:
LGTM, Travis would fail on the two lines you commented because of the lack of spacing around `+`. Generally, you should always do a `make format && make lint` before pushing to avoid this type of CI issue; if you set up your git hooks as described somewhere in https://uwmidsun.atlassian.net/l/c/bnF1kt08 you can do this automatically. Closing since homework is complete.

PR #178, opened 2020-10-04
[NO-MERGE] Soft 999 raiyan sayeed fw 103

DoctaRay:


projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  adc_read_converted(*adc_pin_channel, &adc_pin_data);
+  LOG_DEBUG("ADC data: %d\n", adc_pin_data);
+}
+
+void prv_register_interrupts(void) {
+  adc_get_channel(adc_pin_addr, &adc_pin_channel);
+  adc_set_channel(adc_pin_channel, true);
+
+  gpio_it_register_interrupt(&button_addr, &s_interrupt_settings, INTERRUPT_EDGE_RISING,
+                             prv_button_interrupt_handler, &adc_pin_channel);
```
copperium, line 40: The homework specification says:
> The adc reading should trigger whenever the button is released, not initially when the button is pressed.

`INTERRUPT_EDGE_RISING` means the interrupt will get triggered when the button (active-high) goes from unpressed to pressed, which isn't what you want. (Active-high means that the gpio line is high when the button is pressed and low when it's not pressed.)

projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+#include "soft_timer.h"
+
+static GpioAddress adc_pin_addr = { .port = GPIO_PORT_A, .pin = 6 };
+static GpioAddress button_addr = { .port = GPIO_PORT_B, .pin = 2 };
+
+static GpioSettings adc_pin_settings = {
+  GPIO_DIR_IN,
+  GPIO_STATE_LOW,
+  GPIO_RES_NONE,
+  GPIO_ALTFN_ANALOG,
```
copperium, line 17: Please use the GNU-style `.direction = GPIO_DIR_IN`, `.state = GPIO_STATE_LOW`, etc for initializing settings structs like this. It makes it easier to read and lets you move around the lines without affecting the meaning.

projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+};
+
+static InterruptSettings s_interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,       //
+  .priority = INTERRUPT_PRIORITY_NORMAL,  //
+};
+
+static AdcChannel adc_pin_channel = NUM_ADC_CHANNELS;
```
copperium, line 25: Stylistic nit: this variable name should be `s_adc_pin_channel` since we prefix static variables with `s_`.

projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+
+static void prv_button_interrupt_handler(const GpioAddress *address, void *context) {
+  AdcChannel *adc_pin_channel = context;
+  uint16_t adc_pin_data = 0;
+
+  adc_read_converted(*adc_pin_channel, &adc_pin_data);
+  LOG_DEBUG("ADC data: %d\n", adc_pin_data);
+}
+
+void prv_register_interrupts(void) {
```
copperium, line 35: Stylistic nit: this function should be marked `static` since it's a private function.

copperium (changes requested):
I see you have your FW 102 homework code here too. This is probably because you branched off your old `soft_999_raiyan_sayeed_fw_102` branch without first checking out master. In the future, please do a `git checkout master` before doing `git checkout -b <new_branch_name>` to create a new branch.

To fix this for this branch, you can follow https://stackoverflow.com/q/2369426, probably followed by a force push. Ping me if you're having trouble with the git complexities.

Other than the git stuff, good job! There's one correctness issue and a few stylistic nitpicks, but you should be good after you resolve them.

projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+GpioAddress adc_pin_addr = { .port = GPIO_PORT_A, .pin = 6 };
+GpioAddress button_addr = { .port = GPIO_PORT_B, .pin = 2 };
```
copperium, line 12: These should be marked `static` and prefixed with `s_` (and could also be `const`).

projects/hello_buttons_adcs/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+static void prv_button_interrupt_handler(const GpioAddress *address, void *context) {
+  uint16_t adc_pin_data = 0;
+
+  adc_read_converted(s_adc_pin_channel, &adc_pin_data);
+  LOG_DEBUG("ADC data: %d\n", adc_pin_data);
+}
+
+static void prv_register_interrupts(void) {
+  gpio_it_register_interrupt(&button_addr, &s_interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             prv_button_interrupt_handler, (void *)0);
```
copperium, line 37: Please use `NULL` instead of `(void *)0`. (Although I did prefer your solution of passing around a pointer to `s_adc_pin_channel` as the context, generally it's a good idea to minimize the use of global variables as much as possible.)

copperium (changes requested):
A couple of stylistic nitpicks, then we should be GTG

projects/hello_buttons_adcs/src/main.c
DoctaRay: Is there a general rule for using `static` and `const`s? I noticed in some projects like the pedai board `main.c`, the enums for storage were marked `static` and some the settings were marked `const` (but not the I2C settings).

copperium: In general, anything you don't want to be accessible from other modules should be `static` - the keyword makes it not visible. This includes pretty much any global variable and any private functions. For `const`, in general it's a good idea to mark anything `const` that shouldn't change, like settings. (In pedal board's main.c, the `can_settings` struct really should be `static`.)

copperium:
Looks good, closing this since homework is complete

PR #179, opened 2020-10-04
fw102 hw

nadia-nguyen:


projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,39 @@
+#include <stdint.h>  // for integer types
+
+#include "delay.h"       // for delays
```
g-jessmuir, line 3: Doesn't seem like you use this include, please remove

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  storage->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+  if (storage->counter_a % 2 == 0) {  // for every two increments in Counter A, counter_b++
+    storage->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b);
+  }
+
+  soft_timer_start_millis(500, prv_timer_callback, storage, NULL);  // restart timer
```
g-jessmuir, line 23: Please use a #define for 500, e.g. `#define TIMER_PERIOD_MS 500` at the top after the #includes

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+}
+
+int main() {
+  interrupt_init();   // interrupts need to be initialized for soft timers to work
+  soft_timer_init();  // soft timers need to be initialized before use
+
+  Counters storage = { 0 };  // initialize struct to be all 0
+
+  soft_timer_start_millis(500, prv_timer_callback, &storage, NULL);
```
g-jessmuir, line 32: same comment here as above about #define 500

g-jessmuir (changes requested):
Looks correct, just a few things:
- please don't include the files for the hello_world test project; you can avoid this by not 'git add'ing them before commiting, or just deleting them and commiting again
- just nitpicks for the actual assignment
Aside from that looks good 🙂

copperium:
Looks good, please move on to FW 103. Closing since homework is complete.

PR #180, opened 2020-10-04
[NO-MERGE] Firmware 103 Homework 

sudhish3120:


projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+// buttons_and_adc
+
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+
+static GpioAddress adc_pin_address{ .port = GPIO_PORT_A, .pin = 6 };
+static GpioAddress button_address{ .port = GPIO_PORT_B, .pin = 2 };
```
copperium, line 11: This is C, not C++, so the uniform initialization syntax won't work here. You'll need a `=` between the variable name and the initializer, like `s_adc_pin_address = { ... }`. Also, we prefix static variables with `s_` to make it clear that they're static.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+
+static GpioSettings button_settings = {
+  .direction = GPIO_DIR_IN,
+  .resistor = PIO_RES_PULLDOWN,
+  // Reason for using pulldown: when no signal is received
+  // from the button (unpressed), the value sent to the pin is 0
```
copperium, line 17: You have a typo here: `PIO_RES_PULLDOWN` will need to be `GPIO_RES_PULLDOWN` to compile. Also generally the button will pull the signal down automatically when it's unpressed, but good thought! (And of course, remember the `s_` on `button_settings`.)

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+
+static GpioSettings adc_pin_settings = {
+  .direction GPIO_DIR_IN, .resistor = GPIO_RES_NONE, .alt_function = GPIO_ALTFN_ANALOG
+};
+
+static InterruptSettings interrupt_settings =
+    {
+      .type = INTERRUPT_TYPE_INTERRUPT,
+      .priority = INTERRUPT_PRIORITY_NORMAL,
+    }
```
copperium, line 29: You'll need a semicolon here, and make sure you run `make format` to normalize all the formatting.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  .direction = GPIO_DIR_IN,
+  .resistor = PIO_RES_PULLDOWN,
+  // Reason for using pulldown: when no signal is received
+  // from the button (unpressed), the value sent to the pin is 0
+  .alt_function = GPIO_ALTFN_NONE
+};
+
+static GpioSettings adc_pin_settings = {
+  .direction GPIO_DIR_IN, .resistor = GPIO_RES_NONE, .alt_function = GPIO_ALTFN_ANALOG
+};
```
copperium, line 23: If everything being on one line was the result of a `make format`, you can add `//` at the end of each line to force it not to pack all the lines; otherwise, I'd prefer each field to be on a different line, with a comma on the end of the last line (GCC supports this). Also, of course, don't forget the `s_` on `adc_pin_settings`.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+
+static GpioAddress adc_pin_address{ .port = GPIO_PORT_A, .pin = 6 };
+static GpioAddress button_address{ .port = GPIO_PORT_B, .pin = 2 };
+
+static GpioSettings button_settings = {
+  .direction = GPIO_DIR_IN,
+  .resistor = PIO_RES_PULLDOWN,
+  // Reason for using pulldown: when no signal is received
+  // from the button (unpressed), the value sent to the pin is 0
+  .alt_function = GPIO_ALTFN_NONE
```
copperium, line 18: Nit: I'd prefer a comma at the end here (GCC supports this) to make it easier to move things around and add fields.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+};
+
+static InterruptSettings interrupt_settings =
+    {
+      .type = INTERRUPT_TYPE_INTERRUPT,
+      .priority = INTERRUPT_PRIORITY_NORMAL,
+    }
+
+static void
+prv_callback(const GpioAddress *address, void *context) {
```
copperium, line 32: `make format` should take care of this, but we write the type and qualifiers on the same line as the function declaration, like `static void prv_callback(...) {`.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  uint16_t reading = 0;
+  GpioAddress *button_address = address;
```
copperium, line 34: You don't use this variable, so you don't need it; `make lint` will warn about this. (`make lint`, along with `make test_format`, is run in Travis CI on every commit, so you can't get around it.)

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  GpioAddress *adc_pin_address = context;
+  //  enable the adc pin and log the values it reads
+  adc_read_converted_pin(*adc_pin_address, &reading);
+  LOG_DEBUG("ADC Reading: %i", reading);
```
copperium, line 38: You're missing a `\n` at the end of the `LOG_DEBUG`.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  gpio_init();
+  interrupt_init();
+  adc_init(ADC_MODE_SINGLE);
+
+  gpio_init_pin(&adc_pin_address, &adc_pin_settings);
+  gpio_init_pin(&button_address, &button_settings);
+
+  adc_set_channel_pin(adc_pin_address, true);
+  // I did not use step 4 mentioned in firmware_103 notes because
+  // adc.h mentioned using the method illustrated in firmware_103 is deprecated.
```
copperium, line 51: Nice! This is absolutely true, we need to update the firmware 103 notes :)

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+
+  gpio_it_register_interrupt(&button_address, &interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             // callback is initiated when button's pin changes value to 0
+                             &prv_callback, &adc_pin_address);
```
copperium, line 55: Nitpick: putting a `&` in front of a function name is unnecessary in C (just passing `prv_callback` will function as a function pointer), so we don't do it for consistency.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,61 @@
+// Created by Sudhish M on 2020-10-04.
+// buttons_and_adc
+
```
copperium, line 3: This comment isn't really necessary, it's fine for a homework though.

copperium (changes requested):
I see you still have your FW 102 homework code in this PR, this is probably because you were still on your fw_102 branch when you created your fw_103 branch. You can fix it by looking here: https://stackoverflow.com/questions/2369426/how-to-move-certain-commits-to-be-based-on-another-branch-in-git (might need a force push, but that's ok).

Aside from that, the logic looks correct, but you do have some compile errors that will need to be fixed and some formatting that needs a `make format`; these will need to be fixed before Travis will pass the PR / the homework can be done.

Always make sure your code compiles and links by running `make build PLATFORM=x86 PROJECT=buttons_and_adc`; you can also run it on x86 with `make run PLATFORM=x86 PROJECT=buttons_and_adc`, though that won't do much for this project since you can't exactly push the button in the simulation. Also run `make format` and `make lint` over your code; `make format` will format it and any errors that `make lint` uncovers will need to be fixed.

projects/buttons_and_adc/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+static InterruptSettings s_interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,       //
+  .priority = INTERRUPT_PRIORITY_NORMAL,  //
+};
+
+static void prv_callback(const GpioAddress *address, void *context) {
+  uint16_t reading = 0;
+  GpioAddress *s_adc_pin_address = context;
+
+  adc_read_converted_pin(*s_adc_pin_address, &reading);
```
copperium, line 32: Here `s_adc_pin_address` is a local variable, so it shouldn't be prefixed with `s_`; furthermore, it's shadowing the global `s_adc_pin_address`, so it definitely needs a different name.

copperium (commented):
LGTM, please read the one comment but I think we can call this done. Closing since homework is complete.

PR #181, opened 2020-10-06
[SOFT-265] Add fault handling + EN pin functionality to BTS7200 driver

hewittmcg:
Note: Still working with HW on figuring out the ADC readings we should be getting on a fault, so the current values are placeholders.


projects/power_distribution/src/current_measurement.c
```diff
@@ -69,7 +69,10 @@ StatusCode power_distribution_current_measurement_init(PowerDistributionCurrentS
       return status_code(STATUS_CODE_INVALID_ARGS);
     }
 
+    // Add DSEL, EN0, EN1 pins
```
copperium, line 72: Might want to add a TODO here to convert power distro to using the bts7200 + (eventually) bts7040 drivers for toggling the enable pins which will take quite a lot of restructuring, feel free to make a new ticket if there isn't an appropriate one already

libraries/ms-drivers/inc/bts_7200_current_sense.h
```diff
@@ -57,8 +97,29 @@ StatusCode bts_7200_init_pca9539r(Bts7200Storage *storage, Bts7200Pca9539rSettin
...
+StatusCode bts_7200_enable_output_1(Bts7200Storage *storage);
+
+// Disable output 1 by pulling the IN1 pin low.
+StatusCode bts_7200_disable_output_1(Bts7200Storage *storage);
+
+// Return whether output 0 is enabled or disabled
+bool bts_7200_get_output_0_enabled(Bts7200Storage *storage);
+
+// Return whether output 1 is enabled or disabled
+bool bts_7200_get_output_1_enabled(Bts7200Storage *storage);
```
copperium, line 121: You might consider more fully replicating the `gpio` interface - i.e. define an enum analogous to `GpioState`, define set, toggle, and get methods (each for output 0 and output 1, or maybe take that as a parameter, up to you) and take the state enum as a parameter to the set method. Up to you.

copperium (changes requested):
Haven't looked at the tests yet, but a bit of feedback. Logic looks mostly good.

libraries/ms-drivers/src/bts_7200_current_sense.c
```diff
@@ -61,23 +137,95 @@ StatusCode bts_7200_init_pca9539r(Bts7200Storage *storage, Bts7200Pca9539rSettin
...
+bool bts_7200_get_output_1_enabled(Bts7200Storage *storage) {
+  if (storage->select_pin_type == BTS7200_SELECT_PIN_STM32) {
+    GpioState pin_state;
+    gpio_get_state(storage->enable_1_pin_stm32, &pin_state);
+    return (pin_state == GPIO_STATE_HIGH);
+  } else {
+    Pca9539rGpioState pin_state;
+    pca9539r_gpio_get_state(storage->enable_1_pin_pca9539r, &pin_state);
+    return (pin_state == PCA9539R_GPIO_STATE_HIGH);
+  }
```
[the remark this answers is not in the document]
hewittmcg: Typically I do this for readability -- IIRC Google's/our style guide says to use brackets in cases like this, but it's somewhat ambiguous (see [here](https://google.github.io/styleguide/cppguide.html#Return_Values)).  Is there a standard/general rule of thumb we follow for this across the codebase?

libraries/ms-drivers/src/bts_7200_current_sense.c
```diff
@@ -96,6 +244,19 @@ StatusCode bts_7200_get_measurement(Bts7200Storage *storage, uint16_t *meas0, ui
...
+  if (fault0 || fault1) {
+    // Only call fault cb if it's not NULL
+    if (storage->fault_callback) {
+      storage->fault_callback(fault0, fault1, storage->fault_callback_context);
+    }
+    // Handle fault
+    prv_bts_7200_handle_fault(storage, fault0, fault1);
+  }
+
   return STATUS_CODE_OK;
```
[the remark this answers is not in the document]
hewittmcg: I think I'll go with `STATUS_CODE_INTERNAL_ERROR` if a fault is detected and just call `status_ok_or_return` on `adc_read_raw` and `adc_get_channel` -- those shouldn't conflict, but `adc_read_raw` may return multiple different status codes depending on the specific error it encounters.  Let me know if you'd prefer a different implementation.

projects/power_distribution/src/current_measurement.c
hewittmcg: Sure, created [SOFT-336] for this issue

libraries/ms-drivers/src/bts_7200_current_sense.c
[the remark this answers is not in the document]
copperium: Huh, I don't think I've ever seen that really, but I suppose it's fine in this case.

libraries/ms-drivers/inc/bts_7200_current_sense.h
```diff
@@ -4,49 +4,104 @@
...
   Pca9539rGpioAddress *select_pin;
   GpioAddress *sense_pin;
+  Pca9539rGpioAddress *enable_0_pin;
+  Pca9539rGpioAddress *enable_1_pin;
   uint32_t interval_us;
   Bts7200DataCallback callback;  // set to NULL for no callback
   void *callback_context;
+  Bts7200FaultCallback fault_callback;
+  void *fault_callback_context;
+  int resistor;  // scaling performed by resistor at SENSE pin
```
copperium, line 88: Ah, I know I said int is okay but could you actually use a `uint32_t` or something like that (and in all the other places you set resistor values)? Sorry for the confusion.

You should also specify that it's in milliohms.

libraries/ms-drivers/src/bts_7200_current_sense.c
```diff
@@ -35,66 +43,250 @@ static StatusCode prv_init_common(Bts7200Storage *storage) {
...
+  } else {
+    (*meas0) *= BTS7200_IS_SCALING_NOMINAL;
+    (*meas0) /= storage->resistor;
+  }
+  if (*meas1 <= BTS7200_MAX_LEAKAGE_VOLTAGE_MV) {
+    *meas1 = 0;
+  } else {
+    (*meas1) *= BTS7200_IS_SCALING_NOMINAL;
+    (*meas1) /= storage->resistor;
+  }
```
copperium, line 277: Nit: you could factor out this into a "convert from voltage to current" function. Also you don't need the brackets in `(*meas0/1)`.

libraries/ms-drivers/inc/bts_7200_current_sense.h
```diff
@@ -4,49 +4,104 @@
...
+#define BTS7200_MAX_LEAKAGE_VOLTAGE_MV 2
+
+// Nominal scaling factor k(ILIS) for current output at the SENSE pin in normal operation.
+#define BTS7200_IS_SCALING_NOMINAL 670
+
+// Voltage at the SENSE pin is limited to a max of 3.3V by a diode.
+// Due to to this function, since any fault current will be at least 4.4 mA (see p.g. 49)
+// the resulting voltage will be 4.4 mA * 1.6 kOhm = ~7 V. Due to this,
+// voltages approaching 3.3V represent a fault, and should be treated as such.
+#define BTS7200_MAX_VALID_SENSE_VOLTAGE_MV 3200
```
copperium, line 26: Hm, this looks like it's specific to the power distro board, not generic to every BTS7200? If so it'd be better to make this configurable in the settings so it can vary in future boards. (you could put a "suggested value" in a comment though with these justifications if that makes sense)

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -12,14 +12,57 @@
...
+static uint16_t s_adc_measurement_0 = ADC_EXPECTED_OK_VOLTAGE;
+static uint16_t s_adc_measurement_1 = ADC_EXPECTED_OK_VOLTAGE;
+// To give different measurements between adc readings in bts_7200_get_measurement
+static bool s_adc_measurement_number = 0;
+
 static volatile uint16_t times_callback_called = 0;
 static void *received_context;
 
+static volatile uint16_t s_times_fault_callback_called = 0;
+static void *fault_received_context;
```
copperium, line 35: All of these should be prefixed with `s_` (even though I realize 2 of them were preexisting)

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -12,14 +12,57 @@
...
+}
+
+// Storage is global so we can call bts_7200_stop to stop soft timers and avoid segfaults
+static Bts7200Storage s_storage = { 0 };
+
+// Mocks adc_read_converted to allow for changing the reading during testing.
+// Note that this removes some of the interrupt functionality of the x86 adc implementation,
+// but this shouldn't matter in this case.
+StatusCode TEST_MOCK(adc_read_converted)(AdcChannel adc_channel, uint16_t *reading) {
+  // Return reading based on pin it should corrsepond to.
```
copperium, line 54: minor nitpick: typo `corrsepond`

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -12,14 +12,57 @@
...
+// See bts_7200_current_sense.h for an explanation of this value
+#define ADC_TEST_FAULT_VOLTAGE 3300
+
+// Return ADC_EXPECTED_OK_VOLTAGE by default
+#define ADC_DEFAULT_RETURNED_VOLTAGE_RAW ADC_EXPECTED_OK_VOLTAGE
+
+static uint16_t s_adc_measurement_0 = ADC_EXPECTED_OK_VOLTAGE;
+static uint16_t s_adc_measurement_1 = ADC_EXPECTED_OK_VOLTAGE;
+// To give different measurements between adc readings in bts_7200_get_measurement
+static bool s_adc_measurement_number = 0;
```
copperium, line 29: Nit: I'd prefer you use `true`/`false` from `<stdbool.h>` instead of 0/1 for bools

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -34,31 +77,44 @@ void setup_test(void) {
   i2c_init(TEST_I2C_PORT, &i2c_settings);
 
   times_callback_called = 0;
+  s_times_fault_callback_called = 0;
+
+  s_adc_measurement_0 = ADC_EXPECTED_OK_VOLTAGE;
+  s_adc_measurement_1 = ADC_EXPECTED_OK_VOLTAGE;
```
copperium, line 83: Should you also reset the contexts and `s_adc_measurement_number` here too?

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -285,40 +394,516 @@ void test_bts_7200_current_sense_stop_return_behaviour(void) {
...
+  TEST_ASSERT_OK(bts_7200_start(&s_storage));
+
+  // Same tests as above:
+  TEST_ASSERT_EQUAL(bts_7200_get_output_0_enabled(&s_storage), false);
+  TEST_ASSERT_OK(bts_7200_enable_output_0(&s_storage));
+  TEST_ASSERT_EQUAL(bts_7200_get_output_0_enabled(&s_storage), true);
+  TEST_ASSERT_OK(bts_7200_disable_output_0(&s_storage));
+  TEST_ASSERT_EQUAL(bts_7200_get_output_0_enabled(&s_storage), false);
+  TEST_ASSERT_OK(bts_7200_disable_output_0(&s_storage));
+  TEST_ASSERT_EQUAL(bts_7200_get_output_0_enabled(&s_storage), false);
```
copperium, line 490: The order for the `TEST_ASSERT_` macros is `expected, actual`, so true/false should be first. However there are also `TEST_ASSERT_TRUE` and `TEST_ASSERT_FALSE` macros which really should be used instead (even though I realize there's a bunch of `TEST_ASSERT_EQUAL(true, ...)` in the file already)

libraries/ms-drivers/test/test_bts_7200_current_sense.c
```diff
@@ -285,40 +394,516 @@ void test_bts_7200_current_sense_stop_return_behaviour(void) {
...
+  s_times_fault_callback_called = 0;
+
+  // Wait for soft_timers to expire before continuing
+  // Why does having this cause a segfault????????????
+  delay_ms(BTS7200_FAULT_RESTART_DELAY_MS + 10);
+}
+
+// Test that the fault callback gets called when running bts_7200_start, and
+// there's no segfault/stack overflow.
+// This test causes a segfault for some reason
```
copperium, line 624: Is the segfault still an issue? Doesn't seem to happen for me locally

copperium (changes requested):
Awesome stuff! Mostly a bunch of nitpicks, none should take too long to address.

copperium (approved):

PR #183, opened 2020-10-07
Soft 180 smoke test temp sense

shahhet14:
Created a smoke test that periodically logs raw or converted data from the ADC pins connected to the thermistors.

### DIFF
--- a/projects/smoke_thermistor/src/main.c
+++ b/projects/smoke_thermistor/src/main.c
@@ -6,20 +6,21 @@
 #include "soft_timer.h"
 
 // Toggle the following to select which data is logged
-#define READ_DATA_RAW 1
-#define READ_DATA_CONVERTED 0
+#define READ_DATA_RAW true
+#define READ_DATA_CONVERTED false
+#define NUM_ADC_PINS 6
+#define TIME_BETWEEN_READS_IN_MILLIS 200
 
 int main(void) {
   interrupt_init();
   soft_timer_init();
   gpio_init();
 
-  GpioAddress temp1_addr = { .port = GPIO_PORT_A, .pin = 0 };
-  GpioAddress temp2_addr = { .port = GPIO_PORT_A, .pin = 1 };
-  GpioAddress temp3_addr = { .port = GPIO_PORT_A, .pin = 2 };
-  GpioAddress temp4_addr = { .port = GPIO_PORT_A, .pin = 3 };
-  GpioAddress temp5_addr = { .port = GPIO_PORT_A, .pin = 4 };
-  GpioAddress temp6_addr = { .port = GPIO_PORT_A, .pin = 5 };
+  GpioAddress pin_addresses[NUM_ADC_PINS];
+
+  for (uint8_t pin = 0; pin < NUM_ADC_PINS; pin++) {
+    pin_addresses[pin] = {.port = GPIO_PORT_A, .pin = pin }
+  }
 
   GpioSettings settings = {
     .direction = GPIO_DIR_IN,           //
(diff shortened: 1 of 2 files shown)


projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+
+  GpioAddress temp1_addr = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress temp2_addr = { .port = GPIO_PORT_A, .pin = 1 };
+  GpioAddress temp3_addr = { .port = GPIO_PORT_A, .pin = 2 };
+  GpioAddress temp4_addr = { .port = GPIO_PORT_A, .pin = 3 };
+  GpioAddress temp5_addr = { .port = GPIO_PORT_A, .pin = 4 };
+  GpioAddress temp6_addr = { .port = GPIO_PORT_A, .pin = 5 };
```
copperium, line 22: Please define these in an array! It will make the whole file a lot cleaner.

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+
+  AdcChannel temp5_channel = NUM_ADC_CHANNELS;
+  adc_get_channel(temp5_addr, &temp5_channel);
+  adc_set_channel(temp5_channel, true);
+  LOG_DEBUG("Temp5 ADC Channel: %d\n", temp5_channel);
+
+  AdcChannel temp6_channel = NUM_ADC_CHANNELS;
+  adc_get_channel(temp6_addr, &temp6_channel);
+  adc_set_channel(temp6_channel, true);
+  LOG_DEBUG("Temp6 ADC Channel: %d\n", temp6_channel);
```
copperium, line 67: Please use the new ADC API to use pins instead of channels - look at adc.h for more info. Also use a for loop instead of copy/pasting code 6 times! You can use the `SIZEOF_ARRAY` macro from misc.h on your gpio address array to get the upper bound for your for loops.

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+      adc_read_converted(temp3_channel, &temp_data);
+      LOG_DEBUG("Temp3 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp4_channel, &temp_data);
+      LOG_DEBUG("Temp4 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp5_channel, &temp_data);
+      LOG_DEBUG("Temp5 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp6_channel, &temp_data);
+      LOG_DEBUG("Temp6 converted ddata: %d\n", temp_data);
+    }
+    delay_ms(200);
```
copperium, line 99: This constant should be a `#define` at the top and configurable.

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+      LOG_DEBUG("Temp2 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp3_channel, &temp_data);
+      LOG_DEBUG("Temp3 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp4_channel, &temp_data);
+      LOG_DEBUG("Temp4 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp5_channel, &temp_data);
+      LOG_DEBUG("Temp5 converted ddata: %d\n", temp_data);
+      adc_read_converted(temp6_channel, &temp_data);
+      LOG_DEBUG("Temp6 converted ddata: %d\n", temp_data);
+    }
```
copperium, line 98: Please use for loops here instead of copy/pasting code!

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+    .resistor = GPIO_RES_NONE,          //
+    .alt_function = GPIO_ALTFN_ANALOG,  //
+  };
+
+  gpio_init_pin(&temp1_addr, &settings);
+  gpio_init_pin(&temp2_addr, &settings);
+  gpio_init_pin(&temp3_addr, &settings);
+  gpio_init_pin(&temp4_addr, &settings);
+  gpio_init_pin(&temp5_addr, &settings);
+  gpio_init_pin(&temp6_addr, &settings);
```
copperium, line 36: Use a for loop!

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,101 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+// Toggle the following to select which data is logged
+#define READ_DATA_RAW 1
+#define READ_DATA_CONVERTED 0
```
copperium, line 10: Please use `true`/`false` (actually macros defined in `<stdbool.h>`) instead of 1/0.

copperium (changes requested):

shahhet14:
Used the new ADC API to use pins instead of channels and used loops for steps that are repeated for each pin.

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+
+  GpioAddress pin_addresses[NUM_ADC_PINS];
+
+  for (uint8_t pin = 0; pin < NUM_ADC_PINS; pin++) {
+    pin_addresses[pin].port = GPIO_PORT_A;
+    pin_addresses[pin].pin = pin;
+  }
```
copperium, line 24: IIRC the thermistor pins are on A0-A5 on solar (though A365 is down for maintenance so I can't check, lol), but if we ever want to use this smoke test for other pins, this kind of an opaque way of initializing the addresses and it's hard to change the pins we're reading from. Instead I'd prefer `pin_addresses` to be a static array with an explicit list of the `GpioAddress`es to read from (i.e. `GpioAddress s_pin_addresses[] = { { GPIO_PORT_A, 0}, ... };`) so that it's easier to change.

projects/smoke_thermistor/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+#include "delay.h"
+#include "gpio.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+// Toggle the following to select which data is logged
+#define READ_DATA_RAW true
+#define READ_DATA_CONVERTED false
+#define NUM_ADC_PINS 6
```
copperium, line 11: With the change from the last comment, instead of hardcoding the number of pins separately from the array of pins, I'd prefer you just use `SIZEOF_ARRAY(s_pin_addresses)` every time instead of having a separate constant.

copperium (changes requested):

shahhet14:
Created a static array for the gpio pin addresses and used the size of the array as an upperbound for my for loops.

copperium (approved):

PR #185, opened 2020-10-08
completed timer challenge tutorial for firmware 102

Kyle02:
For some reason, my hello_world project was also loaded in, even though I tried to remove it when staging in git.

### DIFF
--- a/projects/timer_challenge/src/main.c
+++ b/projects/timer_challenge/src/main.c
@@ -1,0 +1,38 @@
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void counter_increment(SoftTimerId timer_id, void *context) {
+  Counters *counter = context;
+
+  counter->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", counter->counter_a);
+
+  if (counter->counter_a % 2 == 0) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", counter->counter_b);
+  }
+
+  soft_timer_start_millis(500, counter_increment, counter, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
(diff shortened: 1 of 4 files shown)


copperium:
FYI our git hooks (somewhat unfortunately) run `git add .` at some point before committing, so that's how your hello_world project snuck in. You can remove it by deleting the hello_world files then committing again.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void counter_increment(SoftTimerId timer_id, void *context) {
```
copperium, line 11: This is a private function, so it should be marked `static` (which makes it not visible from other files) and prefixed with `prv_`.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+
+  counter->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", counter->counter_a);
+
+  if (counter->counter_a % 2 == 0) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", counter->counter_b);
+  }
+
+  soft_timer_start_millis(500, counter_increment, counter, NULL);
```
copperium, line 22: Please use a `#define`'d constant with a descriptive name for the `500` here.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counter = { 0 };
+
+  soft_timer_start_millis(500, counter_increment, &counter, NULL);
```
copperium, line 31: Same comment about the `500` here.

copperium (changes requested):
Looks pretty good, just a couple comments (+ removing hello_world) to address before moving on to Firmware 103.

copperium:
Going to close this since there hasn't been any activity and you've moved on anyways.

PR #186, opened 2020-10-10
[SOFT-201] Cruise Control

GeraldNDA:
It is what it is


GeraldNDA:
Pretty sure these tests are incomplete, so lemme know if you notice cases I missed. Also I'll fix formatting and linting tonight/tomorrow

projects/centre_console/src/main.c
```diff
@@ -52,6 +52,7 @@ void prv_init_fsms() {
   power_off_sequence_init(&s_off_sequence_storage);
   power_fsm_init(&s_power_fsm_storage);
   drive_fsm_init(&s_drive_fsm_storage);
+  cruise_rx_init();
```
g-jessmuir, line 55: is this necessary?

g-jessmuir (approved):
only thing I really see is testing increase/decrease at min/max limits for cruise, but it's probably fine, and linting I guess

projects/centre_console/src/main.c
GeraldNDA: oof mistake lol

PR #187, opened 2020-10-12
[SOFT-151] Fix MCI

g-jessmuir:
1. Drive FSM wasn't previously initialized in main, meaning MCI wouldn't respond to changes in drive state from center console.
2. This issue was found by MPXE. Pedal rx library didn't set the storage's watchdog_id after the watchdog triggered, so if the watchdog triggered and then was later kicked the watchdog id could be out of date. This was causing the timer loop in mci_output to be cancelled.


copperium (approved):
LGTM, not sure if the `watchdog` library could be useful in pedal_rx but that's a question for another day.

PR #188, opened 2020-10-12
Firmware 103 ADC Homework

CheranMahalingam:
Buttons and ADC 103 homework. I am not completely sure whether my implementation of the ADC is correct.


projects/Button_ADC/src/main.c
```diff
@@ -0,0 +1,53 @@
...
+
+int main(void) {
+  adc_init(ADC_MODE_SINGLE);
+  interrupt_init();
+  gpio_init();
+  gpio_it_init();
+  soft_timer_init();
+
+  adc_get_channel(adc_address, &adc_pinA6_channel);
+  adc_set_channel(adc_pinA6_channel, true);
```
copperium, line 45: One thing you forgot - you'll have to initialize `adc_address` as an input with `alt_function` as `GPIO_ALTFN_ANALOG` before you can ADC-read from it; I think you set this up with `adc_settings` but forgot to actually initialize the pin.

projects/Button_ADC/src/main.c
```diff
@@ -0,0 +1,53 @@
...
+};
+
+static InterruptSettings interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+static AdcChannel adc_pinA6_channel = ADC_CHANNEL_6;
+
+static uint16_t channel_data = 0;
```
copperium, line 25: Please prefix static variables with `s_`, that way we can tell that they're static more easily. That being said, these global variables could be local variables, especially `channel_data`. You can avoid having a global variable for `adc_pinA6_channel` by passing a pointer to it around through the `context` parameter on interrupt callbacks; currently you're passing a pointer to `adc_address` around but not using it. It's good practice to minimize the use of global variables as much as possible.

projects/Button_ADC/src/main.c
```diff
@@ -0,0 +1,53 @@
...
+
+static void prv_button_interrupt_handler(const GpioAddress *address, void *context) {
+  adc_read_raw(adc_pinA6_channel, &channel_data);
```
copperium, line 28: The FW 103 homework directions do say:
> Create a project that reads an ADC **converted** reading...

This is a raw reading, look in adc.h for more details.

projects/Button_ADC/src/main.c
```diff
@@ -0,0 +1,53 @@
...
+  LOG_DEBUG("%d\n", channel_data);
+}
+
+void prv_register_interrupts(void) {
```
copperium, line 32: This function should be marked `static` since it's private.

projects/Button_ADC/src/main.c
```diff
@@ -0,0 +1,53 @@
...
+  .resistor = GPIO_RES_NONE,
+  .alt_function = GPIO_ALTFN_ANALOG,
+};
+
+static InterruptSettings interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+static AdcChannel adc_pinA6_channel = ADC_CHANNEL_6;
```
copperium, line 23: Nit: our variable names are generally all lowercase and we don't use lowerCamelCase at all (so this should rather be `adc_pin_a6_channel`). Also, why are you initializing it to `ADC_CHANNEL_6`?

copperium (changes requested):
Great job Cheran, a few things to address before the homework can be marked completed.

copperium:
LGTM, closing since homework is complete.

PR #189, opened 2020-10-13
Patrick Kim fw 102 hw

s524kim:
Pull request for Firmware 102 homework

### DIFF
--- a/projects/fw_102_hw/src/main.c
+++ b/projects/fw_102_hw/src/main.c
@@ -1,0 +1,43 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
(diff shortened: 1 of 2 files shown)


projects/fw_102_hw/src/main.c
```diff
@@ -0,0 +1,43 @@
...
+#include "wait.h"
+
+#define COUNTER_TICK_MS 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void priv_timer_callback(SoftTimerId timer_id, void *context) {
```
g-jessmuir, line 16: private functions should be marked `static`, also by convention we use `prv_` rather than `priv_` :)

g-jessmuir (changes requested):
Just a small nitpick, otherwise looks good 👍 once that fix is pushed we'll call it done

PR #191, opened 2020-10-14
fw 102 homework

rac99:
firmware 102 homework pull request

### DIFF
--- a/projects/fw_102_hw/src/main.c
+++ b/projects/fw_102_hw/src/main.c
@@ -1,0 +1,41 @@
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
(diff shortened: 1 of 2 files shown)


projects/fw_102_hw/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+  storage->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  if (storage->counter_a % 2 == 0) {
+    storage->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b);
+  }
+
+  // start another timer
+  soft_timer_start_millis(500, prv_timer_callback, storage, NULL);
```
copperium, line 25: Please move the constant 500 here and below into a `#define`'d macro symbol with a descriptive name including the unit. That makes it easier to tell what the 500 is as opposed to a magic number, and makes it easier to change by consolidating a constant that currently appears in 2 places into 1.

copperium (changes requested):
Awesome, just one nitpick before we can call it complete.

copperium:
LGTM, please move on to FW 103. Closing since homework is complete.

PR #194, opened 2020-10-15
testing communication protocols and can for fw104

danielzyy:


danielzyy:
Apparently, Travis does not like pushing unused functions. How should I push a file containing just functions?

copperium:
> Apparently, Travis does not like pushing unused functions. How should I push a file containing just functions?

Usually files containing just functions are part of modules with header files, so Travis doesn't complain. I'd say it's fine for homework purposes.

projects/com_protocol/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+  TEST_CAN_EVENT_TX,
+  TEST_CAN_EVENT_FAULT,
+} TestCanEvent;
+
+static CanStorage s_can_storage = { 0 };
+
+CanMessage messageA = {
+  .source_id = CAN_DEVICE_ID,
+  .msg_id = CAN_ID_A,
+  .data = 4015,  // random uint16_t becomes uint64_t (not sure how to prevent)
```
copperium, line 26: Here's the definition of `CanMessage` from `can_msg.h`:
```c
typedef struct CanMessage {
  uint16_t source_id;
  CanMessageId msg_id;
  union {
    uint64_t data;
    uint32_t data_u32[2];
    uint16_t data_u16[4];
    uint8_t data_u8[8];
  };
  CanMsgType type;
  size_t dlc;
} CanMessage;
```
So `data` is in a union with four other fields - fields in a union share the same storage, so you can think of them like different "views" into the blob of data they represent. Then to set a uint16_t, you'll have to set `data_u16[0]`.

projects/com_protocol/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+
+static StatusCode prv_rx_callbackA(const CanMessage *msg, void *context, CanAckStatus *ack_reply) {
+  LOG_DEBUG("Received message with id: 0x%x!\n", msg->msg_id);
+  printf("Data:\n\t %lu", msg->data);
+  // this loop gives me numbers I do not understand
+  // for (uint8_t i = 0; i < msg->dlc; i++) {
+  //   uint8_t byte = 0;
+  //   byte = msg->data >> (i * 8);
+  //   printf("0x%x ", byte);
+  // }
```
copperium, line 47: Depending on the endianness of your system this loop looks like it would successfully read the bytes from the message. You might have more success with `data_u16`.

projects/com_protocol/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+  can_register_rx_handler(CAN_ID_B, prv_rx_callbackB, NULL);
+}
+
+static void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  CanMessage *storage = context;
+
+  // log output
+  LOG_DEBUG("Transmit message: 0x%x!\n", storage->msg_id);
+  if (storage->msg_id == CAN_ID_A) {
+    can_transmit(storage, CAN_ACK_STATUS_OK);
```
copperium, line 92: To request an ACK, you'll have to pass a pointer to a `CanAckRequest` struct as defined in `can_ack.h`:
```c
typedef struct CanAckRequest {
  CanAckRequestCb callback;
  void *context;
  uint32_t expected_bitset;
} CanAckRequest;
```
For the callback you can pass something that will log to show the ACK happening. The expected bitset is a bit complex, it's a uint32_t where bit `n` is 1 if you are expecting an ACK from the device with ID `n`. For this homework it's okay to pass `1 << CAN_DEVICE_ID` as the expected bitset so that you only request an ACK from yourself.

projects/com_protocol/src/i2c_spi_test.c
```diff
@@ -0,0 +1,169 @@
...
+#include "i2c.h"
+#include "log.h"
+#include "spi.h"
+
+// Part 1: I2C
+
+// Set this to true to perform an I2C write.
+#define SHOULD_WRITE true
+#define WRITE_I2C_PORT I2C_PORT_1  // or I2C_PORT_2
+#define WRITE_I2C_ADDRESS 0x48
```
copperium, line 12: I suppose it's okay that you copied the I2C and SPI smoke tests for the homework, but do make sure you understand how to use the I2C and SPI libraries and how the I2C and SPI protocols work.

copperium (changes requested):
Nice job with the homework, a couple of pointers about the CAN library.

projects/com_protocol/src/main.c
danielzyy: But isn't that just setting only the first variable in the array? Is it fine to leave the rest of the array as blank/null?

copperium: Yes it's fine to leave the rest of the array as zeros, only the first `dlc` bytes should be TX'd.

copperium:
Going to close this since there's been no activity and you've moved on to actual tasks anyways.

PR #195, opened 2020-10-15
[SOFT-340] Check for invalid timer ID in soft_timer_remaining_time

hewittmcg:
In `soft_timer.c`, `soft_timer_remaining_time` will segfault when called with an invalid timer ID (anything >= SOFT_TIMER_INVALID_TIMER) since it doesn't perform bounds checking on the `timer_id` passed in, unlike in `soft_timer_cancel`.  

Since we tend to set timer IDs to SOFT_TIMER_INVALID_TIMER after they've finished/been cancelled (ex/ in `bts_7200_current_sense.c`), using `soft_timer_remaining_time` to check if these timers have finished could cause issues, and workarounds will likely complicate the FW.

This PR adds a check for this in the x86 and STM32 implementations of `soft_timer_remaining_time`, as well as a quick test to make sure segfaults don't happen anymore.

### DIFF
--- a/libraries/ms-common/inc/soft_timer.h
+++ b/libraries/ms-common/inc/soft_timer.h
@@ -43,6 +43,6 @@ bool soft_timer_cancel(SoftTimerId timer_id);
 bool soft_timer_inuse(void);
 
 // Checks the time left on a particular timer. Returns a 0 if the timer has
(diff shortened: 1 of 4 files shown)


copperium (approved):
Awesome!

PR #198, opened 2020-10-15
MPXE

jessm:


libraries/mpxe-store/src/store.c
```diff
@@ -0,0 +1,174 @@
...
+  // just linear search for it
+  // if key is NULL just get first one with right type
+  for (uint16_t i = 0; i < MAX_STORE_COUNT; i++) {
+    if (key == NULL && s_stores[i].type == type) {
+      return s_stores[i].store;
+    } else if (key != NULL && s_stores[i].type == type) {
+      if (s_stores[i].key == key) {
+        return s_stores[i].store;
+      }
+    }
```
copperium, line 137: Nit: I think this would be more elegant as:
```c
if (s_stores[i].type == type && (key == NULL || s_stores[i].key == key)) {
  return s_stores[i].store;
}
```

libraries/mpxe-store/src/store.c
```diff
@@ -0,0 +1,174 @@
...
+  size_t export_size = mx_store_info__get_packed_size(&msg);
+  uint8_t *export_buf = malloc(export_size);
+  mx_store_info__pack(&msg, export_buf);
+
+  pthread_mutex_lock(&s_sig_lock);
+  // write proto to fifo
+  ssize_t written = write(s_ctop_fifo, export_buf, export_size);
+  // wait for signal that parent got message
+  pthread_mutex_lock(&s_sig_lock);
+  pthread_mutex_unlock(&s_sig_lock);
```
copperium, line 165: using locks to wait to be signalled by the signal handler thread seems kinda hacky, can't think of a better way to do it though. Maybe a condition variable is a more natural way to do it? (https://web.stanford.edu/~ouster/cgi-bin/cs140-spring14/lecture.php?topic=locks.) Also, should there be a timeout mechanism here? Seems like it would block forever if the parent never signals for some reason, though maybe that's desired

libraries/mpxe-store/src/store.c
```diff
@@ -0,0 +1,174 @@
+#include "store.h"
```
copperium, line 1: I feel like MPXE + in particular the communication between python and store.c deserves one of those big writeups on Confluence about how exactly all of it is implemented when you get time

libraries/ms-common/src/can_fsm.c
```diff
@@ -2,6 +2,7 @@
 #include "can.h"
 #include "can_hw.h"
 #include "can_rx.h"
+#include "log.h"
```
copperium, line 5: This was probably accidental?

libraries/ms-common/src/x86/gpio.c
```diff
@@ -1,9 +1,98 @@
 #include "gpio.h"
+#ifndef MPXE
```
copperium, line 2: Instead of duplicating the entirety of x86/gpio.c, do you think it would be better to instead surround the MPXE-specific parts with `#ifdef MPXE`? It might look a little ugly but it might be better than duplicating all of the code.

libraries/ms-common/src/x86/gpio.c
```diff
@@ -1,9 +1,98 @@
...
+#include <stdint.h>
+
+#include "log.h"
+#include "status.h"
+
+static GpioSettings s_pin_settings[GPIO_TOTAL_PINS];
+static uint8_t s_gpio_pin_input_value[GPIO_TOTAL_PINS];
+
+static uint32_t prv_get_index(const GpioAddress *address) {
+  return address->port * (uint32_t)NUM_GPIO_PORTS + address->pin;
```
copperium, line 13: doesn't this need to be fixed to use `GPIO_PINS_PER_PORT`

libraries/ms-common/src/x86/gpio.c
```diff
@@ -12,7 +101,50 @@ static uint32_t prv_get_index(const GpioAddress *address) {
...
+  MxGpioStore *msg = mx_gpio_store__unpack(NULL, msg_buf.len, msg_buf.data);
+  MxGpioStore *mask = mx_gpio_store__unpack(NULL, mask_buf.len, mask_buf.data);
+
+  for (uint16_t i = 0; i < mask->n_state; i++) {
+    // only update state if mask is set
+    if (mask->state[i] != 0) {
+      s_store.state[i] = msg->state[i];
+      if (s_pin_settings[i].state != (uint8_t)msg->state[i]) {
+        s_pin_settings[i].state = msg->state[i];
+        GpioAddress address = { .port = 0, .pin = i % 16 };
```
copperium, line 121: shouldn't this be `.port = i / GPIO_PINS_PER_PORT, .pin = i % GPIO_PINS_PER_PORT`

libraries/ms-common/src/x86/gpio.c
```diff
@@ -12,7 +101,50 @@ static uint32_t prv_get_index(const GpioAddress *address) {
...
+        gpio_it_trigger_interrupt(&address);
```
copperium, line 122: you need to check whether the interrupt is rising or falling here somehow

libraries/ms-common/src/x86/gpio.c
```diff
@@ -12,7 +101,50 @@ static uint32_t prv_get_index(const GpioAddress *address) {
...
+  store_config();
+  StoreFuncs funcs = {
+    (GetPackedSizeFunc)mx_gpio_store__get_packed_size,
+    (PackFunc)mx_gpio_store__pack,
+    (UnpackFunc)mx_gpio_store__unpack,
+    (FreeUnpackedFunc)mx_gpio_store__free_unpacked,
+    (UpdateStoreFunc)update_store,
+  };
+  s_store.n_state = GPIO_TOTAL_PINS;
+  s_store.state = malloc(GPIO_TOTAL_PINS * sizeof(protobuf_c_boolean));
```
copperium, line 142: I feel like we *technically* should have a free function in the `StoreFuncs` that frees this memory since this technically is a memory leak. Then again maybe not, and we should just document it real well that it's meant to be freed on process exit.

libraries/ms-common/src/x86/gpio.c
```diff
@@ -34,6 +167,8 @@ StatusCode gpio_init_pin(const GpioAddress *address, const GpioSettings *setting
   }
 
   s_pin_settings[prv_get_index(address)] = *settings;
+  s_store.state[prv_get_index(address)] = settings->state;
+  prv_export();
```
copperium, line 171: Doesn't `prv_export()` already do the copying of `s_pin_settings` to `s_store`?

libraries/ms-common/src/x86/gpio.c
```diff
@@ -44,6 +179,8 @@ StatusCode gpio_set_state(const GpioAddress *address, GpioState state) {
   }
 
   s_pin_settings[prv_get_index(address)].state = state;
+  s_store.state[prv_get_index(address)] = state;
+  prv_export();
```
copperium, line 183: same, doesn't `prv_export` already do this?

libraries/ms-common/src/x86/gpio.c
```diff
@@ -55,9 +192,12 @@ StatusCode gpio_toggle_state(const GpioAddress *address) {
   uint32_t index = prv_get_index(address);
   if (s_pin_settings[index].state == GPIO_STATE_LOW) {
     s_pin_settings[index].state = GPIO_STATE_HIGH;
+    s_store.state[prv_get_index(address)] = GPIO_STATE_HIGH;
   } else {
     s_pin_settings[index].state = GPIO_STATE_LOW;
+    s_store.state[prv_get_index(address)] = GPIO_STATE_LOW;
   }
+  prv_export();
```
copperium, line 200: same

libraries/ms-common/src/x86/gpio.c
```diff
@@ -71,8 +211,10 @@ StatusCode gpio_get_state(const GpioAddress *address, GpioState *state) {
   // Behave how hardware does when the direction is set to out.
   if (s_pin_settings[index].direction != GPIO_DIR_IN) {
     *state = s_pin_settings[index].state;
+    *state = s_store.state[index];
```
copperium, line 214: the first assignment here is redundant

libraries/ms-common/src/x86/gpio_it.c
```diff
@@ -5,6 +5,7 @@
 
 #include "gpio.h"
 #include "interrupt_def.h"
+#include "log.h"
```
copperium, line 8: Accidental?

libraries/ms-helper/src/mcp2515.c
```diff
@@ -330,3 +333,384 @@ StatusCode mcp2515_tx(Mcp2515Storage *storage, uint32_t id, bool extended, uint6
...
+
+static const uint8_t s_brp_lookup[NUM_MCP2515_BITRATES] = {
+  // BRP calculated for different bitrates
+  [MCP2515_BITRATE_125KBPS] = MCP2515_CAN_BRP_125KBPS,
+  [MCP2515_BITRATE_250KBPS] = MCP2515_CAN_BRP_250KBPS,
+  [MCP2515_BITRATE_500KBPS] = MCP2515_CAN_BRP_500KBPS,
+};
+
+static Mcp2515Storage *s_mpxe_storage;
+static MxMcp2515Store s_store = MX_MCP2515_STORE__INIT;
```
copperium, line 383: Using a static storage + store means that on MPXE there can't be more than one instance of the driver! There should definitely be one store per storage, otherwise we can't use MPXE if there's more than one MCP2515 on any board in the future.

libraries/ms-helper/src/mcp2515.c
```diff
@@ -330,3 +333,384 @@ StatusCode mcp2515_tx(Mcp2515Storage *storage, uint32_t id, bool extended, uint6
 
   return STATUS_CODE_OK;
 }
+#else
```
copperium, line 336: Same comment as gpio - it'd be nice to not have to duplicate all of this code, what if we want to update the x86 version? Then we have to change two spots!

libraries/ms-helper/src/x86/ads1015.c
```diff
@@ -116,3 +118,152 @@ StatusCode ads1015_read_converted(Ads1015Storage *storage, Ads1015Channel channe
...
+#include "ads1015.pb-c.h"
+#include "ads1015_def.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+#define ADS1015_CHANNEL_UPDATE_PERIOD_US ADS1015_CONVERSION_TIME_US_1600_SPS
+
+static MxAds1015Store s_store = MX_ADS1015_STORE__INIT;
```
copperium, line 134: Here too the static store prevents us from using more than one ADS1015 in a project, unless I'm misunderstanding something. Somehow the store should be per-storage.

copperium (changes requested):
Looked at all of the C code, I'll review the Python at some point

libraries/mpxe-store/src/store.c
```diff
@@ -0,0 +1,174 @@
...
+}
+
+Store *prv_get_first_empty() {
+  for (uint16_t i = 0; i < MAX_STORE_COUNT; i++) {
+    if (s_stores[i].key == NULL && s_stores[i].store == NULL) {
+      return &s_stores[i];
+    }
+  }
+  return NULL;
+}
```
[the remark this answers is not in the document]
g-jessmuir: Not exactly since objpool literally acts like a heap and doesn't allow keyed storage other than the pointer to the node, whereas in this case I'm trying to use it like a dict. This is sorta mimicing the soft timer implementation so I'm not that fussed, but you're right there's probably a better way to do this.

libraries/mpxe-store/src/store.c
```diff
@@ -0,0 +1,174 @@
...
+  s_initialized = true;
+}
+
+void store_register(MxStoreType type, StoreFuncs funcs, void *store, void *key) {
+  s_func_table[type] = funcs;
+  // malloc a proto as a store and return a pointer to it
+  Store *local_store = prv_get_first_empty();
+  if (store == NULL) {
+    return;
+  }
```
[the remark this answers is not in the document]
g-jessmuir: in theory an outbound-only communication protocol could have a null store since it would just export a message to python but we don't actually do that so yeah I'll just do that

libraries/mpxe-store/src/store.c
g-jessmuir: Condition variables require a mutex anyways, I think they're more useful for working with more threads (lol what do I know though) so I doubt we'd reduce much complexity. We could make it a timed wait but that also adds some more complexity in setting up the wait so this is probably fine for now.

libraries/mpxe-store/src/store.c
g-jessmuir: when I get time indeed

libraries/ms-common/src/x86/gpio.c
g-jessmuir: We have this unwritten requirement that there's only 16 interrupt slots, so we assign them based on pin number (not port). This is actually a huge issue and should be documented at some point. Added a comment here explaining this.

libraries/ms-common/src/x86/gpio.c
g-jessmuir: Added comment in store.h `store_register()` mentioning they shouldn't be freed

libraries/ms-helper/src/mcp2515.c
g-jessmuir: can we include this as a future ticket since it's nonessential to what we've got working here? also im lazy, i swear ill do it right after this merge

libraries/mpxe-store/src/store.c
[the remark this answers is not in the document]
g-jessmuir: objpool doesn't allow keyed storage of any kind, and doesn't allow iterating. Soft timer implements a similar pattern as this (different cus the softtimerid keys are the array indices), so I'm fairly certain there isn't a significantly easier way to do this

mpxe/integration_tests/int_test.py
```diff
@@ -0,0 +1,24 @@
+import unittest
+import time
+from mpxe.harness import pm
+from mpxe.harness import project
```
copperium, line 4: You don't use the `time` or `project` imports, you might also consider importing `ProjectManager` directly instead of importing `pm`

mpxe/integration_tests/int_test.py
```diff
@@ -0,0 +1,24 @@
...
+    def tearDown(self):
+        self.manager.end()
+        print("======", "PASS", self._testMethodName, "======")
+
+    def assert_can_data(self, name, field, val):
+        msg = self.manager.can.get_latest_by_name(name)
+        self.assertIsNotNone(msg)
+        self.assertEqual(msg.data[field], val)
+
+    def assert_can_got(self, name):
```
copperium, line 20: Nitpick: I don't *really* like this name, maybe `assert_can_received`?

mpxe/integration_tests/test_can_dump.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+import time
+
+from mpxe.integration_tests import int_test
+
+class TestCanDump(int_test.IntTest):
+    def setUp(self):
+        super(TestCanDump, self).setUp()
```
copperium, line 8: In python 3 you can just do `super().setUp()`

copperium, line 8: In python 3 you can just do `super().setUp()`, this goes for the other super() calls too

mpxe/integration_tests/test_can_dump.py
```diff
@@ -0,0 +1,15 @@
...
+        self.can_dump = self.manager.start('can_dump')
+
+    def test_can_dump(self):
+        time.sleep(1)
```
copperium, line 12: This test confuses me, was this just to test that project/pm works? Should we keep this? It is just adding a second to testing time for no reason, perhaps shorten the delay significantly?

mpxe/integration_tests/test_leds.py
```diff
@@ -0,0 +1,16 @@
...
+from mpxe.integration_tests import int_test
+from mpxe.sims.controller_board_blinking_leds import Leds
+
+class TestLeds(int_test.IntTest):
+    def setUp(self):
+        super(TestLeds, self).setUp()
+        self.leds = self.manager.start('controller_board_blinking_leds', Leds())
+
+    def test_leds(self):
+        time.sleep(1)
```
copperium, line 13: Same comment as previously, this doesn't actually seem to be testing anything?

mpxe/harness/project.py
```diff
@@ -0,0 +1,68 @@
...
+        self.popen.stdin.close()
+        self.ctop_fifo.close()
+        ctop_fifo_path = '/tmp/{}_ctop'.format(self.popen.pid)
+        os.unlink(ctop_fifo_path)
+        self.killed = True
+        print('stopped', self.name)
+
+    def write_store(self, store_type, msg, mask):
+        update = stores_pb2.MxStoreUpdate()
+        update.key = 0
```
copperium, line 51: should `write_store` take an optional `key` keyword arg to support writing with keys?

mpxe/sims/sim.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def set_gpio(self, proj, port, pin, state):
+        ind = (ord(port.capitalize()) - ord('A')) * 16 + pin
+        gpio_msg = gpio_pb2.MxGpioStore()
+        gpio_msg.state.extend([0] * 6 * 16)
+        gpio_msg.state[ind] = state
+        gpio_mask = gpio_pb2.MxGpioStore()
+        gpio_mask.state.extend([0] * 6 * 16)
+        gpio_mask.state[ind] = state
+        proj.write_store(stores_pb2.MxStoreType.GPIO, gpio_msg, gpio_mask)
```
copperium, line 19: I think there needs to be some utility abstractions to make accessing each type of store easier, like this but in another module other than `sim` since this class will get too large. We should have utility modules for accessing gpio, adc, etc - one for each major type of store, so that people writing sims / integration tests don't have to deal with the low level store stuff.

For example, there could be an Ads1015 utility class where one instance corresponds to one instance of the ADS1015 driver once we figure out multiple stores; I think the key stuff in project.py is meant for this. Then the PedalBoard sim could just provide access to an Ads1015 utility class. (Maybe even structure each utility class as a Sim so you could have some sort of hierarchy of sims - a PedalBoard sim has an Ads1015 sim, etc.) This might be best for another ticket though.

mpxe/sims/sim.py
```diff
@@ -0,0 +1,19 @@
+from mpxe.protogen import stores_pb2
```
copperium, line 1: You don't use this import

mpxe/sims/pedal_board.py
```diff
@@ -0,0 +1,16 @@
+from mpxe.protogen import stores_pb2
```
copperium, line 1: You don't use this import?

mpxe/sims/mci.py
```diff
@@ -0,0 +1,47 @@
+from mpxe.protogen import stores_pb2
+from mpxe.protogen import mcp2515_pb2
+from mpxe.protogen import gpio_pb2
+
+import time
+
+from mpxe.sims import sim
+
+GPIO_KEY = (stores_pb2.MxStoreType.GPIO, 0)
```
copperium, line 9: In the absence of the utility modules/classes I described a couple comments ago, I think Sim should at least provide a `get_gpio_state` function to avoid having to mess with the store to get gpio states.

mpxe/sims/mci.py
```diff
@@ -0,0 +1,47 @@
...
+MCP2515_KEY = (stores_pb2.MxStoreType.MCP2515, 0)
+
+class Mci(sim.Sim):
+    def __init__(self):
```
copperium, line 13: Please put some blank lines between methods in this class, lol

mpxe/sims/mci.py
```diff
@@ -0,0 +1,47 @@
...
+            store_gpio_ab()
+        elif proj.stores[GPIO_KEY].state[9] == 1 and self.gpio_ab[9] != 1:
+            time.sleep(0.1)
+            store_gpio_ab()
+            # update precharge monitor and latch out
+            self.set_gpio(proj, 'b', 0, True)
+            time.sleep(0.1)
+            self.set_gpio(proj, 'a', 10, True)
+        else:
+            store_gpio_ab()
```
copperium, line 31: The gpio interfacing stuff here is kinda ugly, there should probably be a few more layers of abstraction; I don't really have any idea what this is doing without diving into the mci schematic

mpxe/sims/controller_board_blinking_leds.py
```diff
@@ -0,0 +1,13 @@
+from mpxe.protogen import stores_pb2
+
+from mpxe.sims import sim
+
+class Leds(sim.Sim):
+    def handle_update(self, pm, proj):
+        states = [int(i) for i in proj.stores[(stores_pb2.MxStoreType.GPIO, 0)].state]
+        # use addution to indicate port - 16 = GPIO_PORT_B, 0 = GPIO_PORT_A
+        leds = [states[16+5], states[16+4], states[16+3], states[0+15]]
+        print('LEDs:', leds)
```
copperium, line 10: definite use case for the gpio utility stuff on Sim / another module

copperium (changes requested):
Sorry for the incredibly large number of comments, lol. It turns out pylint wasn't running on mpxe, I'll push a commit in a second to fix that. It'll throw a ton of errors, mostly missing docstrings and formatting stuff - they will all have to be resolved or suppressed, but most of them are pretty simple.

libraries/ms-common/src/stm32f0xx/gpio_it.c
```diff
@@ -34,6 +35,14 @@ static uint8_t prv_get_irq_channel(uint8_t pin) {
   return 7;
 }
 
+StatusCode gpio_it_get_edge(const GpioAddress *address, InterruptEdge *edge) {
+  if (s_gpio_it_interrupts[address->pin].callback != NULL) {
```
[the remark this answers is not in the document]
g-jessmuir: This is a check for most other functions in gpio_it.c, and I'm inclined to say it's not particularly non-obvious? adding bool registered would mean changing two other checks and also adding some cleanup code in case of bad interrupt register, so I think this is fine but if you still want it I'll add it

mpxe/harness/canio.py
```diff
@@ -0,0 +1,50 @@
...
+class Msg:
+    def __init__(self, metadata, data):
+        self.metadata = metadata
+        self.data = data
+
+class Canio:
+    def __init__(self, max_msgs=10):
+        self.messages = deque(maxlen=max_msgs)
+        self.db = cantools.database.load_file('../system_can.dbc')
+        self.bus = can.interface.Bus('vcan0', receive_own_messages=True, bustype='socketcan')
```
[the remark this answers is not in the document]
g-jessmuir: I feel like there's other stuff to go into running mpxe + hardware so let's leave this for another ticket for now

mpxe/integration_tests/test_leds.py
g-jessmuir: I think it's worth keeping if only as a simple demo to explain what MPXE is doing, very easy to relate it to what's going on in hardware. Since integration tests are going to be fairly lengthy anyways I think an extra second for this is fine

mpxe/sims/sim.py
g-jessmuir: used for storetype in write_store

mpxe/sims/pedal_board.py
g-jessmuir: used for storetype in write_store

mpxe/harness/canio.py
```diff
@@ -0,0 +1,50 @@
+import can
+import cantools
```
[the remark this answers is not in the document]
g-jessmuir: I think fully integrating mpxe into infra is a separate hefty ticket

mpxe/harness/pm.py
```diff
@@ -0,0 +1,82 @@
...
+        self.proj_name_list = os.listdir('/home/vagrant/shared/firmware_xiv/projects')
+        self.killed = False
+        # run listener threads
+        self.poll_thread = threading.Thread(target=self.poll)
+        self.poll_thread.start()
+        self.can = canio.Canio()
+        
+    def start(self, name, sim=None):
+        if name not in self.proj_name_list:
+            raise Exception('invalid project')
```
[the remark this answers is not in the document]
g-jessmuir: not a fan of printing all projects since there's a lot, changed to a slightly more generic message. I hate it when filter.mk just spews

mpxe/harness/pm.py
```diff
@@ -0,0 +1,82 @@
...
+from mpxe.harness import project
+from mpxe.harness import canio
+from mpxe.sims.sim import Sim
+
+class ProjectManager:
+    def __init__(self):
+        # index projects by stdout and ctop_fifo fd
+        self.proj_fds = {}
+        self.statuses = {}
+        self.proj_name_list = os.listdir('/home/vagrant/shared/firmware_xiv/projects')
```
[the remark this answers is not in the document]
g-jessmuir: stored root as a constant in pm.py but it makes imports wonky, would appreciate suggestions since I'm assuming it's best to use that constant then build off it in other modules

mpxe/harness/pm.py
```diff
@@ -0,0 +1,82 @@
...
+    def poll(self):
+        def prep_poll():
+            poll = select.poll()
+            for fd in self.proj_fds.keys():
+                poll.register(fd, select.POLLIN)
+            return poll
+
+        def handle_poll_res(res):
+            if res[1] != select.POLLIN:
+                raise Exception('done')
```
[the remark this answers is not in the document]
g-jessmuir: not super sure on how StopIteration worked so I added an error type, is it ok for it to just be `pass`?

mpxe/harness/pm.py
```diff
@@ -0,0 +1,82 @@
...
+            # res is tuple (fd, event)
+            proj = self.proj_fds[res[0]]
+            # Check if we should check stdout or ctop
+            if res[0] == proj.popen.stdout.fileno():
```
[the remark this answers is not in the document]
g-jessmuir: maybe? but then you have to register both dict keys for polling. I think this is fine for now, I was feeling lazy so I left it as is

mpxe/harness/project.py
```diff
@@ -0,0 +1,68 @@
...
+        # Subprocess is expected to create child to parent fifo, open in raw bytes mode
+        self.ctop_fifo = open(ctop_fifo_path, 'rb')
+        if self.ctop_fifo is None:
+            raise Exception('failed to open ctop fifo')
+        ctop_fd = self.ctop_fifo.fileno()
+        ctop_fl = fcntl.fcntl(ctop_fd, fcntl.F_GETFL)
+        fcntl.fcntl(ctop_fd, fcntl.F_SETFL, ctop_fl | os.O_NONBLOCK)
+
+        self.sim = sim
+        print('started', self.name)
```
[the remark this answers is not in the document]
g-jessmuir: removed

mpxe/harness/pm.py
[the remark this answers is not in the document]
copperium: yeah that's alright

mpxe/integration_tests/deps.mk
```diff
@@ -0,0 +1,9 @@
+# controller_board_blinking_leds must be first in this list
```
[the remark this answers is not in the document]
g-jessmuir: lol we saw this together when make was being spooky, I don't remember what we did to fix it but yeah the comment was outtdated

mpxe/integration_tests/deps.mk
```diff
@@ -0,0 +1,9 @@
...
+# add new projects to the bottom
+
+# MPXE_LIBS += mpxe-gen
```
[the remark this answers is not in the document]
g-jessmuir: removed

libraries/libcore/inc/log.h
```diff
@@ -23,9 +23,31 @@ typedef enum {
...
     if ((level) >= LOG_LEVEL_VERBOSITY) {                                     \
       printf("[%u] %s:%u: " fmt, (level), __FILE__, __LINE__, ##__VA_ARGS__); \
     }                                                                         \
   } while (0)
+#else
+#include <pthread.h>
+#include <stdbool.h>
+
+extern pthread_mutex_t log_lock;
+extern bool store_lib_inited;
```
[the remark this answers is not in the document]
g-jessmuir: Addressed by SOFT-308 ticket by Anthony Chang

copperium:
@jessm Could you do a final sign off before we merge?

jessm:
LGTM! kinda wild to see it passing CI, thanks for that

copperium (approved):

PR #199, opened 2020-10-16
[SOFT-323] Refactor BTS7040, add fault handling

hewittmcg:
Changes made here more or less refactor the BTS7040 driver following the format of the BTS7200 driver, adding fault handling and output enable/disable abstractions.

I created `bts7xxx_common.h` and `bts7xxx_common.c` to hold functions/structs that are common to both the drivers for simplicity (I think we briefly discussed this about a week or so ago).  There's probably a way to factor out some of the initialization functions into this as well, but I don't think that's worth the added complexity.

The names of the drivers have also been changed from `bts_7200_current_sense` and `bts_7040_7008_current_sense` to `bts_7200_load_switch` and `bts_7040_load_switch` respectively.  I don't think we use BTS7008s anywhere at the moment, and from what I can tell from taking a quick look at its datasheet, it's a lot closer to the BTS7200.


projects/power_distribution/src/current_measurement.c
```diff
@@ -75,23 +76,34 @@ StatusCode power_distribution_current_measurement_init(PowerDistributionCurrentS
     bts_7200_settings.select_pin = &s_hw_config.bts7200s[i].dsel_pin;
     bts_7200_settings.enable_0_pin = &s_hw_config.bts7200s[i].en0_pin;
     bts_7200_settings.enable_1_pin = &s_hw_config.bts7200s[i].en1_pin;
+
+    // Add reistor, min/max fault voltages
```
copperium, line 80: Typo: `reistor`

libraries/ms-drivers/inc/bts7xxx_common.h
```diff
@@ -0,0 +1,54 @@
...
+
+// Holds pin-specific info for SEL pins
+typedef struct {
+  GpioAddress *select_pin_stm32;
+  Pca9539rGpioAddress *select_pin_pca9539r;
+  Bts7xxxPinType pin_type;
+} Bts7xxxSelectPin;
+
+// Broad function to enable the pin passed in.
+StatusCode bts7xxx_enable_pin(Bts7xxxEnablePin *pin);
```
copperium, line 39: I think it would be an interesting idea to roll the `fault_in_progress` handling into this function, it looks doable. Then in power distro (and other future projects with a mix of bts7200 and bts7040), to enable a pin, we can just store its `Bts7xxxEnablePin *` and call `bts7xxx_enable_pin` with it instead of worrying about whether a given pin is on a bts7200 or bts7040.

I get the feeling that this module was meant to be private to the bts7200 and bts7040 drivers, though my suggestion would make this function part of the public API of the drivers. You might consider then splitting this module into a `bts7xxx_common` which is public and `bts7xxx_common_impl` which holds private code-sharing functions. I do think `bts7xxx_enable_pin`, `bts7xxx_disable_pin`, and `bts7xxx_get_pin_enabled` at least should be public. Any functions that are part of the public API should be unit tested (functions that are private to bts7200/7040 will/should be tested by those modules' tests and so should be fine without unit tests of their own).

libraries/ms-drivers/src/bts_7040_load_switch.c
```diff
@@ -0,0 +1,177 @@
...
+  storage->max_fault_voltage_mv = settings->max_fault_voltage_mv;
+
+  // initialize the enable pin
+  GpioSettings enable_settings = {
+    .direction = GPIO_DIR_OUT,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_NONE,
+    .alt_function = GPIO_ALTFN_NONE,
+  };
+
```
copperium, line 65: I think a lot of this copy/pasting could instead be reused between the bts7200 and bts7040 drivers. Perhaps a `Bts7xxxStorage` struct is in order which contains all the common one-off fields (`resistor`, `min_fault_voltage_mv`, `enable_pin`, etc) between the two drivers? Then you could include a `Bts7xxxStorage` struct in `Bts7200Storage` and `Bts7040Storage` and factor out most of the copy/pasting here and elsewhere into non-public bts7xxx functions.

The only things that *really* have to be different between the two modules are the measurement reading and fault detection logic, I think that the rest of it could be factored out into `bts7xxx` by taking `Bts7xxxStorage *` pointers.

copperium (changes requested):
Awesome job Hewitt! Most of the comments are nitpicks and typos with a couple structural changes or opportunities for factoring out code.

While you're in the business of renaming things, if it's not too much, could you rename `bts_7200_load_switch` to `bts7200_load_switch` and change the function names and uses, and the same for `bts_7040_load_switch` to `bts7040_load_switch`? I must have written the blasted underscore a long time ago and it's inconsistent with the rest of the drivers. Thanks!

PR #201, opened 2020-10-21
[SOFT-347] Fix ADC returning old values on STM32

copperium:
The STM32 version of `adc.c` would return old values from `adc_read_raw` (probably/especially when called form within a soft timer). This is because to block until the ADC reading was ready, we would wait for the EOSEQ (end-of-sequence) bit to change from 1 to 0. However, the EOSEQ bit is always 0 until the end of the sequence, at which point it changes to 1; however, the ADC IRQHandler (interrupt handler) is called whenever EOSEQ changes to 1 and resets it to 0 before the blocking while loop can see the EOSEQ value. Thus, we would just never block, and adding a `!` wouldn't fix it.

The fix is to track whether we're currently converting on our own with a `volatile bool` and wait on that instead.

Also added a notice about not using it from a high-priority interrupt since that'll cause deadlock (caused by the NVIC queuing the ADC interrupt, which is also high priority, to after the calling interrupt finishes), and clarified the unit of the readings from `adc_read_converted`/`adc_read_converted_pin`.

### DIFF
--- a/libraries/ms-common/inc/adc.h
+++ b/libraries/ms-common/inc/adc.h
@@ -55,10 +55,13 @@ StatusCode adc_set_channel_pin(GpioAddress address, bool new_state);
 // completes a conversion
 StatusCode adc_register_callback_pin(GpioAddress address, AdcPinCallback callback, void *context);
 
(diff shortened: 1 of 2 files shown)


g-jessmuir (approved):
LGTM, thanks for explanation comment

PR #202, opened 2020-10-22
[SOFT-282] Calibrate solar scaling factors

copperium:
These calibrated values were derived from smoke test results. Other notes:
* It turns out the ACS722 current sense IC adds a bias to the output voltage, so I added a field to unbias it.
* The SPV1020 MPPT current scaling factor is based on exactly one data point, so it needs to be updated with a better value after more calibration in the future.


g-jessmuir (approved):
LGTM

PR #203, opened 2020-10-22
[SOFT-293] Determine the solar board based on detection pin

copperium:
Rev 2 of solar sense gives us a pin that's high on the 6 MPPT board and low on the 5 MPPT board. This implements checking it.


g-jessmuir (approved):
LGTM

PR #204, opened 2020-10-23
A very late submission of the fw 102 hmk

Fionon:
Sorry, I know I'm super behind

- Fion Lin

### DIFF
--- a/projects/firmware_102_hmk/src/main.c
+++ b/projects/firmware_102_hmk/src/main.c
@@ -1,0 +1,49 @@
+#include "interrupt.h"   // interrupts are required for soft timers
+#include "log.h"         // for printing
+#include "soft_timer.h"  // for soft timers
+#include "wait.h"        // for wait function
+
+#define COUNTER_MS_A 500   // milliseconds between coin flips
+#define COUNTER_MS_B 1000  // milliseconds between coin flips
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void prv_blink_timeout_a(SoftTimerId timer_id, void *context) {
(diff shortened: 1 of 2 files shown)


projects/firmware_102_hmk/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+int main(void) {
+  LOG_DEBUG("Hello World!\n");
+
+  interrupt_init();
+  soft_timer_init();
+
+  Counters storage = { 0 };  // we use this to initialize a struct to be all 0
+
+  soft_timer_start_millis(COUNTER_MS_A, prv_blink_timeout_a, &storage, NULL);
+  soft_timer_start_millis(COUNTER_MS_B, prv_blink_timeout_b, &storage, NULL);
```
copperium, line 41: The homework does say:
> Uses a **single soft timer** to increment counter_a every half a second, and counter_b every second.

So refactor to use one soft timer, then you should be good

projects/firmware_102_hmk/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  // Infinite loop
+  while (true) {
+    // Wait for interrupts
+    wait();
+  }
+  return 0;
+}
+\n
```
copperium, line 49: Don't think you meant for this \n to be here?

copperium (changes requested):
Nice job! Usually instead of forking the repo, we just clone it and make branches - see [our git workflow](https://uwmidsun.atlassian.net/l/c/TeCYpJjC). It's fine for this PR, but for FW 103 and the future make sure to use our workflow.

Just one thing to address before moving on to FW 103.

PR #205, opened 2020-10-27
Soft 306 adc_read: python implementation

selinahsu:
python side of the adc_read function, in the babydriver project


projects/baby_driver/scripts/gpio_port.py
```diff
@@ -0,0 +1,15 @@
+"""Constants for identifying GPIO ports."""
+
+import enum
+
+class GpioPort(enum.Enum):
```
copperium, line 5: Ah, it seems like some of the tests are failing because inheriting from `Enum` makes e.g. `GpioPort.A` not equal to 0, instead it's some weird enum object. I think you should just leave this as a normal class and not have it inherit from Enum.

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
...
+    if port < 0 or port >= GpioPort.NUM_GPIO_PORTS:
+        raise ValueError("ERROR: invalid GPIO port")
+    if pin < 0 or pin >= NUM_PINS_PER_PORT:
+        raise ValueError("ERROR: invalid GPIO pin number")
+    if raw not in (True, False):
+        raise ValueError("ERROR: raw must be a bool value")
+
+    # Send CAN message
+    data = [
+        (BabydriverMessageId.ADC_READ_COMMAND, 1),
```
copperium, line 31: If you specify the `babydriver_id` keyword parameter like `babydriver_id=BabydriverMessageId.ADC_READ_COMMAND` in the `send_message` call, you won't have to specify this here; the babydriver ID is added onto the beginning of the packed data for you.

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+    data = can_util.can_pack(data)
+    can_util.send_message(data)
+
+    # Wait to receive the first CAN message with data
+    data_mssg = can_util.next_message()
+
+    # Check if received message is an adc_read
+    if data_mssg.data[0] != BabydriverMessageId.ADC_READ_DATA:
+        raise Exception("ERROR: did not receive adc read data")
```
copperium, line 44: Please use the `babydriver_id` keyword argument on `next_message` to wait specifically for an `ADC_READ_DATA` message. That'll let you remove this check, `next_message` will raise a `ValueError` if the message it received wasn't a what was specified in `babydriver_id`.

Also, is `mssg` a typo? Shouldn't it just be `msg`?

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Extract the low and high bytes of the ADC conversion
+    result_low = data_mssg.data[1]
+    result_high = data_mssg.data[2]
+
+    # Wait to receive the second CAN message with status
+    status_mssg = can_util.next_message()
```
copperium, line 51: Here too please use the `babydriver_id` keyword argument to wait specifically for a `STATUS` message. Also: `mssg`?

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Check if status is not ok (if it's nonzero)
+    if status_mssg.data[1] != BabydriverMessageId.STATUS:
+        raise Exception("ERROR: received a nonzero STATUS_CODE: {}".format(status_mssg.data[1]))
```
copperium, line 55: While `BabydriverMessageId.STATUS` is technically equal to 0, it's meant for the babydriver ID, not the OK status - checking if the status is nonzero should be done with an explicit comparison to 0. Another nit: I'd prefer you assign `status_mssg.data[1]` to a descriptive variable rather than referencing it twice.

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
+"""Python implementation of the adc_read function for Babydriver."""
+
+import can_util
+from gpio_port import GpioPort
+from message_defs import BabydriverMessageId
+
+NUM_PINS_PER_PORT = 16
```
copperium, line 7: This constant should definitely go in gpio_port.py since it's generic to gpio ports in general.

projects/baby_driver/scripts/gpio_port.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    An enumeration to specify GPIO ports by name (e.g. GpioPort.A instead of 0).
+    """
+    A = 0
+    B = 1
+    C = 2
+    D = 3
+    E = 4
+    F = 5
+    NUM_GPIO_PORTS = 6
```
copperium, line 15: I know it's in the enum in C, but `NUM_GPIO_PORTS` should really be a top level constant on its own, not a member of the enum.

projects/baby_driver/scripts/test_adc_read.py
```diff
@@ -0,0 +1,49 @@
...
+
+RAW_MAX = 4095
+CONVERTED_MAX = 3000
+
+class TestADCRead(unittest.TestCase):
+    """Test Babydriver's adc_read function"""
+
+    @patch('can_util.send_message')
+    @patch('can_util.next_message')
+    def test_check_range_pin(self):
```
copperium, line 17: Ah, it turns out I lied: this test is failing because `unittest` is trying to pass it `mock_next_message` and `mock_send_message` as parameters, but the function doesn't take them. You'll have to add them to the parameter list.

projects/baby_driver/scripts/test_adc_read.py
```diff
@@ -0,0 +1,49 @@
...
+        """Tests Maximum values"""
+        self.assertLessEqual(adc_read(GpioPort.A, 0, True), RAW_MAX)
+        self.assertLessEqual(adc_read(GpioPort.A, 1, True), RAW_MAX)
+        self.assertLessEqual(adc_read(GpioPort.A, 2, True), RAW_MAX)
+        self.assertLessEqual(adc_read(GpioPort.A, 0, False), CONVERTED_MAX)
+        self.assertLessEqual(adc_read(GpioPort.A, 1, False), CONVERTED_MAX)
+        self.assertLessEqual(adc_read(GpioPort.A, 2, False), CONVERTED_MAX)
```
copperium, line 24: This test seems to be testing properties of the ADC implementation, rather than the Python interface. I think a better test of the Python would be to use the mocks to set the return value of `next_message` to a Message that matches CAN messages the C should provide, then check that `adc_read` correctly parses and returns the values from it.

You should also have another test that uses the `side_effect` of `mock_send_message` to intercept the data being sent and ensures that it's in the correct format.

projects/baby_driver/scripts/test_adc_read.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @patch('can_util.send_message')
+    @patch('can_util.next_message')
+    def test_fail_conditions(self, mock_send_message, mock_next_message):
```
copperium, line 28: It turns out, per [the documentation](https://docs.python.org/3/library/unittest.mock-examples.html#patch-decorators), that the `patch` decorator actually passes the mock arguments in the reverse order that the decorators are laid out, so this argument list should be `self, mock_next_message, mock_send_message` to be correct.

projects/baby_driver/scripts/test_adc_read.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertRaises(ValueError, adc_read, -1, 1, 0)
+        # Tests fail conditions for pin parameter
+        self.assertRaises(ValueError, adc_read, GpioPort.A, NUM_PINS_PER_PORT, 0)
+        self.assertRaises(ValueError, adc_read, GpioPort.A, -1, 0)
+        # Tests fail conditions for raw parameter
+        self.assertRaises(ValueError, adc_read, GpioPort.A, 1, -1)
+
+        # Tests fail conditions for can_util.send_message, in index 0
+        mock_send_message.side_effect = Message([-1])
+        self.assertRaises(Exception, adc_read, GpioPort.A, 0, 0)
```
copperium, line 42: I'm not sure what the fail condition you're going for here is? `send_message` doesn't return anything we use, and I'm not sure what setting `side_effect` to a Message instance is supposed to do.

projects/baby_driver/scripts/test_adc_read.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # Tests fail condition for can_util.next_message, in index 1
+        mock_next_message.side_effect = Message([0, -1])
+        self.assertRaises(Exception, adc_read, GpioPort.A, 0, 0)
```
copperium, line 46: First: did you mean `return_value` instead of `side_effect`?

As well, this isn't the correct syntax for initializing `Message`. It takes a `message_id`, a `device_id` (you can probably ignore these), and a `**kwargs` parameter which its docstring says feeds directly into `can.Message` as documented here: https://python-can.readthedocs.io/en/master/message.html. So you'll need to pass a `data` keyword argument containing the list of bytes you want to be in the `data` property on the Message; you can use `can_pack` here to build the list of bytes.

If you want more complex return value behaviour, the unittest.mock documentation says you can set `side_effect` to be a function that is called with the same arguments as `next_message` is in the code, and whatever it returns will be `next_message`'s return value.

projects/baby_driver/scripts/adc_read.py
```diff
@@ -0,0 +1,58 @@
...
+
+def adc_read(port, pin, raw):
```
copperium, line 9: One thing you forgot: you need to `from adc_read import adc_read` and `from gpio_port import GpioPort` in the top level of `repl_setup.py` so that the REPL has access to the function without importing it.

copperium (changes requested):
Nice job! To run your tests, you can just run `python3 projects/baby_driver/scripts/test_adc_read.py` (or in general just run the unit test file). Also, if you want to try out the function in the babydriver REPL, or at least test that you've imported things correctly in `repl_setup.py`, you can run `make babydriver PLATFORM=x86`; this should start up the C babydriver project and drop you into the REPL where you should be able to call `adc_read` if you've imported everything correctly.

projects/baby_driver/scripts/test_adc_read.py
selinahsu: Ah that's a typo, I meant for `side_effect` to be `return_value`

copperium (approved):

PR #206, opened 2020-10-27
[SOFT-350] Improve the mux driver

copperium:
This just moves ownership of real muxes' enable and output pins from the `mux` driver to the individual project. `mux` is a generic driver whose only function is to write a binary number into GPIO pins. In solar, it was used for a demux which didn't even have enable or output pins, so they had no business being in the driver's control.

This also solves the part of the BTS7200 smoke test issue where the mux enable pin was wrongly high instead of low - power_distribution's `current_measurement` takes care of that now, as does `smoke_bts7200`.

I also took the liberty to incorporate a few of the changes from soft_151_power_distro_validation, so smoke_bts7200 should have half a chance of working now.


g-jessmuir (approved):
Makes sense, good change

PR #208, opened 2020-10-28
gpio_set function for Babydriver project

CheranMahalingam:
Created gpio_set function and adjusted associated scripts.


projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,32 @@
...
+    Sets a GPIO pin
+    Args:
+        port: port of the GPIO pin to set
+        pin: pin number of the GPIO pin to set
+        state: 0 if setting low, 1 if setting high
+    Returns:
+        A boolean value. If true, the CAN message was sent successfully, otherwise
+        a ValueError or TimeoutError was raised.
+    """
+    if 0 <= port <= 6 and 0 <= pin <= 15 and 0 <= state <= 1:
```
copperium, line 16: You should use `GpioPort.NUM_GPIO_PORTS` here instead of 6. Also port = 6 is invalid, so it should be `0 <= port < 6 and 0 <= pin < 16`. You should probably have a `NUM_PINS_PER_PORT` constant somewhere and use it instead of 16.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,32 @@
...
+def gpio_set(port, pin, state):
+    """
+    Sets a GPIO pin
+    Args:
+        port: port of the GPIO pin to set
+        pin: pin number of the GPIO pin to set
+        state: 0 if setting low, 1 if setting high
+    Returns:
+        A boolean value. If true, the CAN message was sent successfully, otherwise
+        a ValueError or TimeoutError was raised.
```
copperium, line 14: I'd prefer this not return anything and raise an exception if something goes wrong - that means you don't need the try/except around the function calls.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,32 @@
+"""This module provides the Python implementation to set GPIO pins using the CAN protocol"""
+
+import can_util
+
+def gpio_set(port, pin, state):
+    """
+    Sets a GPIO pin
+    Args:
+        port: port of the GPIO pin to set
```
copperium, line 9: Request: could you make it so that you can pass in a string like `'a'` or `'A'` as the port and have that be treated like the appropriate port number? Look at Selina's adc_read PR for how she did it.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,32 @@
...
+        A boolean value. If true, the CAN message was sent successfully, otherwise
+        a ValueError or TimeoutError was raised.
+    """
+    if 0 <= port <= 6 and 0 <= pin <= 15 and 0 <= state <= 1:
+        can_pack_args = [(port, 1), (pin, 1), (state, 1)]
+    else:
+        return False
+
+    try:
+        can_util.send_message(babydriver_id=1, data=can_util.can_pack(can_pack_args))
```
copperium, line 22: Please use `BabydriverMessageId.GPIO_SET` instead of 1 here. You can import it with `from message_defs import BabydriverMessageId`.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,32 @@
...
+    except ValueError:
+        return False
+
+    try:
+        can_util.next_message(babydriver_id=1)
```
copperium, line 27: You should be waiting for a status message, that's id 0. So use `BabydriverMessageId.STATUS` instead of 1 here. Also, `next_message` returns a Message object (see can_util.py), so please capture it in a variable and check its status (i.e. its `data[1]` attribute - see the confluence page linked in the ticket for a description of status messages). If the status returned by the status message is nonzero, that means something went wrong on the C side, so raise an exception specifying the nonzero status.

projects/baby_driver/scripts/repl_setup.py
```diff
@@ -9,8 +9,11 @@ def setup_default_channel():
 
     # We import these here so that they aren't imported for the REPL
     # pylint: disable=import-outside-toplevel
+    # pylint: disable=unused-import
     import argparse
     import can_util
+    import gpio_port
+    import gpio_set
```
copperium, line 16: Ah, when I wrote import them in repl_setup.py, I meant import them on the top level so that the REPL has access to them. If you type `make babydriver`, you should be able to use `gpio_set`. Also, could you instead do `from gpio_set import gpio_set` and `from gpio_port import GpioPort` so that you can do `gpio_set(...)` from the REPL instead of `gpio_set.gpio_set(...)`?

projects/baby_driver/scripts/test_gpio_set.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TestGpioSet(unittest.TestCase):
+    """Test gpio_set function"""
+
+    @patch('can_util.send_message')
+    @patch('can_util.next_message')
+    def test_low_bound(self, mock_send_message, mock_next_message):
+        """Tests minimum values for pin and port"""
+        mock_send_message.side_effect = None
+        mock_next_message.side_effect = None
```
copperium, line 16: `side_effect` is `None` by default, you don't have to set it explicitly; this goes for the other places you do that too

copperium (changes requested):
Awesome job with the module and the tests Cheran! Just a couple things, some will require changing the tests a bit but there's [an `assertRaises` function in unittest] (https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) so it shouldn't be that bad to change.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,46 @@
...
+        Exception: if a non-zero status code was received
+    """
+
+    # If given port as a string, it is converted to a string
+    if isinstance(port, str):
+        port = getattr(GpioPort, port.capitalize())
+
+    # Checks whether valid parameters were passed into gpio_set
+    if port < 0 or port >= GpioPort.NUM_GPIO_PORTS:
+        raise ValueError("Expected port between A and {}".format(chr(GpioPort.NUM_GPIO_PORTS + 64)))
```
copperium, line 28: Nit: the magic number 64 is a bit cryptic, I'd prefer `ord('A') - 1`.

projects/baby_driver/scripts/gpio_set.py
```diff
@@ -0,0 +1,46 @@
...
+    if pin < 0 or pin >= NUM_PINS_PER_PORT:
+        raise ValueError("Expected pin between 0 and {}".format(NUM_PINS_PER_PORT - 1))
+    # pylint: disable=consider-using-in
+    if state != 0 and state != 1:
+        raise ValueError("Expected state of 0 (low) or 1 (high)")
+
```
copperium, line 34: I think pylint's guidance is good here, `state not in (0, 1)` is cleaner.

copperium (changes requested):
This is amazing. Two nitpicks, then we can get this merged!

copperium (approved):

PR #209, opened 2020-10-29
Aryan Gosalia Module 102 code

aryan9154:


projects/102_Test/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+
+  counter->counter_a++;
+  LOG_DEBUG("Counter A: %u\n", (counter->counter_a));
+
+  if ((counter->counter_a) % 2 == 0) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %u\n", (counter->counter_b));
+  }
+
+  soft_timer_start_seconds(0.5, prv_counter_timeout, counter, NULL);
```
copperium, line 24: I'm kind of surprised this works, a more conventional way to do it would be `soft_timer_start_millis(500, prv_counter_timeout, counter, NULL)` but this works too. In any case, please factor out the 0.5 constant that you use twice into a `#define`'d constant with an appropriate name that specifies the unit (seconds) - this makes it easier to read and change.

projects/102_Test/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  struct Counters counter;
```
copperium, line 31: Since we have the `typedef` in the definition of `Counters`, we don't need to say `struct Counters`, we can just do `Counters counter;`. You should also zero-initialize it with `= { 0 }`.

projects/102_Test/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+
+  LOG_DEBUG("%u %u", counter.counter_a, counter.counter_b);
```
copperium, line 33: I don't think this line is necessary?

projects/102_Test/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+static void prv_counter_timeout(SoftTimerId timer_id, void *context) {
+  Counters *counter = context;
+
+  counter->counter_a++;
+  LOG_DEBUG("Counter A: %u\n", (counter->counter_a));
+
+  if ((counter->counter_a) % 2 == 0) {
+    counter->counter_b++;
+    LOG_DEBUG("Counter B: %u\n", (counter->counter_b));
+  }
```
copperium, line 22: Nitpicks: you don't need the brackets around `(counter->counter_a)` and `(counter->counter_b)`.

copperium (changes requested):
Nice job! You seem to have committed your `debug_test` and `hello_world` projects too, please remove them; you can do that by deleting them from the projects directory. Other than that, just a few nitpicks, then you can move on to FW 103.

projects/debug_test/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+
+  prv_init_leds(&s_storage);
+  uint32_t buffer[20];
+  LOG_DEBUG("%ld LEDs set up\n", s_storage.num_leds);
```
copperium, line 58: FYI, `size_t` has a different size depending on whether you're compiling for STM32 or x86, which is why Travis is failing on this line; the correct format specifier to use for `size_t` is `%zu`.

copperium (commented):
FW 102 code LGTM, I see you didn't include the debug_test and hello_world code in the FW 103 submission so we'll call it good. Left one comment on the debug_test code as to why Travis CI is failing just as an FYI, you don't have to fix it. Closing this since homework is complete.

copperium (commented):
FW 102 code LGTM, I see you didn't include the debug_test and hello_world code in the FW 103 submission so we'll call it good. Left one comment on the debug_test code as to why Travis CI is failing just as an FYI, you don't have to fix it. Closing this since homework is complete.

PR #211, opened 2020-10-29
Soft 158 add mpxe adt7476a

JarvisWeng:


mpxe/integration_tests/test_adt7476a.py
```diff
@@ -0,0 +1,42 @@
...
+from mpxe.integration_tests import int_test
+from mpxe.sims.adt7476a import Adt7476a
+
+class TestAdt7476a(int_test.IntTest):
+    def setUp(self):
+        super(TestAdt7476a, self).setUp()
+        self.adt7476a = self.manager.start('smoke_adt7476a', Adt7476a())
+
+    def test_adt7476a(self):
+        time.sleep(1)
```
g-jessmuir, line 13: this seems excessive enough to just loop...

mpxe/sims/adt7476a.py
```diff
@@ -0,0 +1,25 @@
...
+        # for some reason there's a GPIO store initializes before
+        # a ADT7476a store is initialized
+        # that's why I need to if to check first
+        # print(stores)
+        if (stores_pb2.MxStoreType.ADT7476A, 0) in stores:
+            adt         = stores[(stores_pb2.MxStoreType.ADT7476A, 0)]
+            self.speed  = [int(adt.speed[i]) for i in range(len(adt.speed)) if i < 3]
+            self.status = [adt.status[i] for i in range(len(adt.status)) if i < 3]
+
+    def handle_log(self, pm, proj, log):
```
g-jessmuir, line 18: you can leave this function out since it's implemented by the subclass so it gets called automatically

mpxe/sims/adt7476a.py
```diff
@@ -0,0 +1,25 @@
...
+        print('[{}]'.format(proj.name), log)
+
+    def assert_store_value(self, proj, speed, status, channel):
```
g-jessmuir, line 21: Nit: should be `assert_store_values` since you're asserting multiple values

mpxe/sims/adt7476a.py
```diff
@@ -0,0 +1,25 @@
+from mpxe.protogen import stores_pb2
+from mpxe.protogen import adt7476a_pb2
+
+from mpxe.sims import sim
+
+class Adt7476a(sim.Sim):
+    def handle_update(self, pm, proj):
+        stores  = proj.stores
+        # for some reason there's a GPIO store initializes before
```
g-jessmuir, line 9: this is expected behaviour so probably don't need a comment for it, also should remove the comment `# print(stores)`. Ideally don't use first person in comments since this code will probably be with the team longer than you'll be

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -101,3 +102,174 @@ StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
 
   return STATUS_CODE_OK;
 }
+
```
g-jessmuir, line 105: so we recently changed this structure in MPXE, we're now gonna try to just use #ifdefs within the original code since that should make it cleaner and more obvious what's going on in MPXE vs not in MPXE, so rather than having the whole file in a #else / #endif, just take the MPXE specific parts and add them to the original code with a #ifdef / #endif. There should be examples in drivers on the `mpxe` branch. Sorry about this, we're still ironing out the details.

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -101,3 +102,174 @@ StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "soft_timer.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+static MxAdt7476aStore s_store = MX_ADT7476A_STORE__INIT;
+static uint32_t speed[NUM_ADT_PWM_PORTS] = { 0 };
```
g-jessmuir, line 121: please use `s_speed` and `s_status`

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -101,3 +102,174 @@ StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
...
+static bool status[NUM_ADT_PWM_PORTS] = { 0 };
+
+static void prv_update_speed() {
```
g-jessmuir, line 124: I think we should just have one general export function; if prv_update_speed is called it could incorrectly set the state of the python since it doesn't update everything together.

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -101,3 +102,174 @@ StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
...
+    if (mask->status[i] != 0) {
+      speed[i] = msg->speed[i];
+      status[i] = msg->status[i];
+    }
+  }
+
+  mx_adt7476a_store__free_unpacked(msg, NULL);
+  mx_adt7476a_store__free_unpacked(mask, NULL);
+  prv_update_speed();
+  prv_update_status();
```
g-jessmuir, line 153: collapse to one export function

g-jessmuir (changes requested):
Seems mostly fine, importantly just check comment on adt747a.c.

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -2,6 +2,168 @@
...
+#include "interrupt.h"
+#include "soft_timer.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+static MxAdt7476aStore s_store = MX_ADT7476A_STORE__INIT;
+static uint32_t s_speed[NUM_ADT_PWM_PORTS] = { 0 };
+static bool s_status[NUM_ADT_PWM_PORTS] = { 0 };
+
+static void prv_update_store() {
```
g-jessmuir, line 23: this should be `prv_export_store()`, I guess just by convention we've been going with 'export' means push something to python, whereas 'update' means read something from python and update the local value

libraries/ms-drivers/src/adt7476a_fan_controller.c
g-jessmuir: I think you might've misunderstood what I meant by #ifdefs within the original code, check the mcp2515.c driver on `mpxe` branch for what I mean by specific parts, and check with me if you have questions

g-jessmuir (changes requested):
Seems mostly good, just check the comment about using multiple #ifdefs

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -27,36 +78,53 @@ StatusCode adt7476a_set_speed(I2CPort port, uint8_t speed_percent, AdtPwmPort pw
...
+    s_status[0] = reg1 ? true : false;
+    s_status[1] = reg2 ? true : false;
+
+    prv_export_store();
+  #endif
+  return (reg1 || reg2) ? STATUS_CODE_UNKNOWN : STATUS_CODE_OK;
 }
 
 StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
+  #ifdef MPXE
```
g-jessmuir, line 125: per google styleguide we (mostly) follow, https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives, preprocessor directives go at the start of the line. So remove the spaces before any #, but keep the indentation of the inner code the same as the other code in the block

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -27,36 +78,53 @@ StatusCode adt7476a_set_speed(I2CPort port, uint8_t speed_percent, AdtPwmPort pw
...
+  StatusCode reg2 =
+      (i2c_read_reg(port, adt7476a_i2c_read_address, ADT7476A_INTERRUPT_STATUS_REGISTER_2,
+                    register_2_data, ADT7476A_REG_SIZE));
+  #ifdef MPXE
+    s_status[0] = reg1 ? true : false;
+    s_status[1] = reg2 ? true : false;
+
+    prv_export_store();
+  #endif
+  return (reg1 || reg2) ? STATUS_CODE_UNKNOWN : STATUS_CODE_OK;
```
g-jessmuir, line 121: I think this can remain as STATUS_CODE_OK, since the point of the return here is to indicate whether or not we successfully got the status, not whether or not the status is ok

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -10,6 +9,58 @@
...
 
+#ifdef MPXE
+#include <stdlib.h>
+#include "adt7476a.pb-c.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+static MxAdt7476aStore s_store = MX_ADT7476A_STORE__INIT;
+static uint32_t s_speed[NUM_ADT_PWM_PORTS] = { 0 };
+static bool s_status[NUM_ADT_PWM_PORTS] = { 0 };
```
g-jessmuir, line 20: it seems like whenever we set `s_speed` or `s_status` we export the store right after. Instead can we remove these arrays and just set the status directly in s_store?

g-jessmuir (changes requested):
just some formatting + some simplification

libraries/ms-drivers/src/adt7476a_fan_controller.c
JarvisWeng: Sorry? what do you mean about that? You mean just set `s_store.speed[0]` directly and call the store export after each function?

libraries/ms-drivers/src/adt7476a_fan_controller.c
JarvisWeng: I got rid of the `status_ok_or_return` function. Will that still return if it fails then?

g-jessmuir: i2c_read will never fail on x86 so we don't need to worry there

libraries/ms-drivers/src/adt7476a_fan_controller.c
g-jessmuir: yeah, do you think that would significantly reduce legibility? seems like an ok cut to me

JarvisWeng: sure

g-jessmuir (approved):

PR #212, opened 2020-10-29
[SOFT-323] Fix issues with bts7200_stop and bts7040_stop

hewittmcg:
It appears that there's a small chance that bts7200_stop and bts7040_stop will return false even if the measurement timer is active when they're called.  This has caused Travis CI to unexpectedly fail a few times.

Since we don't use the return value of these functions anywhere in the codebase, and it doesn't seem likely that we will in future, this changes them to void functions to avoid any issues with this, and changes the test sequence correspondingly so we (hopefully) don't have random failures anymore.


copperium (approved):
LGTM

PR #215, opened 2020-10-30
fw 103 hw

aryan9154:


projects/fw_103/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+  // initialize libraries
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+  gpio_it_init();
+
+  // assign ports and pins for ADC input and Button input
+  GpioAddress adc_button_addresses[] = {
+    [ADC_IN] = { .port = GPIO_PORT_A, .pin = 6 },  //
+    [BTN_IN] = { .port = GPIO_PORT_B, .pin = 2 },  //
```
copperium, line 31: I like the spirit, but these constants `ADC_IN` and `BTN_IN` aren't defined, so this won't compile. Perhaps two variables would be better instead?

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+    .resistor = GPIO_RES_NONE,          //
+    .alt_function = GPIO_ALTFN_ANALOG,  //
+  };
+
+  // initialize settings for button interrupt
+  InterruptSettings btn_interrupt_settings =
+      {
+        .type = INTERRUPT_TYPE_INTERRUPT,       //
+        .priotity = INTERRUPT_PRIORITY_NORMAL,  //
+      }
```
copperium, line 47: Typo: `priotity`. Also you need a semicolon on the end here.

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+  // initialize adc
+  gpio_init_pin(&adc_button_addresses[0], &adc_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  // set channel to default (0), then get the channel using the get func with the port/pin
+  // set input cahnnel to true, enable it
+  AdcChannel in_channel = NUM_ADC_CHANNELS;
+
+  adc_get_channel(adc_button_addresses[0], &in_channel) adc_set_channel(in_channel, true);
```
copperium, line 58: You're going to need a semicolon between these two statements.

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+
+  // initialize the button interrupt with btn info and callback func
+  gpio_it_register_interrupt(&adc_button_addresses[1], &btn_interrupt_settings,
+                             INTERRUPT_EDGE_RISING, prv_btn_interrupt_handler, &in_channel);
```
copperium, line 62: The FW103 homework says:
> The adc reading should trigger whenever the button is released, not initially when the button is pressed.

The button is active-high, which means that the signal is high when the button is pressed and low when it's released. So a rising interrupt edge would mean that the interrupt is triggered when the button is first pressed (the signal "rises" from low to high). Which interrupt edge would make the interrupt get triggered when the button is released?

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+
+// A6 input, B2 input, log output
+
+// callback func, takes in the channel and reads ADC data from channel into local var
+// then logs the var data
+static void prv_btn_interrupt_handler(const GpioAddress *address, void *context) {
+  AdcChannel *in_channel = (AdcChannel *)context;
+  uint16_t in_data = 0;
+
+  adc_read_raw(*in_channel, &in_data);
```
copperium, line 17: The homework does say to use an adc *converted* reading, rather than a raw read like this. The difference is that converted readings with `adc_read_converted` are scaled so that the unit is millivolts, whereas the raw readings are just whatever garbage came out of the ADC.

Incidentally, there is a new API in `adc.h` that takes GpioAddresses instead of AdcChannels, look in adc.h if you're interested. You don't need to use it for this homework, but an FYI for the future.

copperium (changes requested):
Nice! Make sure you always at least try compiling your code with `make build` to make sure it compiles. It's best to try running it with `make run`, but that's not really possible with this project. A few changes before we can call it done.

copperium:
LGTM, closing this since homework is complete.

PR #216, opened 2020-10-31
Soft 334 mpxe ads1259

Max-MZ:


mpxe/sims/ads1259.py
```diff
@@ -0,0 +1,33 @@
...
+
+        ads1259_mask = ads1259_pb2.MxAds1259Store()
+        ads1259_mask.rx_data_lsb = 1
+        ads1259_mask.rx_data_mid = 1
+        ads1259_mask.rx_data_msb = 1
+
+        proj.write_store(stores_pb2.MxStoreType.ADS1259, ads1259_msg, ads1259_mask)
+
+    def assert_store_value_lsb(self, proj, reading):
+        assert(proj.stores[ADS1259_KEY].rx_data_lsb  == reading)
```
g-jessmuir, line 26: nit: you have extra spaces in front of the `==`s here

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -1,6 +1,7 @@
 #include "ads1259_adc.h"
-
+#ifndef MPXE
```
g-jessmuir, line 2: so we recently changed this structure in MPXE, we're now gonna try to just use #ifdefs within the original code since that should make it cleaner and more obvious what's going on in MPXE vs not in MPXE, so rather than having the whole file in a #else / #endif, just take the MPXE specific parts and add them to the original code with a #ifdef / #endif. There should be examples in drivers on the mpxe branch. Sorry about this, we're still ironing out the details.

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -121,3 +122,172 @@ StatusCode ads1259_get_conversion_data(Ads1259Storage *storage) {
...
+#include "log.h"
+#include "math.h"
+#include "soft_timer.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+static MxAds1259Store s_store = MX_ADS1259_STORE__INIT;
+
+static void update_store(ProtobufCBinaryData msg_buf, ProtobufCBinaryData mask_buf) {
+  LOG_DEBUG("UPDATING STORE:\n");
```
g-jessmuir, line 141: let's leave out logs here to stay consistent with other drivers in MPXE

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -121,3 +122,172 @@ StatusCode ads1259_get_conversion_data(Ads1259Storage *storage) {
...
+  }
+  return STATUS_CODE_OK;
+}
+
+// calculate check-sum based on page 29 of datasheet
+static Ads1259StatusCode prv_checksum(Ads1259Storage *storage) {
+  uint8_t sum = (uint8_t)(s_store.rx_data_lsb + s_store.rx_data_mid + s_store.rx_data_msb +
+                          ADS1259_CHECKSUM_OFFSET);
+  LOG_DEBUG("STORAGE CHK_SUM: %x\n", (uint8_t)s_store.chk_sum);
+  LOG_DEBUG("CALCULATED SUM %x\n", sum);
```
g-jessmuir, line 224: should probably not have logs here, logic is that if we're emulating a ramp or other continuous function in python for the current this would print a lot, it's also not super useful info (not obviously actionable / interpretable to reader)

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -121,3 +122,172 @@ StatusCode ads1259_get_conversion_data(Ads1259Storage *storage) {
...
+  spi_exchange(storage->spi_port, payload, 1, (uint8_t *)&storage->rx_data, NUM_ADS_RX_BYTES);
+  code = prv_checksum(storage);
+  (*storage->handler)(code, storage->error_context);
+
+  storage->conv_data.MSB = (uint8_t)s_store.rx_data_msb;
+  storage->conv_data.MID = (uint8_t)s_store.rx_data_mid;
+  storage->conv_data.LSB = (uint8_t)s_store.rx_data_lsb;
+  LOG_DEBUG("msb data %u \n", storage->conv_data.MSB);
+  LOG_DEBUG("mid data %u \n", storage->conv_data.MID);
+  LOG_DEBUG("lsb data %u \n", storage->conv_data.LSB);
```
g-jessmuir, line 261: should also probably not have logs here for same reasoning as above except more so

libraries/ms-drivers/src/ads1259_adc.c
```diff
@@ -121,3 +122,172 @@ StatusCode ads1259_get_conversion_data(Ads1259Storage *storage) {
...
+  Ads1259Storage *storage = (Ads1259Storage *)context;
+  Ads1259StatusCode code;
+  uint8_t payload[] = { ADS1259_READ_DATA_BY_OPCODE };
+  spi_exchange(storage->spi_port, payload, 1, (uint8_t *)&storage->rx_data, NUM_ADS_RX_BYTES);
+  code = prv_checksum(storage);
+  (*storage->handler)(code, storage->error_context);
+
+  storage->conv_data.MSB = (uint8_t)s_store.rx_data_msb;
+  storage->conv_data.MID = (uint8_t)s_store.rx_data_mid;
+  storage->conv_data.LSB = (uint8_t)s_store.rx_data_lsb;
```
g-jessmuir, line 258: could we actually change the structure of this to just insert the final converted value from python into `storage->reading`? my logic is that anytime we want to mock with this we'll want to insert a real voltage value, but it's much harder to try to insert a bunch of bytes (since we'd basically have to do the conversion backwards). So basically the last line of `prv_convert_data()` would be `storage->reading = s_store->reading;`, but also lemme know your thoughts on this

g-jessmuir (changes requested):
Main comment is that I think we should insert a real value rather than the raw ADC bytes read, also we're gonna move to using inline #ifdefs rather than duplicating all the code

libraries/ms-drivers/src/ads1259_adc.c
Max-MZ: We could, but wouldn't that mess with the checksum? Since it's calculated directly from `storage->rx_data` which has the `msb`, `lsb`, and `mid` elements.

g-jessmuir: I don't think so, the checksum is internal to the driver only, so if you return the inserted reading at the end of `prv_convert_data` it shouldn't affect anything other than external driver users

Max-MZ: Ah yeah my bad I interpreted that as replacing the 3 stores as one `reading` store.

g-jessmuir (approved):

PR #218, opened 2020-10-31
[SOFT-322] MPXE adc

anthony-chang:
…o enum

### DIFF
--- a/libraries/ms-common/src/x86/adc.c
+++ b/libraries/ms-common/src/x86/adc.c
@@ -1,6 +1,5 @@
 #include "adc.h"
 
-#ifndef MPXE
 #include <stddef.h>
 #include "log.h"
 
@@ -11,6 +10,18 @@
 // Vdda locked at 3300 mV.
 // adc_read_converted should always return close to 2V
 // temperature reading always returns 293 kelvin.
+
+#ifdef MPXE
+#include <stdlib.h>
+
+#include "adc.pb-c.h"
+#include "status.h"
+#include "store.h"
+#include "stores.pb-c.h"
+
+static MxAdcStore s_store = MX_ADC_STORE__INIT;
+
+#endif
 
 #define ADC_RETURNED_VOLTAGE_RAW 2500
 #define ADC_CONTINUOUS_CB_FREQ_MS 50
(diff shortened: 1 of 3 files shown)


mpxe/protos/adc.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto3";
+
+message MxAdcStore {
+  repeated uint32 reading = 2;
```
g-jessmuir, line 4: nit: why is this field number 2? I feel like we should just default to using the smallest number (1)

libraries/ms-common/src/x86/adc.c
```diff
@@ -176,32 +178,200 @@ StatusCode adc_read_converted(AdcChannel adc_channel, uint16_t *reading) {
   return STATUS_CODE_OK;
 }
 
-// the following functions are wrappers over the legacy AdcChannel API dealing with GpioAddresses
```
g-jessmuir, line 179: seems like your branch might be outdated (these shouldn't be being deleted), maybe merge `mpxe` in or rebase

libraries/ms-common/src/x86/adc.c
```diff
@@ -1,7 +1,9 @@
+#include "adc.h"
+
+#ifndef MPXE
```
g-jessmuir, line 3: so we recently changed this structure in MPXE, we're now gonna try to just use #ifdefs within the original code since that should make it cleaner and more obvious what's going on in MPXE vs not in MPXE, so rather than having the whole file in a #else / #endif, just take the MPXE specific parts and add them to the original code with a #ifdef / #endif. There should be examples in drivers on the mpxe branch. Sorry about this, we're still ironing out the details.

libraries/ms-common/src/x86/adc.c
```diff
@@ -176,32 +178,200 @@ StatusCode adc_read_converted(AdcChannel adc_channel, uint16_t *reading) {
...
+  adc_set_channel(ADC_CHANNEL_REF, true);
+}
+
+StatusCode adc_set_channel(AdcChannel adc_channel, bool new_state) {
+  if (adc_channel >= NUM_ADC_CHANNELS) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+  s_active_channels[adc_channel] = new_state;
+
+  prv_export();
```
g-jessmuir, line 272: this doesn't actually change any state so we shouldn't need to export here

libraries/ms-common/src/x86/adc.c
```diff
@@ -176,32 +178,200 @@ StatusCode adc_read_converted(AdcChannel adc_channel, uint16_t *reading) {
...
+
+    default:
+      break;
+  }
+
+  uint16_t vdda;
+  adc_read_converted(ADC_CHANNEL_REF, &vdda);
+  *reading = (adc_reading * vdda) / 4095;
+
+  s_store.reading[adc_channel] = *reading;
```
g-jessmuir, line 371: We shouldn't be setting the reading here. This function reads a value from the hardware, but the hardware is simulated by python, so setting the value here would be overwriting the value set by python. Instead we should just return the value set by python as in `return s_store.reading[adc_channel]`

libraries/ms-common/src/x86/adc.c
```diff
@@ -176,32 +178,200 @@ StatusCode adc_read_converted(AdcChannel adc_channel, uint16_t *reading) {
...
+  uint16_t reading;
+} AdcInterrupt;
+
+static MxAdcStore s_store = MX_ADC_STORE__INIT;
+
+static AdcInterrupt s_adc_interrupts[NUM_ADC_CHANNELS];
+
+static bool s_active_channels[NUM_ADC_CHANNELS];
+
+static void prv_export() {
```
g-jessmuir, line 210: since this library is purely for sensing hardware values and not for setting them, we never need to export data to python

g-jessmuir (changes requested):
Main things are that we changed format to just add inline #ifdefs in the original code rather than creating a whole new version of the library, and also since ADC is purely a sensor library we never need to export to python, lemme know if that doesn't make sense

g-jessmuir (approved):

PR #219, opened 2020-10-31
[SOFT-328] Fix BTS7200 validation issues

copperium:
- fix the BTS7200 smoke test by filling in fields added in driver update
- fix uint16 overflow in BTS7200 and BTS7040 drivers when converting voltage -> current
- add a bias field in BTS7200 and BTS7040 so we can subtract any bias found in the output
- configure all of this for power distro and the smoke test using the experimentally determined resistor value (1160, making a scaling factor of 670/1160=0.578) and bias (-7.5, rounded to -8)


hewittmcg (approved):
LGTM! 🎃🎃🎃

g-jessmuir (approved):
Seems good 👍

PR #220, opened 2020-10-31
[SOFT-330] Babydriver project: firmware implementation of gpio_get 

sudhish3120:


projects/baby_driver/inc/gpio_get.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+// This module provides the function gpio_get_init(), which can be used to listen
+// for gpio_get CAN messages from the babydriver. Once the message is received,
+// the state of the requested GPIO pin is obtained and sent back to the baby driver using
+// CAN_TRANSMIT_BABYDRIVER
```
copperium, line 6: According to the [coding standards](https://uwmidsun.atlassian.net/l/c/Rhbg3XGT), the header description comment format should be:
```c
// One line description of the module.
// Requires ... to be initialized.

// [More description if desired]
```
In particular your module should have which modules/libraries it required to be initialized in the header so that people using the module (e.g. you, writing your test file) know what needs to be initialized before using the module. That includes all transitive dependencies - e.g. you depend on `dispatcher`, which in turn depends on interrupts, soft timers, the event queue, and CAN, so you should list all of them.

copperium (changes requested):
Awesome, just a few comments. You can run your tests with `make test PLATFORM=x86 PROJECT=baby_driver TEST=gpio_get`.

projects/baby_driver/test/test_gpio_get.c
```diff
@@ -0,0 +1,178 @@
...
+    .resistor = GPIO_RES_NONE,        //
+    .alt_function = GPIO_ALTFN_NONE,  //
+  };
+
+  GpioAddress address = {
+    .port = VALID_PORT,  //
+    .pin = VALID_PIN,    //
+  };
+
+  TEST_ASSERT_OK(gpio_init_pin(&address, &settings));
```
copperium, line 134: Okay, so here's the deal: on x86, if a pin is initialized as an input, the `gpio` library will return values from an uninitialized array you can't modify in `gpio_get_state` - this is a design flaw with the library. So, with the initialization corrected above, you won't be able to modify the state of the gpio pin.

However, to save the day are what is called *mocks*. When you mock a function, you write an alternate implementation of that function which is called instead of the original function within a test. So, I would suggest that you mock `gpio_get_state`, so that you can return a value of your choosing instead of relying on the whims of the `gpio` library. We have some macro magic for mocks, so it'll look like this:
```c
StatusCode TEST_MOCK(gpio_get_state)(const GpioAddress *address, GpioState *state) {
  // your alternate implementation of gpio_get_state...
  *state = s_returned_state; // for example
  return STATUS_CODE_OK;
}
```
Then you just have to add the following line to projects/baby_driver/rules.mk to inform the build system about your mock:
```makefile
$(T)_adc_read_MOCKS := gpio_get_state
```
Then instead of actually initializing a GPIO pin to set the state returned by `gpio_get_state`, you can just set a static variable, avoiding the whole initialization problem with the `gpio` library.

copperium (changes requested):

copperium (approved):

PR #222, opened 2020-11-04
[SOFT-348] Finalize drv120 update

mitchellostler:
Changes made:

1. Drv120_driver 
- Created settings struct, which takes pin, status pin, GpioItCallback, context
- When drv120_relay_init() called with settings struct, will check if status pin is null, if not, initializes gpio_it
- Interrupt triggered on falling edge of STATUS pin -> see data sheet 

2. Implementation
- It was decided that callback should be defined in relay_fsm.c and set in main.c, as it calls fault_handler_raise_fault()
- Settings struct created in main.c, passed pin getter functions defined in solar_config.h: config_get_drv120_status_pin() is currently set to return NULL, and must be altered  to commented return value for this functionality to work
- Solar fault EE_SOLAR_FAULT_DRV120 added, this is raised by fault_handler_raise_fault if gpio_it_triggered


libraries/ms-drivers/inc/drv120_relay.h
```diff
@@ -1,17 +1,28 @@
...
 #include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
 #include "status.h"
 
+typedef struct Drv120RelaySettings {
+  GpioAddress *pin;        // Pin initialized as output for relay
+  GpioAddress *status;     // Pin set to trigger interrupt on error
+  GpioItCallback handler;  // Error handler
+  void *context;           // Context for error handler
```
copperium, line 18: Nitpicks on the names:
* I know `pin` was the parameter name for `drv120_relay_init`, but it's vague now since there's more than one pin. I'd prefer `enable_pin` or something since it's the relay enable pin.
* Similarly, could we call `status` instead `status_pin`? Seeing `.status = ...` in a settings struct initialization would make me think it takes a StatusCode.
* If we call `handler` instead `error_handler` or `error_callback` then you don't need the comment on it anymore.

libraries/ms-drivers/inc/drv120_relay.h
```diff
@@ -1,17 +1,28 @@
...
 
 #include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
 #include "status.h"
 
+typedef struct Drv120RelaySettings {
+  GpioAddress *pin;        // Pin initialized as output for relay
+  GpioAddress *status;     // Pin set to trigger interrupt on error
+  GpioItCallback handler;  // Error handler
```
copperium, line 17: Reusing `GpioItCallback` here means that every callback needs a `const GpioAddress *` parameter that they shouldn't have to care about. To make this nicer, I'd suggest `typedef`ing a new `Drv120ErrorCallback` function pointer type that just takes a `void *context`. Then you can just save that callback in a static variable alongside `s_relay_pin` and call it in a static gpio interrupt callback function.

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -1,15 +1,26 @@
...
+// Gpio_it callback function
+static bool s_gpio_it_callback_called = false;
+void prv_callback(const GpioAddress *address, void *context) {
+  LOG_DEBUG("DRV120 Interrupt Triggered!\n");
+  s_gpio_it_callback_called = true;
+}
+
 // Pin used to enable/disable relay
 static const GpioAddress s_test_drv120_pin = { GPIO_PORT_A, 8 };
+static const GpioAddress s_test_drv120_status = { GPIO_PORT_A, 9 };
```
copperium, line 17: If we're in the business of sneakily choosing "test" gpio pins to align with the real ones, I'd suggest making this PA6, since that's the real status pin.

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -44,3 +64,16 @@ void test_drv120_relay(void) {
...
   TEST_ASSERT_FALSE(closed);
 }
+
+void test_callback_registered(void) {
+  LOG_DEBUG("Initializing gpio interrupt\n");
+  Drv120RelaySettings settings = {
+    .pin = &s_test_drv120_pin,
+    .status = &s_test_drv120_status,
+    .handler = prv_callback,
+    .context = NULL,
```
copperium, line 74: Might also want to assign an arbitrary non-null context and test that it was received so you know that contexts are passed correctly.

projects/solar/src/relay_fsm.c
```diff
@@ -59,3 +59,8 @@ StatusCode relay_fsm_close(void) {
 StatusCode relay_fsm_open(void) {
   return event_raise_priority(RELAY_EVENT_PRIORITY, SOLAR_RELAY_EVENT_OPEN, 0);
 }
+
+void relay_err_cb(const GpioAddress *address, void *context) {
+  LOG_DEBUG("DRV120 RELAY ERR TRIGGERED\n");
+  fault_handler_raise_fault(EE_SOLAR_FAULT_DRV120, 0);
+}
```
copperium, line 66: Instead of having this be a public function, I would suggest moving the drv120 initialization to `relay_fsm_init` and making this function static. (I think this makes sense because relay_fsm "owns" the drv120 in some sense.)

Also, `fault_handler_raise_fault` already logs, so you don't have to here.

projects/solar/inc/solar_config.h
```diff
@@ -29,6 +29,8 @@ const CanSettings *config_get_can_settings(void);
 
 const GpioAddress *config_get_drv120_relay_pin(void);
```
copperium, line 30: Could you rename `config_get_drv120_relay_pin` to `config_get_drv120_enable_pin` since it's the enable pin, not the relay pin?

projects/solar/src/solar_config.c
```diff
@@ -96,6 +96,7 @@ static const CanSettings s_can_settings = {
 };
 
 static const GpioAddress s_drv120_relay_pin = { GPIO_PORT_A, 8 };
+static const GpioAddress s_drv120_status_pin = { NUM_GPIO_PORTS, 0 };
```
copperium, line 99: Like I mentioned above, the status pin is PA6.

projects/solar/src/solar_config.c
```diff
@@ -136,6 +137,12 @@ const GpioAddress *config_get_drv120_relay_pin(void) {
   return &s_drv120_relay_pin;
 }
 
+// To enable status interrupt for drv120, add pin below, and return its address here
+const GpioAddress *config_get_drv120_status_pin(void) {
+  // return &s_drv120_status_pin;
+  return NULL;
+}
```
copperium, line 144: If you want to make it easy to disable the status pin, I'd suggest a `#define` at the top like `#define ENABLE_DRV120_FAULT_HANDLING true`. Then here you can do:
```c
return ENABLE_DRV120_FAULT_HANDLING ? &s_drv120_status_pin : NULL;
```

projects/solar/test/test_relay_fsm.c
```diff
@@ -26,13 +30,35 @@
   })
 
 static const GpioAddress s_test_relay_pin = { .port = GPIO_PORT_A, .pin = 6 };
+static const GpioAddress s_test_status_pin = { .port = GPIO_PORT_A, .pin = 7 };
 
 static CanStorage s_can_storage;
 static RelayFsmStorage s_storage;
 
+static bool relay_err_rx_cb_called;
+StatusCode prv_err_rx_cb(const CanMessage *msg, void *context, CanAckStatus *ack_reply) {
```
copperium, line 39: This should be static eh?

copperium (changes requested):

libraries/ms-drivers/src/drv120_relay.c
```diff
@@ -1,10 +1,16 @@
...
+#include "log.h"
 
 // Storage for EN pin on DRV120 to open/close relay
 static GpioAddress s_relay_pin;
+static Drv120ErrorCallback s_callback;
 
-StatusCode drv120_relay_init(GpioAddress *pin) {
-  s_relay_pin = *pin;
+static void prv_drv120_gpio_it_cb(const GpioAddress *address, void *context) {
+  s_callback(context);
```
copperium, line 9: Add a guard so we don't call `s_callback` if it's null to avoid segfaults. (User code *should* always set `error_handler` to something non-null if `status_pin` is set, but if they don't the resulting segfault here is near impossible to debug.)

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -1,15 +1,30 @@
 // Brief test sequence for DRV120
 
+#include "delay.h"
 #include "drv120_relay.h"
 #include "log.h"
 #include "ms_test_helpers.h"
 
+// Gpio_it callback function
+static bool s_gpio_it_callback_called = false;
+static int rcv_context;
```
copperium, line 10: Use a `uint8_t` or some other fixed-width type instead of `int`

libraries/ms-drivers/test/test_drv120_relay.c
```diff
@@ -44,3 +68,18 @@ void test_drv120_relay(void) {
   TEST_ASSERT_OK(drv120_relay_get_is_closed(&closed));
   TEST_ASSERT_FALSE(closed);
 }
+
+void test_callback_registered(void) {
+  LOG_DEBUG("Initializing gpio interrupt\n");
+  int test_ctx = 1;
```
copperium, line 74: Here too

projects/solar/src/solar_config.c
```diff
@@ -132,10 +136,15 @@ const CanSettings *config_get_can_settings(void) {
   return &s_can_settings;
 }
 
-const GpioAddress *config_get_drv120_relay_pin(void) {
+const GpioAddress *config_get_drv120_enable_pin(void) {
   return &s_drv120_relay_pin;
 }
 
+// To enable status interrupt for drv120, add pin below, and return its address here
```
copperium, line 143: This comment's outdated, eh?

copperium (changes requested):
A few minor things

copperium (approved):

PR #223, opened 2020-11-04
Soft 331 gpio get python

Kyle02:


projects/baby_driver/scripts/gpio_get.py
```diff
@@ -0,0 +1,44 @@
...
+
+    status_msg = can_util.next_message(
+        babydriver_id=BabydriverMessageId.GPIO_GET_DATA,
+        channel=can_util.default_channel
+    )
+
+    if status_msg.msg.data[0] != BabydriverMessageId.STATUS:
+        raise Exception("ERROR: never recieved status message")
+
+    msg_data = gpio_data_msg.msg.data[0]
```
copperium, line 40: A few things about this line:
1. Like above, you don't need to access the `msg` property, just `gpio_data_msg.data` will work
2. It should be `data[1]` since the pin state is in the 2nd uint8; the first uint8 is the babydriver ID.
3. `msg_data` is a bit of a vague name, perhaps `raw_state` would be better.

projects/baby_driver/scripts/gpio_get.py
```diff
@@ -0,0 +1,44 @@
...
+
+    gpio_pin_state = msg_data == 1
```
copperium, line 42: Nit: if the firmware side theoretically returns a state value which isn't 0 or 1, we should return True for any non-zero value. So this should really be `msg_data != 0`, or even `bool(msg_data)`.

(Also, it looks cleaner if you put brackets around the `==` or `!=` when you're assigning to such a comparison: `gpio_pin_state = (msg_data != 0)`.)

projects/baby_driver/scripts/gpio_port.py
```diff
@@ -0,0 +1,14 @@
+"""Constants for GpioPort"""
+
+class GpioPort():
```
copperium, line 3: You don't need the brackets () at the end of `GpioPort`.

projects/baby_driver/scripts/message_defs.py
```diff
@@ -17,3 +17,7 @@ class BabydriverMessageId:
     """
 
     STATUS = 0
+
+    #gpio_get constants
```
copperium, line 21: nitpick: please please always put a space after the `#` in a comment, though this one is unnecessary

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+        '''Tests minimum values for port and pin'''
+        gpio_pin_data = [1]
+        gpio_pin_msg = Message(
+            message_id=BABYDRIVER_CAN_MESSAGE_ID,
+            device_id=BABYDRIVER_DEVICE_ID,
+            data=gpio_pin_data,
+        )
+
+        status_data = [0]
+        status_msg = Message(
```
copperium, line 26: This data lists need to include the babydriver ID as the first element, since that's the protocol. I'd also prefer that you use `can_pack` to pack the data even though it's technically unnecessary. (Also applies to the other message data lists below.)

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(True, gpio_get(GpioPort.A, 0))
+
+        gpio_pin_data = [0]
+        gpio_pin_msg = Message(
+            message_id=BABYDRIVER_CAN_MESSAGE_ID,
+            device_id=BABYDRIVER_DEVICE_ID,
+            data=gpio_pin_data,
+        )
+        mock_next_message.side_effect = [gpio_pin_msg, status_msg]
+        self.assertEqual(False, gpio_get(GpioPort.A, 0))
```
copperium, line 43: There are `assertTrue` and `assertFalse` methods you can call instead of `assertEqual`. (Also applies to the assertEqual calls below.)

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+            data=gpio_pin_data,
+        )
+
+        status_data = [0]
+        status_msg = Message(
+            message_id=BABYDRIVER_CAN_MESSAGE_ID,
+            device_id=BABYDRIVER_DEVICE_ID,
+            data=status_data,
+        )
+        mock_send_message.side_effect = None
```
copperium, line 31: I don't think you explicitly have to set `mock_send_message.side_effect = None`. (Also applies below.)

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(True, gpio_get(GpioPort.F, 15))
+
+        gpio_pin_data = [0]
+        gpio_pin_msg = Message(
+            message_id=BABYDRIVER_CAN_MESSAGE_ID,
+            device_id=BABYDRIVER_DEVICE_ID,
+            data=gpio_pin_data,
+        )
+        mock_next_message.side_effect = [gpio_pin_msg, status_msg]
+        self.assertEqual(False, gpio_get(GpioPort.F, 15))
```
copperium, line 74: Once you implement it, you should also test in a place or two that you can pass strings as the port, like `gpio_get('f', 15)` here instead

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_fail_conditions(self, mock_next_message, mock_send_message):
+        '''Tests fail conditions'''
+        gpio_pin_data = [1]
+        gpio_pin_msg = Message(
+            message_id=BABYDRIVER_CAN_MESSAGE_ID,
+            device_id=BABYDRIVER_DEVICE_ID,
+            data=gpio_pin_data,
+        )
+
+        status_data = [1]
```
copperium, line 88: You should add a comment saying why this should cause a failure

projects/baby_driver/scripts/test_gpio_get.py
```diff
@@ -0,0 +1,120 @@
...
+            data=status_data,
+        )
+
+        #invalid port fail
+        mock_next_message.side_effect = [gpio_pin_msg, status_msg]
+        self.assertRaises(ValueError, gpio_get, -1, 0)
+        mock_next_message.side_effect = [gpio_pin_msg, status_msg]
+        self.assertRaises(ValueError, gpio_get, 16, 0)
+
+        #invalid pin fail
```
copperium, line 112: please put a space after the `#` in comments!

projects/baby_driver/scripts/can_util.py
```diff
@@ -114,7 +114,6 @@ def send_message(
         data=data,
         is_extended_id=False
     )
-
```
copperium, line 117: did you delete this accidentally?

copperium (changes requested):

projects/baby_driver/scripts/can_util.py
Kyle02: Yes I did, I was trying to see what the can message looked like, so I inserted a print statement there and deleted it, I may have deleted an extra line by accident

projects/baby_driver/scripts/gpio_get.py
```diff
@@ -0,0 +1,47 @@
...
+    gpio_data_msg = can_util.next_message(
+        babydriver_id=BabydriverMessageId.GPIO_GET_DATA,
+    )
+
+    status_msg = can_util.next_message(
+        babydriver_id=BabydriverMessageId.STATUS,
+    )
+
+    if status_msg.data[1] != 0:
+        raise Exception("ERROR: never recieved status message")
```
copperium, line 41: When this exception is raised, we *have* received the status message - the status is just nonzero (i.e. non-ok). So this error message should be something like `"Non-OK status returned: {}".format(status_msg.data[1])`; even better would be to assign `status_msg.data[1]` to a variable like `status` so that we're not repeating the cryptic array access.

copperium (changes requested):
Literally just one thing and then we can get this merged!

copperium (approved):

PR #226, opened 2020-11-09
homework 102

abdullahs26:


projects/firmware_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  Counters *count = context;
+  count->counter_a += 1;
+  LOG_DEBUG("Counter A:%d\n", count->counter_a);
+
+  if (count->counter_a % 2 == 0 && count->counter_a != 0) {
```
AveryChiu64, line 19: I don't think you need to check count->counter_a != 0 since you increment counter_a before this line. This isn't a huge deal but just a bit redundant.

projects/firmware_102/src/main.c
```diff
@@ -0,0 +1,44 @@
+#include <stdbool.h>
```
AveryChiu64, line 1: It's good that you put comments in your code and I think you did this for learning and to remind yourself of what everything does in the library. I encourage you to continue doing this if you are not sure what something does. However, when you are actually writing code and pushing it into the repo some of these comments are sort of self-explanatory. For example, I don't think you need to make a commend that the wait.h header includes the wait function.

projects/firmware_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+  Counters *count = context;
+  count->counter_a += 1;
+  LOG_DEBUG("Counter A:%d\n", count->counter_a);
+
+  if (count->counter_a % 2 == 0 && count->counter_a != 0) {
+    count->counter_b += 1;
+    LOG_DEBUG("Counter B:%d\n", count->counter_b);
+  }
+
+  soft_timer_start_millis(500,                 // timer duration
```
AveryChiu64, line 24: These comments are not really necessary as well, but this is fine if you are just doing this for your learning. As I said above, make sure when you are actually writing code that will be a part of this repo that redundant comments like these are removed. We do this to make sure things don't get cluttered. Also, the more documentation you write, the more it needs to be managed when changes are made and the more work you need to do.

.metadata/.plugins/org.eclipse.cdt.make.core/specs.c
```diff
@@ -0,0 +1 @@
+
```
AveryChiu64, line 1: I think you opened the repo in eclipse or something. Please do not commit all these other extraneous files for your other projects. This will clutter the repo.

AveryChiu64 (changes requested):
Exercise is good, please remove all the extraneous files when commiting. I think you should switch to an editor like VScode which goes not generate all these files. I have left some very minor comments in the exercise, but it is generally good. Keep up the good work.

projects/firmware_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+                          NULL);
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters count = { 0 };
+
+  soft_timer_start_millis(500,                 // timer duration
```
AveryChiu64, line 36: Though you comment on the timer duration, it would be better if you defined it from the beginning rather than hardcoding it as 500. Reading through this, if you did not put the comment as 500. I would not know what this is. Is this 500 seconds? 500 miliseconds? Make sure to define this at the top.

projects/firmware_102/src/main.c
AveryChiu64: Almost forgot to add, make sure to #define the time in milliseconds at the top. Right now it is hardcoded which is not good practice.

PR #228, opened 2020-11-10
homework 102

abdullahs26:


projects/fw_102/src/fw_102.c
```diff
@@ -0,0 +1,38 @@
+#include <stdlib.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+#define COUNTER_INCREMENT_MS 500;  // 500 milliseconds between incrementing counter A
```
AveryChiu64, line 6: I see that you've defined 500 up here but you forgot to pass this into your calls to start the soft timer. The whole point of defining COUNTER_INCREMENT_MS is so we can use it later rather than just hardcoding 500 (ie we sort of use COUNTER_INCREMENT_MS like a const variable so we don't have to hardcode the numbers into our code).
If you aren't sure why we use #define versus a const variable, observe the following 
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1591803936/C+Reference+for+Midnight+Sun+Firmware+Team

projects/fw_102/src/fw_102.c
```diff
@@ -0,0 +1,38 @@
...
+  count->counter_a += 1;
+  LOG_DEBUG("Counter A:%d\n", count->counter_a);
+
+  if (count->counter_a % 2 == 0) {  // this is so counter b is incremented every other function
+                                    // call, so it is incremented every 1000 ms
+    count->counter_b += 1;
+    LOG_DEBUG("Counter B:%d\n", count->counter_b);
+  }
+
+  soft_timer_start_millis(500, prv_timer_callback, count, NULL);
```
AveryChiu64, line 24: Need to pass in COUNTER_INCREMENT_MS rather than 500 here and everywhere else you call this function.

projects/fw_102/src/fw_102.c
AveryChiu64, line 6: I see that you've defined this up here but you haven't actually used it in your code. The whole point of this is so we don't have to hardcode 500. You can pass COUNTER_INCREMENT_MS in when you call your softtimers. The #define sort of acts like declaring a const variable. To understand why we use #define versus const variables, refer to this
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1591803936/C+Reference+for+Midnight+Sun+Firmware+Team

projects/fw_102/src/fw_102.c
AveryChiu64, line 24: Pass in your macro here (ie pass in COUNTER_INCREMENT_MS rather than 500) and make sure to fix this everywhere else you call this as well.

AveryChiu64 (changes requested):
Looking good. Much more cleaner this time. It looks like you just forgot one small thing.

projects/fw_102/src/fw_102.c
abdullahs26: I think I forgot to save it once I made that change to the rest of the code, my apologies. I updated it now.

AveryChiu64: Oh yeah, almost forgot but there is a big mistake here. The #define statement doesn't need a semicolon at the end. Refer to firmware 102 to understand how to use this statement better.
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1282768912/Firmware+102+Supplementary+Material

PR #234, opened 2020-11-25
[SOFT-306] Babydriver: minor adc_read compatibility fixes

copperium:
Bugfixes:
* `adc` wasn't initialized in firmware, causing weird readings
* fixed a couple mixups in message IDs in Python, now C and Python are compatible
* added a guard in dispatcher to avoid possible segfaults when a babydriver message is received which doesn't have a callback registered

Quality of life improvement: in the Python `adc_read` function, `raw` now defaults to False, meaning a converted read is the default.

### DIFF
--- a/libraries/ms-common/test/test_adc.c
+++ b/libraries/ms-common/test/test_adc.c
@@ -11,6 +11,9 @@
 
 #define ADC_MOCK_READING 999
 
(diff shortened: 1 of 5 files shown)


g-jessmuir (approved):
seems good to me

PR #235, opened 2020-11-26
[Soft 380] mpxe log locking cleanup

anthony-chang:


g-jessmuir (changes requested):
I think there's just the makefile change we also need, just add
```
ifneq (,$(IS_MPXE))
$(T)_DEPS += mpxe-gen mpxe-store
endif
```
to libcore/rules.mk and make sure it compiles I think

g-jessmuir (approved):

PR #237, opened 2020-11-28
Soft 370 max6643

Max-MZ:


libraries/ms-drivers/inc/max6643_fan_controller.h
```diff
@@ -0,0 +1,23 @@
...
+} Max6643Settings;
+
+typedef struct {
+  GpioAddress fanfail_pin;
+  GpioAddress overtemp_pin;
+  GpioItCallback fanfail_callback;
+  GpioItCallback overtemp_callback;
+  void *fanfail_callback_context;
+  void *overtemp_callback_context;
+} Max6643Storage;
```
copperium, line 21: You don't actually need a storage eh? Since you're only registering GPIO interrupt callbacks, and you have no other functions which take a storage. You can get away with removing the storage and just working from the settings.

libraries/ms-drivers/inc/max6643_fan_controller.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "gpio.h"
+#include "gpio_it.h"
```
copperium, line 3: At the top here between the #pragma once and the start of the includes you need the header comments:
```c
// Brief description of the driver.
// Requires X, Y, and Z to be initialized.
```
You should also note that we're assuming the high/low temperature threshold selection pins are set in hardware, as is the full speed pin. (I agree with the decision not to include full speed pin handling in the driver since the nand-gate logic is part of the solar board rather than the IC, that logic will go in solar later.)

libraries/ms-drivers/src/max6643_fan_controller.c
```diff
@@ -0,0 +1,35 @@
+#include <stddef.h>
+#include <string.h>
+
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "max6643_fan_controller.h"
```
copperium, line 7: Nitpick: the `#include "max6643_fan_controller.h"` should be at the top of the file separated from the other includes by a blank line.

libraries/ms-drivers/src/max6643_fan_controller.c
```diff
@@ -0,0 +1,35 @@
...
+  storage->fanfail_pin = settings->fanfail_pin;
+  storage->overtemp_callback = settings->overtemp_callback;
+  storage->fanfail_callback = settings->fanfail_callback;
+  storage->fanfail_callback_context = settings->fanfail_callback_context;
+  storage->overtemp_callback_context = settings->overtemp_callback_context;
+
+  static InterruptSettings s_interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_NORMAL,
+  };
```
copperium, line 26: It isn't good practice to declare static variables inside functions - that makes the value persist throughout calls to the function and it's very confusing. In this case you don't need it to take up memory for the whole program by making it static, so just make it a normal non-static variable. (You can make it `const` if you prefer.)

libraries/ms-drivers/test/test_max6643_fan_controller.c
```diff
@@ -0,0 +1,99 @@
...
+  s_mock_statuses.FANFAIL_INTERRUPT_STATUS = 0;
+  s_mock_statuses.OVERTEMP_INTERRUPT_STATUS = 0;
+}
+
+void teardown_test(void) {}
+
+// Test initializing with valid parameters
+void test_max6643_init_works(void) {
+  GpioAddress test_fanfail_pin = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress test_overtemp_pin = { .port = GPIO_PORT_A, .pin = 1 };
```
copperium, line 43: To avoid redeclaring this at the start of each test you can make these static const variables.

libraries/ms-drivers/test/test_max6643_fan_controller.c
```diff
@@ -0,0 +1,99 @@
...
+
+  // add callback to storage
+  TEST_ASSERT_OK(max6643_init(&s_storage, &valid_settings));
+
+  // trigger interrupt and fetch data
+  s_mock_statuses.FANFAIL_INTERRUPT_STATUS = 1;
+  gpio_it_trigger_interrupt(&test_fanfail_pin);
+
+  s_mock_statuses.OVERTEMP_INTERRUPT_STATUS = 1;
+  gpio_it_trigger_interrupt(&test_overtemp_pin);
```
copperium, line 98: This test will pass even if the callbacks aren't called at all! I'd recommend making the callbacks increment some static variable like `s_times_fanfail_callback_called` and checking that after the `gpio_it_trigger_interrupt` call.

libraries/ms-drivers/test/test_max6643_fan_controller.c
```diff
@@ -0,0 +1,99 @@
...
+  GpioAddress test_fanfail_pin = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress test_overtemp_pin = { .port = GPIO_PORT_A, .pin = 1 };
+
+  Max6643Settings valid_settings = {
+    .fanfail_pin = test_fanfail_pin,
+    .overtemp_pin = test_overtemp_pin,
+    .fanfail_callback = (GpioItCallback)prv_test_max6643_fanfail_interrupt_callback,
+    .overtemp_callback = (GpioItCallback)prv_test_max6643_overtemp_interrupt_callback,
+    .fanfail_callback_context = NULL,
+    .overtemp_callback_context = NULL
```
copperium, line 82: You should assign the contexts to some distinct nonnull pointers and check that they were passed correctly by assigning some static variable to the context in the callbacks.

libraries/ms-drivers/test/test_max6643_fan_controller.c
```diff
@@ -0,0 +1,99 @@
+#include "interrupt.h"
+#include "log.h"
+#include "max6643_fan_controller.h"
```
copperium, line 3: Nitpick: this include should be at the start and separated by a blank line.

copperium (changes requested):
Nice! I suppose this is a pretty simple driver, I still think it's valuable because it's abstracting over the active-low stuff going on with the fanfail and overtemperature pins.

libraries/ms-drivers/src/max6643_fan_controller.c
```diff
@@ -0,0 +1,28 @@
...
+#include "gpio_it.h"
+#include "interrupt.h"
+
+// Callbacks being set here
+StatusCode max6643_init(Max6643Settings *settings) {
+  if (settings == NULL) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  InterruptSettings s_interrupt_settings = {
```
copperium, line 16: This should just be called `interrupt_settings` since it isn't static anymore

libraries/ms-drivers/test/test_max6643_fan_controller.c
```diff
@@ -0,0 +1,83 @@
...
+static uint8_t s_times_overtemp_callback_called;
+static uint8_t s_fanfail_callback_context;
+static uint8_t s_overtemp_callback_context;
+static const GpioAddress test_fanfail_pin = { .port = GPIO_PORT_A, .pin = 0 };
+static const GpioAddress test_overtemp_pin = { .port = GPIO_PORT_A, .pin = 1 };
+
+static void prv_test_max6643_fanfail_interrupt_callback(const GpioAddress *address, void *context) {
+  LOG_DEBUG("FANFAIL INTERRUPT CALLBACK TRIGGERED\n");
+
+  uint8_t *fanfail_context = (uint8_t *)context;
```
copperium, line 23: Nitpick: you don't actually need the cast since it's a void pointer (also applies to the same line in the next function)

copperium (changes requested):
Just one thing and a nitpick, sorry

copperium (approved):

PR #238, opened 2020-11-29
smoketest for drv120

nourishnew:


projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
+#include <stdbool.h>
+
+#include "delay.h"
```
copperium, line 3: You don't use delay.h, so you don't need to include it

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+
+// Toggle the following to select which data is logged
+#define ENABLE_PIN \
+  { .port = GPIO_PORT_A, .pin = 8 }
+#define STATUS_PIN \
+  { .port = GPIO_PORT_A, .pin = 6 }
+#define TIME_INTERVAL 500
+
+static void prv_error_handler(void *context) {
+  LOG_DEBUG("Error %s\n", "error occured");
```
copperium, line 20: `Error error occured` isn't a very descriptive error message. With the DRV120 driver, this error handling callback will be called if the DRV120 signals undervoltage lockout or thermal shutdown, so I think that this would be a good message: `Warning: DRV120 signalled undervoltage lockout or thermal shutdown`. Also, it would be nice to use `LOG_WARN` instead of `LOG_DEBUG` here since it's a warning.

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+    *toggle = false;
+
+  } else {
+    drv120_relay_close();
+    *toggle = true;
+  }
+
+  soft_timer_start_millis(TIME_INTERVAL, prv_timer_callback, toggle, NULL);
+}
+int main(void) {
```
copperium, line 37: Nitpick: please put a blank line between these functions.

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+  gpio_it_init();
+  gpio_init();
+  GpioAddress status_pin_address = STATUS_PIN;
+  GpioAddress enable_pin_address = ENABLE_PIN;
+  bool toggle = true;
+
+  Drv120RelaySettings relay_settings = { .enable_pin = &enable_pin_address,
+                                         .status_pin = &status_pin_address,
+                                         .error_handler = prv_error_handler,
+                                         NULL };
```
copperium, line 49: The style we prefer would be:
```c
Drv120RelaySettings relay_settings = {
  .enable_pin = &enable_pin_address,
  .status_pin = &status_pin_address,
  .error_handler = prv_error_handler,
  .context = NULL,
};
```
If `make format` is giving you trouble and putting it on one line, you can add `//` comments at the end of each line. (You can also omit the `context` since it's just null and you aren't using it.)

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+  LOG_DEBUG("Error %s\n", "error occured");
+}
+
+static void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  bool *toggle = context;
+
+  if (*toggle) {
+    drv120_relay_open();
+    *toggle = false;
+
```
copperium, line 29: Nitpick: this blank line looks kind of awkward, please remove it

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+#include "delay.h"
+#include "drv120_relay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// Toggle the following to select which data is logged
```
copperium, line 12: The defines here aren't "selecting which data is logged", they're configuring the enable and status pins of the DRV120 IC, so your comment should say something about that instead.

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+#define ENABLE_PIN \
+  { .port = GPIO_PORT_A, .pin = 8 }
+#define STATUS_PIN \
+  { .port = GPIO_PORT_A, .pin = 6 }
+#define TIME_INTERVAL 500
```
copperium, line 17: 500ms is *really* fast if we want to observe anything, I'd set it to 1000ms or 2000ms.

projects/smoke_drv120/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+
+#include "delay.h"
+#include "drv120_relay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
```
copperium, line 11: You should have a comment here telling the hardware person running the smoke test what it does, i.e. it periodically opens and closes the relay and logs any error conditions that the DRV120 signals on the status pin.

copperium (changes requested):
Nice job, a few changes

copperium:
Also FYI: our standard for branch names is `soft_XYZ_description`, so in your case it would be `soft_382_smoketest_drv120`.

nourishnew:
@copperium Check it out

copperium (approved):

PR #239, opened 2020-11-30
soft_363 memory usage data formatting

s524kim:


platform/stm32f0xx/scripts/pretty_size.sh
```diff
@@ -0,0 +1,23 @@
...
+
+read -ra temp <<< "$(echo -e "$size_data_raw" | sed -n '2p')"
+
+data="${temp[1]}"
+bss="${temp[2]}"
+
+bytes_used=$(("$bss + $data"))
+bytes_total=16384 #16kb available
+bytes_remaining=$(("$bytes_total - $bytes_used"))
+bytes_threshold=2048 #need 2kb static memory left
```
copperium, line 14: I think a space after the `#` is standard, please add that in these comments.

platform/stm32f0xx/scripts/pretty_size.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Power distribution used ${bytes_used}/${bytes_total} bytes of static memory," \
+     "leaving $bytes_remaining bytes for stack space."
```
copperium, line 17: Power distribution is just one project of many - instead of always saying "Power distribution", you should get the project from the arm-none-eabi-size output as the name of the .elf file without the .elf suffix. For example, if you build `smoke_i2c` with `make build PROJECT=smoke_i2c`, it should say `smoke_i2c used 184/16384 bytes of static memory...`

copperium (changes requested):
Awesome job Patrick, and amazing stuff with the `sed` magic! Just one thing to change + a nitpick.

copperium (approved):

PR #240, opened 2020-12-02
Soft 372 i2c_write firmware implementation

CheranMahalingam:


projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,247 @@
...
+  TEST_ASSERT_EQUAL(14, tx_len_test);
+  for (uint8_t i = 0; i < 7; i++) {
+    TEST_ASSERT_EQUAL(data_1[i + 1], tx_data_test[i]);
+  }
+  for (uint8_t j = 0; j < 7; j++) {
+    TEST_ASSERT_EQUAL(data_2[j + 1], tx_data_test[j + 7]);
+  }
+
+  // Test receiving multiple messages and writing over i2c without register
+  uint8_t command_2[8] = { BABYDRIVER_MESSAGE_I2C_WRITE_COMMAND, I2C_PORT_2, 0x55, 14, 0, 0, 0, 0 };
```
copperium, line 131: Same comment applies to this command message.

projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,247 @@
...
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+
+  TEST_ASSERT_EQUAL(2, s_times_callback_called);
+  TEST_ASSERT_EQUAL(BABYDRIVER_MESSAGE_STATUS, s_received_data[0]);
+  TEST_ASSERT_EQUAL(s_status_return, s_received_data[1]);
+
+  TEST_ASSERT_EQUAL(I2C_PORT_2, i2c_test);
+  TEST_ASSERT_EQUAL(0x55, addr_test);
+  TEST_ASSERT_EQUAL(0, reg_test);
```
copperium, line 157: This test will actually pass even if your code calls `i2c_write_reg` when `is_reg` is 0, since you set `reg_test = 0` in your `i2c_write` mock and you passed 0 in the `reg` field in the command_2 message here. You need another way to differentiate - maybe pass a nonzero `reg` in command_2.

projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,247 @@
...
+  TEST_ASSERT_EQUAL(I2C_PORT_1, i2c_test);
+  TEST_ASSERT_EQUAL(0x74, addr_test);
+  TEST_ASSERT_EQUAL(3, reg_test);
+  TEST_ASSERT_EQUAL(14, tx_len_test);
+  for (uint8_t i = 0; i < 7; i++) {
+    TEST_ASSERT_EQUAL(data_1[i + 1], tx_data_test[i]);
+  }
+  for (uint8_t j = 0; j < 7; j++) {
+    TEST_ASSERT_EQUAL(data_2[j + 1], tx_data_test[j + 7]);
+  }
```
copperium, line 128: Pro tip: you can instead use `TEST_ASSERT_EQUAL_UINT8_ARRAY(expected, actual num_elements)`, like `TEST_ASSERT_EQUAL_UINT8_ARRAY(data_1 + 1, tx_data_test, 7)`.

projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// Test graceful failure when invalid arguments are sent
+void test_invalid_args_i2c(void) {
```
copperium, line 168: I like the invalid cases!

projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,247 @@
...
+  CAN_TRANSMIT_BABYDRIVER(command[0], command[1], command[2], command[3], command[4], command[5],
+                          command[6], command[7]);
+
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+
+  TEST_ASSERT_EQUAL(1, s_times_callback_called);
+  TEST_ASSERT_EQUAL(BABYDRIVER_MESSAGE_STATUS, s_received_data[0]);
+  TEST_ASSERT_EQUAL(s_status_return, s_received_data[1]);
+}
```
copperium, line 247: Nice test for the timeout condition! So that this test doesn't take 750ms to run, consider making the timeout configurable via passing a `uint32_t` into `i2c_write_init` and storing it in a static variable. Then you can pass something small in the tests so it doesn't take a long time to run, and pass the normal timeout in main.c.

copperium, line 247: Edge cases I'd like to see covered (possibly some of them in the same / existing test cases):
* a case with `tx_len < 7`
* a case with `tx_len == 7`
* a case with `tx_len` greater than 7 but not divisible by 7
* a case with `tx_len == 255` (use loops to transmit your data messages)

copperium (changes requested):
Nice job, a bunch of things to consider.

projects/baby_driver/test/test_i2c_write.c
```diff
@@ -0,0 +1,387 @@
...
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+  clock_t end_time = clock();
+  // Stores time taken to receive a status message in milliseconds and truncates the number to 0
+  // decimals
+  uint32_t time_taken = 1000 * ((end_time - start_time) / (double)CLOCKS_PER_SEC);
+
+  TEST_ASSERT_EQUAL(1, s_times_callback_called);
+  TEST_ASSERT_EQUAL(STATUS_CODE_TIMEOUT, s_received_status);
+  // It is expected that it will take 10ms +/- 10ms for status to be returned
+  TEST_ASSERT_INT32_WITHIN(10, 10, time_taken);
```
copperium, line 386: This timing-checking stuff seems like it's liable to fail inconsistently, I think it's fine to just check that the timeout was triggered after the timeout period + a bit.

copperium (changes requested):
Just a couple more things. I notice the tests fail really inconsistently, that's mostly the CAN library's fault and I'll probably get the bugfix for it merged over the break.

copperium (approved):

PR #244, opened 2020-12-06
[Soft 344] smoke bts7040

sudhish3120:


copperium:
Travis seems to be failing because you've got merge conflict markers (`>>>>>> HEAD` etc) in some babydriver files. I'm not sure how they got there but please remove them to get Travis to pass.

projects/smoke_bts7040/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "mux.h"
+#include "pca9539r_gpio_expander.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// experimentally determined value: scaling factor of 670/1160 ~ 0.578 ohms, bias of -8 mA
+#define SMOKE_BTS7040_SENSE_RESISTOR 1160
+#define SMOKE_BTS7040_BIAS (-8)
```
copperium, line 20: These values were experimentally determined for the BTS7200s, but the BTS7040s are different and we haven't done any experimentation with them. In the future we'll use this smoke test to calibrate what these values should be. I'd say that to match the power distribution project's values (found in current_measurement.h), set the resistor to 1210 and the bias to 0, and change the comment to say that these parameters need to be calibrated.

projects/smoke_bts7040/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+#define SMOKE_BTS7040_MIN_FAULT_VOLTAGE_MV 3200
+#define SMOKE_BTS7040_MAX_FAULT_VOLTAGE_MV 10000
+
+// Smoke test settings. Can be modified to fit testing purpose.
+#define CURRENT_MEASURE_INTERVAL_MS 500  // Set wait time between each set of readings
+#define IS_FRONT_POWER_DISTRO true       // Set whether to test FRONT or REAR power distro
+
+// Maximum number of channels that can be tested,
+// which is the same number of bts7200 used in front/rear power distribution
+#define MAX_TEST_CHANNELS 8
```
copperium, line 31: There's only 6 BTS7040s though, eh? Since there's only 6 entries in each `bts7040s` array in the hw configs, if you try to index past 5 with a channel of 6 or 7, you'll get a segfault or other undefined behaviour. So this should be 6. (Also change the comment from bts7200 to bts7040.)

projects/smoke_bts7040/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+
+#define I2C_PORT I2C_PORT_2
+#define PIN_I2C_SCL CONTROLLER_BOARD_ADDR_I2C2_SCL
+#define PIN_I2C_SDA CONTROLLER_BOARD_ADDR_I2C2_SDA
+
+// Set of channels to be tested, the array contains the indices of the BTS7200 in the front/rear HW
+// config Details at
+// https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1419740028/BTS7200+smoke+test+user+guide
+static uint8_t s_test_channels[] = { 0, 1, 2, 3, 4, 5, 6, 7 };
```
copperium, line 40: * the array should not contain 6 or 7 because there are only 6 channels (indices in the bts7040s arrays)
* change `BTS7200` to `BTS7040` in the comment
* the linked Confluence page is for the BTS7200 smoke test and says which channel (index) corresponds to each bts7200. It might be a good idea to make an analogous page for this smoke test and link it instead, but the incorrect link should definitely be removed to avoid confusion.

copperium (changes requested):
Great job, just a couple bts7200 leftovers to change

copperium (approved):

PR #245, opened 2020-12-07
Soft 385 babydriver can sending

selinahsu:
Babydriver CAN sending functions: not totally sure about the testing. Right now I'm just checking fail conditions for some exceptions I wrote in.


projects/baby_driver/scripts/can_send.py
```diff
@@ -0,0 +1,59 @@
...
+        data: a dictionary of field names (strings) to values (integers)
+    """
+    try:
+        msg_obj = DBC_DATABASE.get_message_by_name(msg_name.upper())
+        msg_obj = msg_obj.encode(data)
+    except: # pylint: disable=raise-missing-from
+        raise Exception("CAN message could not be encoded with dbc data.")
+
+    bus = can_util.get_bus(channel)
+    can_msg = can_util.Message(arbitration_id=msg_obj.frame_id, data=msg_obj)
```
copperium, line 58: You overwrote the `msg_obj` object with the byte data from `encode`, so `msg_obj.frame_id` will throw an exception! You need a different name for the `encode`'d data.

projects/baby_driver/scripts/can_send.py
```diff
@@ -0,0 +1,59 @@
...
+    Args:
+        msg_name: the string name of the message
+        channel: CAN channel
+        data: a dictionary of field names (strings) to values (integers)
+    """
+    try:
+        msg_obj = DBC_DATABASE.get_message_by_name(msg_name.upper())
+        msg_obj = msg_obj.encode(data)
+    except: # pylint: disable=raise-missing-from
+        raise Exception("CAN message could not be encoded with dbc data.")
```
copperium, line 55: Again `get_message_by_name` and `encode` will raise exceptions so you don't need this try/except, but if you want to raise a custom exception (and get rid of the pylint disable) use `raise ... from e`

projects/baby_driver/scripts/test_can_send.py
```diff
@@ -0,0 +1,37 @@
...
+from can_send import can_send_raw, load_dbc, can_send
+
+
+class TestCanSendRaw(unittest.TestCase):
+    """Tests functions in Babydriver's can_send module"""
+
+    @patch('can_util.send_message')
+    def can_send_raw(self, mock_send_message):
+        """Tests that can_send_raw raises an Exception if CAN msg fails to send"""
+        mock_send_message.return_value = can.CanError
```
copperium, line 19: This would just return the class `can.CanError` from `send_message` (which you can do: classes are first-class objects in Python and even have their own classes, "metaclasses"!) - what you can do is set `side_effect` to `can.CanError` and that'll throw CanError.

projects/baby_driver/scripts/test_can_send.py
```diff
@@ -0,0 +1,37 @@
...
+        mock_msg_id = 0
+        mock_data = ["item1", "item2", "item3"]
```
copperium, line 21: These should be ints right? to be fully correct

projects/baby_driver/scripts/test_can_send.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_load_dbc(self, mock_cantools_load_file):
+        """Tests that load_dbc raises an Exception if no file is found"""
+        mock_cantools_load_file.return_value = None
+        self.assertRaises(Exception, load_dbc, "./some-file-path")
+
+    @patch('cantools.database.can.Message.encode')
+    def test_can_send(self, mock_encode):
+        """Tests that can_send raises an Exception if no msg_obj is created"""
+        mock_encode.return_value = None
+        self.assertRaises(Exception, can_send, "ex-msg-name")
```
copperium, line 34: The real `cantools.database.load_file` and `encode` will never return None, they'll raise exceptions, so you should change this to check that.

projects/baby_driver/scripts/test_can_send.py
```diff
@@ -0,0 +1,37 @@
...
+import can_util
+import cantools
+from can_send import can_send_raw, load_dbc, can_send
+
+
+class TestCanSendRaw(unittest.TestCase):
+    """Tests functions in Babydriver's can_send module"""
+
+    @patch('can_util.send_message')
+    def can_send_raw(self, mock_send_message):
```
copperium, line 17: This needs to start with `test_` for unittest to recognize it as a test method.

projects/baby_driver/scripts/test_can_send.py
copperium: Also, unless you mock the `get_message_by_name` function, `can_send` will throw an `AttributeError` since `DBC_DATABASE` is None before it gets to the `encode` function.

projects/baby_driver/scripts/test_can_send.py
```diff
@@ -0,0 +1,37 @@
...
+import unittest
+from unittest.mock import patch
+
+import can
+import can_util
+import cantools
+from can_send import can_send_raw, load_dbc, can_send
+
+
+class TestCanSendRaw(unittest.TestCase):
```
copperium, line 13: Ideas for testing your functions:
* mock `can_util.send_message`, call `can_send_raw`, and just check that the parameters passed to `send_message` are what you expect
* mock `encode`, `get_message_by_name`, and `can_util.send_message`, call `can_send`, and check that `send_message` gets the parameters you expect - that'll check that all the data flow is working and avoid errors like the `msg_obj` thing above

copperium (changes requested):

projects/baby_driver/scripts/repl_setup.py
```diff
@@ -8,12 +8,12 @@
 from gpio_get import gpio_get
 from gpio_set import gpio_set
 from adc_read import adc_read
+from can_send import can_send_raw, load_dbc, can_send
 
 def setup_default_channel():
     """Set up the default CAN channel."""
 
     # We import these here so that they aren't imported for the REPL
-    # pylint: disable=import-outside-toplevel
```
copperium, line 16: This comment is necessary in current pylint / the one in CI, if it's giving you an error on your local machine do `pip install pylint` to update it and try again.

copperium (changes requested):
Just the one pylint thing @CheranMahalingam, it'll get caught by CI anyways when I merge master.

copperium (approved):

PR #246, opened 2020-12-13
Soft 384 x86 wait

Max-MZ:
Introduces `sigsuspend()` to `wait()` on x86, allowing for suspension of a thread until delivery of a signal.


libraries/ms-common/test/test_wait.c
```diff
@@ -0,0 +1,39 @@
+#include "wait.h"
```
copperium, line 1: There is one case that will break with this change: RXing CAN messages triggers an interrupt on stm32, but not on x86, so x86 `wait()` won't wake up when a CAN message is RX'd. So, we need to change x86/can_hw.c so that it triggers an x86 interrupt when the RX thread RXes a CAN message.

We *could* do this by registering a dummy interrupt handler with `x86_interrupt_register_handler` in `can_hw_init`, then triggering it with `x86_interrupt_trigger` in the RX thread when it RXes a CAN message. You could pass `NULL` for the handler if you use `INTERRUPT_TYPE_EVENT` interrupts, since event-type interrupts don't trigger the interrupt handler.

However, I think it might be cleaner to add a `x86_interrupt_wake` function in the `x86_interrupt` module which just uses `sigqueue` to send a high-priority interrupt to a reserved interrupt ID. The function would do roughly the same thing as `x86_interrupt_trigger`, just take no arguments, use `INTERRUPT_PRIORITY_HIGH` always, and for the interrupt ID use say `NUM_X86_INTERRUPT_INTERRUPTS`, which is ignored by the signal handler, so no handler will be triggered. Then when you have that function you can call it from within the CAN HW RX thread, and `wait()` should wake up.

You should have a test for this condition in `test_wait.c`, but to test quickly you can check that babydriver works: if you do `make babydriver PLATFORM=x86`, then type something like `gpio_set('a', 2, True)` at the REPL that comes up, it should work if `wait()` wakes up on CAN RXes and throw a `TimeoutError` if `wait()` doesn't.

copperium (changes requested):
Awesome stuff. Sorry for the late review, exams got in the way. Most of the comments are trivial, only the last couple are beefier.

copperium (changes requested):
This is amazing. Most of the comments are minor nitpicks, it's really only the strange stuff in `test_blink_event_generator.c` that's blocking.

copperium:
The only blocker is the comment about prefixing a bunch of stuff with `s_`, otherwise LGTM

copperium (approved):

PR #247, opened 2020-12-19
smoketest for max6643 fan controller

nourishnew:


projects/smoketest_max6643/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  gpio_it_init();
+  gpio_init();
+  GpioAddress fanfail_pin_address = FANFAIL_PIN;
+  GpioAddress overtemp_pin_address = OVERTEMP_PIN;
+  GpioAddress fan_pin_address = FAN_PIN;
+  bool fan_speed_toggle = FAN_SPEED_HIGH;
```
copperium, line 32: You don't need to assign the #define'd macros to variables before using them, you can just use them directly.

projects/smoketest_max6643/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+    .direction = GPIO_DIR_OUT,  //
+    .state = GPIO_STATE_LOW,    //
+  };
+  if (fan_speed_toggle) {
+    gpio_init_pin(&fan_pin_address, &gpio_high_settings);
+    gpio_set_state(&fan_pin_address, GPIO_STATE_HIGH);
+  } else {
+    gpio_init_pin(&fan_pin_address, &gpio_low_settings);
+    gpio_set_state(&fan_pin_address, GPIO_STATE_LOW);
+  }
```
copperium, line 47: `gpio_init_pin` already sets the state of the pin as set by the `state` field of `GpioSettings`, so you don't need the `gpio_set_state` calls.

projects/smoketest_max6643/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+#include "max6643_fan_controller.h"
+#include "wait.h"
+
+#define FANFAIL_PIN \
+  { .port = GPIO_PORT_B, .pin = 4 }
+#define OVERTEMP_PIN \
+  { .port = GPIO_PORT_B, .pin = 5 }
+#define FAN_PIN \
+  { .port = GPIO_PORT_B, .pin = 3 }
+#define FAN_SPEED_HIGH true
```
copperium, line 15: This controls full speed, not "high speed", so `FAN_FULL_SPEED` might be a better name.

projects/smoketest_max6643/src/main.c
```diff
@@ -0,0 +1,58 @@
+#include <stdbool.h>
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "max6643_fan_controller.h"
+#include "wait.h"
+
```
copperium, line 8: Generally we put a comment at the start of smoke tests explaining how to use it and what each of the configurable items (i.e. the defines) mean. This helps the hardware person using the smoke test to validate the board to understand how to use the smoke test.

copperium (changes requested):

projects/smoketest_max6643/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+int main(void) {
+  interrupt_init();
+  gpio_it_init();
+  gpio_init();
+  GpioAddress fanfail_pin_address = { .port = GPIO_PORT_B, .pin = 4 };
+  GpioAddress overtemp_pin_address = { .port = GPIO_PORT_B, .pin = 5 };
+  GpioAddress fan_pin_address = { .port = GPIO_PORT_B, .pin = 3 };
+  bool fan_speed_toggle = FAN_FULL_SPEED;
```
copperium, line 31: Sorry, I meant you don't have to use these intermediary variables between the #defines and the Max6643Settings. You still need the defines, you just don't need to assign them to variables, so the settings struct would have lines like `.overtemp_pin  = OVERTEMP_PIN`.

copperium (changes requested):
One more thing, a bit of a miscommunication.

projects/smoketest_max6643/src/main.c
nourishnew: done

copperium (approved):

PR #249, opened 2020-12-21
Soft 379 spi exchange python

DoctaRay:


projects/baby_driver/scripts/test_spi_exchange.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Test rx_len > 7
+        # pylint: disable=line-too-long
+        mock_next_message.side_effect = (Message(data=data_msg), Message(data=data_msg), Message(data=status_msg))
+        self.assertEqual(spi_exchange(
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1], 8
+        ), [0, 0, 0, 0, 0, 0, 0, 0])
+
+        # Test multiple messages, low rx_len
+        # pylint: disable=line-too-long
```
copperium, line 99: If the lines are too long, then wrap the line rather than just disabling the message!

projects/baby_driver/scripts/test_spi_exchange.py
```diff
@@ -0,0 +1,141 @@
...
+from message_defs import BABYDRIVER_DEVICE_ID, BABYDRIVER_CAN_MESSAGE_ID, BabydriverMessageId
+
+OK_STATUS = 0
+FAILING_STATUS = 1
+class TestSPIExchange(unittest.TestCase):
+    """Test spi_exchange function"""
+
+    @patch('can_util.send_message')
+    @patch('can_util.next_message')
+    def test_send_message(self, mock_next_message, mock_send_message):
```
copperium, line 18: Your tests are fairly comprehensive, I would also like to see tests with `len(tx_bytes) == 255` and with `rx_len == 255` since they're the maximum values. You can use Python's multiplying lists by ints feature to help with constructing the necessary lists `[0] * 5 == [0, 0, 0, 0, 0]`.

projects/baby_driver/scripts/test_spi_exchange.py
```diff
@@ -0,0 +1,141 @@
...
+        # pylint: disable=line-too-long
+        mock_next_message.side_effect = (Message(data=data_msg), Message(data=data_msg), Message(data=status_msg))
+        self.assertEqual(spi_exchange(
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1], 3
+        ), [0, 0, 0])
+
+    @patch('can_util.send_message')
+    @patch('can_util.next_message')
+    # pylint: disable=unused-argument
+    def test_fail_conditions(self, mock_next_message, mock_send_message):
```
copperium, line 108: You should also test what happens when `len(tx_bytes) == 256` and when `rx_len == 256` since that's one past the maximum length.

projects/baby_driver/scripts/test_spi_exchange.py
```diff
@@ -0,0 +1,141 @@
...
+from unittest.mock import patch
+
+from spi_exchange import spi_exchange
+from gpio_port import GpioPort
+from can_util import Message
+from message_defs import BABYDRIVER_DEVICE_ID, BABYDRIVER_CAN_MESSAGE_ID, BabydriverMessageId
+
+OK_STATUS = 0
+FAILING_STATUS = 1
+class TestSPIExchange(unittest.TestCase):
```
copperium, line 13: Nitpick: please put a blank line before the class declaration. I like the named constants instead of just using 0 and 1! Also, usually the recommendation is to treat acronyms like SPI as words in CamelCase: `TestSpiExchange`.

copperium (changes requested):
Awesome stuff, a few things to clean up.

projects/baby_driver/scripts/spi_exchange.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    if spi_port not in (1, 2):
+        raise ValueError("ERROR: Expected SPI port 1 or 2")
+    if spi_mode not in (0, 1, 2, 3):
+        raise ValueError("ERROR: Expected mode between 0 and 3")
+    if not 0 <= len(tx_bytes) <= 256:
+        raise ValueError("ERROR: numbers of tx_bytes must be between 0 and 256")
+    if not 0 <= rx_len <= 256:
+        raise ValueError("ERROR: rx_len must be a non-negative integer between 0 and 256")
```
copperium, line 44: The maximum value is actually 255 not 256 since 256 won't fit in a uint8. So the conditions and error messages should exclude 256. The tests are passing because a value >= 256 causes a ValueError later in can_pack.

Also, it's kind of ambiguous whether the `not` applies to the 0 or the whole `0 <= len(tx_bytes) <= 256`, please put brackets around the `0 <= len(tx_bytes) <= 256` and similarly with `rx_len`.

copperium (changes requested):
Just one more thing, sorry

copperium (approved):

PR #250, opened 2020-12-29
SOFT-390 pcf8523 RTC

AveryChiu64:
Driver for the pcf8523 + binary-coded decimal convertor

### DIFF
--- a/libraries/ms-common/inc/bcd.h
+++ b/libraries/ms-common/inc/bcd.h
@@ -1,0 +1,9 @@
+#pragma once
+
+#include <stdint.h>
+// Binary coded decimal converter for two-digit numbers
+// A value of 0 will be returned if an invalid value is passed in
+
+uint8_t dec_to_bcd(uint8_t dec_value);
+
+uint8_t bcd_to_dec(uint8_t bcd_value);
--- a/libraries/ms-common/src/bcd.c
+++ b/libraries/ms-common/src/bcd.c
@@ -1,0 +1,16 @@
+#include "bcd.h"
+
+uint8_t dec_to_bcd(uint8_t dec_value) {
+  if (dec_value > 99) {
+    return 0;
+  }
+  uint8_t tens = dec_value / 10;
+  uint8_t ones = dec_value % 10;
+  return (tens << 4) + ones;
+}
+
+uint8_t bcd_to_dec(uint8_t bcd_value) {
+  uint8_t tens = (bcd_value & 0b11110000) >> 4;
+  uint8_t ones = bcd_value & 0b00001111;
+  return (tens > 9 || ones > 9) ? 0 : (tens * 10 + ones);
+}
--- a/libraries/ms-drivers/inc/pcf8523_rtc_defs.h
+++ b/libraries/ms-drivers/inc/pcf8523_rtc_defs.h
@@ -1,0 +1,75 @@
+#pragma once
+#include <stdint.h>
+
+#define CR_BASE_ADDR 0x00
+#define TIME_BASE_ADDR 0x03
+#define ALARM_BASE_ADDR 0X0A
+
+#define CR ((CrRegDef *)CR_BASE_ADDR)
+#define TIME ((TimeRegDef *)TIME_BASE_ADDR)
+#define ALARM ((AlarmRegDef *)ALARM_BASE_ADDR)
+
+typedef struct {
+  volatile uint8_t CR1;
+  volatile uint8_t CR2;
+  volatile uint8_t CR3;
+} CrRegDef;
+
+typedef struct {
+  volatile uint8_t SECONDS;
+  volatile uint8_t MINUTES;
+  volatile uint8_t HOURS;
+  volatile uint8_t DAYS;
+  volatile uint8_t WEEKDAYS;
+  volatile uint8_t MONTHS;
+  volatile uint8_t YEARS;
+} TimeRegDef;
+
+typedef struct {
+  volatile uint8_t MINUTE_ALARM;
+  volatile uint8_t HOUR_ALARM;
+  volatile uint8_t DAY_ALARM;
+  volatile uint8_t WEEKDAY_ALARM;
+  volatile uint8_t OFFSET;
+} AlarmRegDef;
+
+typedef enum {
+  CIE = 0,
+  AIE,
+  SIE,
+  TIME_12_24,
+  SR,
+  STOP,
+  T,
+  CAP_SEL,
+} Control1;
+
+typedef enum {
+  CTBIE = 0,
+  CTAIE,
+  WTAIE,
+  AF,
+  SF,
+  CTBF,
+  CTAF,
+  WTAF,
+} Control2;
+
+typedef enum {
+  CTBIE = 0,
+  CTAIE,
+  WTAIE,
+  AF,
+  SF,
+  CTBF,
+  CTAF,
+  WTAF,
+} Control2;
+
+typedef enum {
+  BLIE = 0,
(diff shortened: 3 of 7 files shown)


libraries/ms-drivers/src/pcf8523_rtc.c
```diff
@@ -0,0 +1,79 @@
...
+  // Set starting register address (this will auto-increment)
+  uint8_t starting_reg = SECONDS;
+  i2c_write(port, I2C_ADDR, &starting_reg, 1);
+
+  // Read time registers
+  uint8_t data[NUM_TIME_REG];
+  i2c_read(port, I2C_ADDR, data, (sizeof(data)));
+
+  // Store time
+  time->seconds = bcd_to_dec(data[0]);
```
copperium, line 41: Bit 7 of the seconds register is `OS` for "oscillator stop" - it's 1 if "clock integrity is not guaranteed; oscillator has stopped or been interrupted". To avoid this condition showing up as weird numbers of seconds, you might want to handle the OS bit separately with a `LOG_WARN` or something, then set it to 0 for the seconds bcd conversion.

libraries/ms-drivers/test/test_pcf8523_rtc.c
```diff
@@ -0,0 +1,76 @@
...
+#define TEST_CONFIG_PIN_I2C_SCL \
+  { GPIO_PORT_B, 10 }
+#define TEST_CONFIG_PIN_I2C_SDA \
+  { GPIO_PORT_B, 11 }
+
+I2CSettings i2c = {
+  .speed = I2C_SPEED_FAST,         //
+  .sda = TEST_CONFIG_PIN_I2C_SDA,  //
+  .scl = TEST_CONFIG_PIN_I2C_SCL,  //
+};
```
copperium, line 20: This should be static+prefixed with `s_`, or just a local variable in `setup_test`.

libraries/ms-drivers/test/test_pcf8523_rtc.c
```diff
@@ -0,0 +1,76 @@
...
+void setup_test(void) {
+  gpio_init();
+  i2c_init(I2C_PORT_1, &i2c);
+}
+
+void teardown_test(void) {}
+
+void test_pcf8523_init_valid() {
+  TEST_ASSERT_OK(pcf8523_init(I2C_PORT_1));
+}
```
copperium, line 31: The I2C port should be `TEST_I2C_PORT` eh?

libraries/ms-drivers/test/test_pcf8523_rtc.c
```diff
@@ -0,0 +1,76 @@
...
+    .seconds = 100,
+    .minutes = 100,
+    .hours = 100,
+    .days = 100,
+    .weekdays = 100,
+    .months = 100,
+    .years = 100,
+  };
+  TEST_ASSERT_NOT_OK(pcf8523_set_time(&time_overflow));
+}
```
copperium, line 76: You need to initialize pcf8523 at the start of these tests eh?

libraries/ms-drivers/test/test_pcf8523_rtc.c
```diff
@@ -0,0 +1,76 @@
...
+  Pcf8523Time time_get = { 0 };
+  TEST_ASSERT_OK(pcf8523_set_time(&time_set));
+  TEST_ASSERT_OK(pcf8523_get_time(&time_get));
+  // Note: We don't check seconds since the timer will start as soon as we've finished setting it
+  TEST_ASSERT_TRUE((time_get.minutes == 1) && (time_get.hours == 0) && (time_get.days == 1) &&
+                   (time_get.weekdays == 0) && (time_get.months == 1) && (time_get.years == 0));
+}
+
+// Test setting time to 0
+void test_invalid_set_time() {
```
copperium, line 62: These argument lists should be `(void)`

libraries/ms-drivers/inc/pcf8523_rtc_defs.h
```diff
@@ -0,0 +1,62 @@
...
+  WTAF,
+} Control2;
+
+typedef enum {
+  BLIE = 0,
+  BSIE,
+  BLF,
+  BSF,
+  PM,
+} Control3;
```
copperium, line 62: You might want to add prefixes on some of these names - `T` is a bit of an ambiguous identifier but `PCF8523_CONTROL_1_T` is less ambiguous. Also you're missing `NUM_` members at the end of the control enums.

copperium (changes requested):

libraries/ms-common/inc/bcd.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <stdint.h>
+// Binary coded decimal converter for two-digit numbers
+// A value of 0xFF will be returned if an invalid value is passed in
+
+#define INVALID 0xFF
```
copperium, line 7: You should call this something BCD-specific like `INVALID_BCD_ARGUMENT` since it'll pollute the global namespace of anything that includes it, which might want their own `INVALID` macros.

libraries/ms-common/src/bcd.c
```diff
@@ -0,0 +1,16 @@
...
+  }
+  uint8_t tens = dec_value / 10;
+  uint8_t ones = dec_value % 10;
+  return (tens << 4) | ones;
+}
+
+uint8_t bcd_to_dec(uint8_t bcd_value) {
+  uint8_t tens = (bcd_value & 0b11110000) >> 4;
+  uint8_t ones = bcd_value & 0b00001111;
+  return (tens > 9 || ones > 9) ? 0XFF : (tens * 10 + ones);
```
copperium, line 15: I've never seen an uppercase `X` in `0XFF` before, kinda surprised that it works. `0xFF` is more traditional, but you should use the macro from bcd.h in any case.

libraries/ms-drivers/src/pcf8523_rtc.c
```diff
@@ -0,0 +1,91 @@
...
+  uint8_t starting_reg = SECONDS;
+  i2c_write(s_port, PCF8523_I2C_ADDR, &starting_reg, 1);
+
+  // Read time registers
+  uint8_t data[NUM_TIME_REG];
+  i2c_read(s_port, PCF8523_I2C_ADDR, data, (sizeof(data)));
+
+  // If the 7th bit is enabled if the oscillator has stopped
+  if ((time->seconds) & (1 << 7)) {
+    LOG_WARN("Clock integrity is not guaranteed; oscillator has stopped or been interrupted");
```
copperium, line 48: You'll want a `\n` on the end here

libraries/ms-drivers/src/pcf8523_rtc.c
```diff
@@ -0,0 +1,91 @@
...
+}
+
+StatusCode pcf8523_set_time(Pcf8523Time *time) {
+  if (time->seconds > 59 || time->minutes > 59 || time->hours > 23 || time->days > 31 ||
+      time->days < 1 || time->weekdays > 6 || time->months > 12 || time->months < 1 ||
+      time->years > 99) {
+    return STATUS_CODE_INVALID_ARGS;
+  }
+  // Stop the timer
+  uint8_t stop[2] = { CR1, STOP_CR1_SETTINGS };
```
copperium, line 70: You still need to include the capacitor settings in the stop settings, I think - I'd suggest you OR `STOP_CR1_SETTINGS` with `s_cap << PCF8523_CR1_CAP_SEL`.

copperium (approved):

PR #252, opened 2020-12-31
[SOFT-386] Update centre console button pins to latest rev

copperium:
Tiny change: the button pins are changing for centre console's latest rev. This updates them.


AveryChiu64 (approved):
Looks good. I have attached the schematic here and it looks like it matches.
image

PR #253, opened 2021-01-04
SOFT-183 smoke steering

AveryChiu64:
Smoke test for steering board. Allows you to see the values from the ADC and will log the event that is raised (eg if the horn button is pressed the console should output "horn").


AveryChiu64:
This will need to be verified in person. It will be merged once it can be verified.

jessm (approved):

PR #255, opened 2021-01-10
FW 102 homework: code review request

hzhou0:
Review code?

### DIFF
--- a/projects/hello_world/src/hello_world.c
+++ b/projects/hello_world/src/hello_world.c
@@ -1,0 +1,35 @@
+//
+// Created by henry on 2021-01-09.
+//
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+void callback(SoftTimerId id, void *context) {
+  static bool b_increment = false;
+  Counters *Counter = context;
+  LOG_DEBUG("%d\n", ++Counter->counter_a);
+  if (b_increment) {
+    LOG_DEBUG("%d\n", ++Counter->counter_b);
+    b_increment = false;
+  } else {
+    b_increment = true;
+  }
+}
+int main(void) {
+  Counters Counter = { 0 };
+  interrupt_init();
+  soft_timer_init();
+  while (true) {
+    if(!soft_timer_inuse()){
+      soft_timer_start(0.5 * 1000000, callback, &Counter, NULL);
+    }
+  }
+  return 0;
+}
--- a/projects/hello_world/rules.mk
+++ b/projects/hello_world/rules.mk
@@ -1,0 +1,9 @@
+# Defines $(T)_SRC, $(T)_INC, $(T)_DEPS, and $(T)_CFLAGS for the build makefile.
+# Tests can be excluded by defining $(T)_EXCLUDE_TESTS.
+# Pre-defined:
+# $(T)_SRC_ROOT: $(T)_DIR/src
+# $(T)_INC_DIRS: $(T)_DIR/inc{/$(PLATFORM)}
+# $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
+
+# Specify the libraries you want to include
+$(T)_DEPS := ms-common


projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+int main(void) {
+  Counters Counter = { 0 };
+  interrupt_init();
+  soft_timer_init();
+  while (true) {
+    if(!soft_timer_inuse()){
+      soft_timer_start(0.5 * 1000000, callback, &Counter, NULL);
+    }
```
copperium, line 32: This is definitely the most creative solution to this homework I've seen :) The solution hinted at by the homework is to start one soft timer in main, then at the end of the soft timer callback, start *another* soft timer to form a loop. That way you don't have to constantly poll here, and you can just use a plain old `while (true) {}` and let the soft timers do their thing.

Other tips:
* you can use `soft_timer_start_millis` (see soft_timer.h) to avoid having to multiply by 1000000, you can just specify the duration in milliseconds
* it's good practice to `#define` a constant for the duration of the soft timer and use that instead of the magic numbers here - make sure the constant has a descriptive name and mentions the unit it's in (i.e. ms or us)
* I see Travis CI (our automatic build running CI system) failed because the `if` line isn't formatted correctly - you can use `make format` to autoformat our code in accordance with our style.

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
...
+  LOG_DEBUG("%d\n", ++Counter->counter_a);
+  if (b_increment) {
+    LOG_DEBUG("%d\n", ++Counter->counter_b);
+    b_increment = false;
+  } else {
+    b_increment = true;
+  }
+}
+int main(void) {
+  Counters Counter = { 0 };
```
copperium, line 26: Please only use UpperCamelCase for types (like `Counters`) and not for local variables like `Counter`. This also goes for the `Counter` variable in `callback` above.

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
...
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+void callback(SoftTimerId id, void *context) {
```
copperium, line 14: Since this is a private function, our [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) say that it should be declared `static` and prefixed with `prv_`. The `static` means that it can't be used by other files.

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
...
+  static bool b_increment = false;
```
copperium, line 15: A function-local static variable is a clever way to get around the "no global variables" requirement, but we generally don't use them unless there's a *really good reason*. Instead, consider using whether `counter_a` is even or odd to decide whether to increment `counter_b`.

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
...
+  Counters *Counter = context;
+  LOG_DEBUG("%d\n", ++Counter->counter_a);
+  if (b_increment) {
+    LOG_DEBUG("%d\n", ++Counter->counter_b);
```
copperium, line 19: Using pre-incrementing is clever, but it also makes it a bit harder to read. Consider just incrementing after printing on another line.

Also it's technically not required I suppose, but to match the screenshot in the module, add `Counter A:` and `Counter B:` to the logging.

projects/hello_world/src/hello_world.c
copperium, line 14: Style nit: to make it easier to read, we leave a blank line between top-level elements like struct declarations and function definitions. So there should be a blank line here and between `callback` and `main`.

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,35 @@
+//
+// Created by henry on 2021-01-09.
+//
```
copperium, line 3: This is probably autogenerated junk from your IDE - it's okay this time, but for future code please remove it.

copperium (changes requested):
Hey Henry! Thanks for taking the initiative to get started on the homework. I've gone ahead and invited you to the github organization so that you don't have to fork the repo next time. Come to the FW 101 tutorial tomorrow (schedule on Slack) if you're interested. Make sure to address the changes here before moving on to the FW 103 homework.

PR #256, opened 2021-01-10
[SOFT-405] Fix Babydriver's next_message skipping messages when the C side is too fast

copperium:
An interesting bug that popped up when testing x86 `wait()`: when the Python side expects two or more messages from the C side in sequence, sometimes the time it takes to process the first message and start listening for the second one can actually exceed the time it takes to send the second CAN message from the C side (which can be as low as ~400us)! This causes missed messages and broken transactions.

The fix is to use python-can's `BufferedReader` listener, which puts all messages into a queue, or buffer, so that we don't miss any messages. Because we need to catch all messages, even when `next_message` isn't called, I decided to cache the bus/reader (and the required `Notifier` which makes it work) so that there's only one per channel running all the time.

### DIFF
--- a/projects/baby_driver/scripts/can_util.py
+++ b/projects/baby_driver/scripts/can_util.py
@@ -1,22 +1,54 @@
 """This module implements a low-level wrapper over Python-CAN's CAN API specific to Babydriver."""
 
 import time
+from collections import namedtuple
+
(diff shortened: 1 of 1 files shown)


g-jessmuir (approved):
good change

PR #259, opened 2021-01-19
Soft 395 lockout detection

DoctaRay:


projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
+#include "can_transmit.h"
```
copperium, line 1: You need to include the `front_uv_detector.c` header file too, in the first line, separated by the others by a blank line.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
...
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+
+static GpioAddress s_uv_comp_pin_address = { .port = GPIO_PORT_B, .pin = 0 };
```
copperium, line 6: Instead of using this constant static variable, take (a pointer to) the UV comparator pin as input in `front_uv_detector_init`. You don't even need to store it statically, you can just pass it into `gpio_it_register_interrupt` and be done with it. In main.c, you can then use the `FRONT_UV_COMPARATOR_PIN` macro you defined to get the argument's value.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
...
+
+static InterruptSettings s_interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
```
copperium, line 11: Since this is only used in one function and never changes, I'd prefer you instead make it a local variable in `prv_register_interrupts`. That way it only takes up stack space during the function rather than all the time. (Also remove the `s_` prefix when you make it a local variable.)

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
...
+
+static void prv_pin_interrupt_handler(const GpioAddress *address, void *context) {
+  CAN_TRANSMIT_UV_CUTOFF_NOTIFICATION();
+
+  LOG_WARN("UV lockout was triggered\n");
+}
+
+void prv_register_interrupts(void) {
```
copperium, line 19: Since this is a private function, it should be `static` - otherwise there could be an obscure error if another file had a `prv_register_interrupts`.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
...
+  // callback occurs when pin turns low i.e falling
+  gpio_it_register_interrupt(&s_uv_comp_pin_address, &s_interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             prv_pin_interrupt_handler, NULL);
+}
+
+StatusCode front_uv_detector_init() {
+  interrupt_init();
+  status_ok_or_return(gpio_init());
+  gpio_it_init();
```
copperium, line 28: You don't need to do this here: in a multi-module project like power distribution, important libraries like interrupts, gpio, and gpio_it are only initialized in main.c. Initializing a library twice could cause problems like wiping all GPIO pins' states. This is the reason behind the `Requires ... to be initialized` comments in headers. You can just remove these three lines.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,32 @@
...
+  prv_register_interrupts();
+
+  return STATUS_CODE_OK;
```
copperium, line 31: `gpio_it_register_interrupt` returns a StatusCode, but `prv_register_interrupt` doesn't, so any possible errors from `gpio_it_register_interrupt` get swallowed. Since `front_uv_detector_init` is going to be real short without the initialization lines, I'd recommend you just move the `gpio_it_register_init` line here, `return` it, and remove `prv_register_interrupts` entirely.

libraries/codegen-tooling/inc/can_msg_defs.h
```diff
@@ -62,6 +62,7 @@ typedef enum {
   SYSTEM_CAN_MESSAGE_CRUISE_CONTROL_COMMAND = 41,
   SYSTEM_CAN_MESSAGE_AUX_DCDC_VC = 43,
   SYSTEM_CAN_MESSAGE_DCDC_TEMPS = 44,
+  SYSTEM_CAN_MESSAGE_UV_CUTOFF_NOTIFICATION = 45,
```
copperium, line 65: Also make a PR in codegen-tooling-msxiv, I'll merge it at the same time as this gets merged.

projects/power_distribution/test/test_front_uv_detector.c
```diff
@@ -0,0 +1,55 @@
...
+  event_queue_init();
+  front_uv_detector_init();
+
+  s_interrupt_ran = false;
+}
+
+void teardown_test(void) {}
+
+// test that CAN message gets sent after lockout
+void test_uv_front_detector_notification() {
```
copperium, line 46: This parameter list should be `(void)`

projects/power_distribution/test/test_front_uv_detector.c
```diff
@@ -0,0 +1,55 @@
+#include "can_transmit.h"
+#include "front_uv_detector.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "ms_test_helpers.h"
+#include "pd_events.h"
+
+static GpioAddress s_uv_comp_pin_address = { .port = GPIO_PORT_B, .pin = 0 };
```
copperium, line 8: Use `FRONT_UV_COMPARATOR_PIN` from pin_defs.h instead of hardcoding the value here.

projects/power_distribution/test/test_front_uv_detector.c
```diff
@@ -0,0 +1,55 @@
...
+    .loopback = true,
+    .bitrate = CAN_HW_BITRATE_500KBPS,
+    .rx_event = POWER_DISTRIBUTION_CAN_EVENT_RX,
+    .tx_event = POWER_DISTRIBUTION_CAN_EVENT_TX,
+    .fault_event = POWER_DISTRIBUTION_CAN_EVENT_FAULT,
+    .tx = { GPIO_PORT_A, 12 },
+    .rx = { GPIO_PORT_A, 11 },
+  };
+  can_init(&s_can_storage, &can_settings);
+}
```
copperium, line 26: Since the CAN device is constant (it's always `SYSTEM_CAN_DEVICE_POWER_DISTRIBUTION_FRONT` since this module is always on front PD), you don't need this function, which is used to initialize CAN with a different CAN device per test - you can just call `initialize_can_and_dependencies` from ms_test_helper_can.h in `setup_test`, which will also initialize gpio, interrupts, soft timers, and the event queue for you.

projects/power_distribution/test/test_front_uv_detector.c
```diff
@@ -0,0 +1,55 @@
...
+  s_interrupt_ran = true;
+  return STATUS_CODE_OK;
+}
+
+void setup_test(void) {
+  gpio_init();
+  interrupt_init();
+  soft_timer_init();
+  event_queue_init();
+  front_uv_detector_init();
```
copperium, line 38: Do a `TEST_ASSERT_OK` on this (macro from test_helpers.h) to verify that it returns OK. (Asserts are valid in setup_test and teardown_test by the way.)

projects/power_distribution/test/test_front_uv_detector.c
```diff
@@ -0,0 +1,55 @@
+#include "can_transmit.h"
+#include "front_uv_detector.h"
```
copperium, line 2: Nitpick: front_uv_detector.h should be the first include and should be separated from the others by a blank line since it's the one you're testing.

copperium (changes requested):

projects/power_distribution/src/front_uv_detector.c
DoctaRay: When I take these lines out, the tests stop succeeding (the trigger function in the test doesn't occur and I'm guessing the test is waiting for an RX/TX when no CAN message gets sent). Is there something I'm missing here?

copperium: Ah, you still need to initialize `gpio_it` in your `setup_test`, `initialize_can_and_dependencies` doesn't do that for you.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,30 @@
...
+static GpioAddress s_uv_comp_pin_address = FRONT_UV_COMPARATOR_PIN;
+
+static void prv_pin_interrupt_handler(const GpioAddress *address, void *context) {
+  CAN_TRANSMIT_UV_CUTOFF_NOTIFICATION();
+
+  LOG_WARN("UV lockout was triggered\n");
+}
+
+StatusCode front_uv_detector_init(GpioAddress *detector_pin) {
+  gpio_it_init();
```
copperium, line 18: Initialize this in `setup_test` not here! In power distribution, main.c already initializes `gpio_it`, and initializing it again could interfere with other modules.

projects/power_distribution/src/front_uv_detector.c
```diff
@@ -0,0 +1,30 @@
+#include "front_uv_detector.h"
+
+#include "can_transmit.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "pin_defs.h"
+
+static GpioAddress s_uv_comp_pin_address = FRONT_UV_COMPARATOR_PIN;
```
copperium, line 9: This variable is unused now, please remove it. (And you can also remove the `pin_defs.h` include.)

projects/power_distribution/src/main.c
```diff
@@ -158,6 +159,9 @@ int main(void) {
       .strobe_blink_delay_us = STROBE_BLINK_INTERVAL_US,
     };
     rear_power_distribution_strobe_blinker_init(&strobe_blinker_settings);
+  } else {
+    // initialize UV cutoff detector
+    front_uv_detector_init(&(GpioAddress)FRONT_UV_COMPARATOR_PIN);
   }
```
copperium, line 165: Having this as the `else` clause creates a kind of confusing if statement like this:
```c
if (!s_is_front_power_distribution) {
  // ... initialize strobe blinker for rear
} else {
  // ... initialize UV cutoff detector for front
}
```
It's a bit ugly since the `else` clause is based on a double negative. I'd prefer you refactor it to look like this:
```c
if (s_is_front_power_distribution) {
  // ... initialize UV cutoff detector for front
} else {
  // ... initialize strobe blinker for rear
}
```
That way you get rid of the double negative. When you do, move the comment about the rear strobe blinker to inside the else clause so it stays with that block of code.

copperium (changes requested):
Just a couple of quick things.

copperium:
Also make a PR adding your message to codegen-tooling-msxiv so that it can be merged at the same time.

copperium (approved):

PR #260, opened 2021-01-20
FW102 homework revised

hzhou0:


projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,33 @@
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+#define INTERVAL 0.5
```
copperium, line 7: It's not immediately clear from this what units `INTERVAL` is in - add a suffix like `INTERVAL_S` to tell readers that it's in seconds. (Also, there should be a blank line between the includes and any defines.)

projects/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,33 @@
...
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+void prv_callback(SoftTimerId id, void *context) {
```
copperium, line 13: Nitpick: there should be a blank line between top-level elements like structs and callbacks.

copperium (commented):
Awesome job Henry. I added a couple of comments for your future reference, but I'll mark this as complete and close the PR for now. Feel free to move on to FW 103 (and come to the presentation on Saturday at 2pm!).

PR #261, opened 2021-01-21
Soft 369 add multiple stores

JarvisWeng:


g-jessmuir:
please change PR to merge to branch `mpxe` instead of `master`, makes it much easier to read through rather than trying to sift through 1500 lines of change :')

libraries/mpxe-store/src/store.c
```diff
@@ -14,7 +14,9 @@
 #include "log.h"
 #include "stores.pb-c.h"
 
-#define MAX_STORE_COUNT 64
+#define MAX_STORE_COUNT 257
```
g-jessmuir, line 17: Not super impressed you wanted to go about it this way, what about stores beyond just pca9539r and gpio? not a super great change

mpxe/sims/pca9539r.py
```diff
@@ -6,8 +6,8 @@
 class Pca9539r(sim.Sim):
     def handle_update(self, pm, proj):
         stores = proj.stores
-        if (stores_pb2.MxStoreType.PCA9539R, 0) in stores:
-            pca        = stores[(stores_pb2.MxStoreType.PCA9539R, 0)]
+        if (stores_pb2.MxStoreType.PCA9539R, 116) in stores:    #116 is the address used in the smoke test
```
g-jessmuir, line 9: use hex since it's the normal thing for i2c addresses

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -20,33 +19,38 @@ static I2CPort s_i2c_port = NUM_I2C_PORTS;
...
 
+  int k = (intptr_t)(key);
   for (uint16_t i = 0; i < mask->n_state; i++) {
     // only update state if mask is set
     if (mask->state[i] != 0) {
-      s_store.state[i] = msg->state[i];
-      if (s_pin_settings[PCA9539_I2C_ADDRESS][i].state != (uint8_t)msg->state[i]) {
-        s_pin_settings[PCA9539_I2C_ADDRESS][i].state = msg->state[i];
+      s_stores[k].state[i] = msg->state[i];
+      if (s_pin_settings[k][i].state != (uint)msg->state[i]) {
```
g-jessmuir, line 45: does this actually compile? what's uint? shouldn't' it be uint8_t?

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -58,9 +62,12 @@ static void prv_init_store(void) {
...
     (UpdateStoreFunc)update_store,
   };
-  s_store.n_state = NUM_PCA9539R_GPIO_PINS;
-  s_store.state = malloc(NUM_PCA9539R_GPIO_PINS * sizeof(protobuf_c_boolean));
-  store_register(MX_STORE_TYPE__PCA9539R, funcs, &s_store, NULL);
+  for (int i = 0; i < MAX_I2C_ADDRESSES; ++i) {
+    s_stores[i] = (MxPca9539rStore)MX_PCA9539R_STORE__INIT;
+    s_stores[i].n_state = NUM_PCA9539R_GPIO_PINS;
+    s_stores[i].state = malloc(NUM_PCA9539R_GPIO_PINS * sizeof(protobuf_c_boolean));
+    store_register(MX_STORE_TYPE__PCA9539R, funcs, &s_stores[i], (void *)(intptr_t)i);
```
g-jessmuir, line 69: can we change this to only register stores for initialized  pca9539rs? the i2c address is passed to pca9539r_gpio_init so this shouldn't be a problem.

g-jessmuir (changes requested):

libraries/mpxe-store/src/store.c
JarvisWeng: Yeah, I just added the total for now and I figured we could increase the maximum store number as we need it.
I just needed 257 for my testing of pca9539r. I could change it to a 1000.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
JarvisWeng: It does compile. Yes you're right, it should be uint8_t. My bad.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
JarvisWeng: I asked you about that before and you said to just initialize all of them. I can change that though

libraries/mpxe-store/src/store.c
```diff
@@ -14,7 +14,9 @@
 #include "log.h"
 #include "stores.pb-c.h"
 
-#define MAX_STORE_COUNT 64
+#define MAX_STORE_COUNT 1000
```
g-jessmuir, line 17: I think my point was that we shouldn't need that many since we shouldn't be initializing them all for all the i2c addresses etc., sorry should have clarified

g-jessmuir (changes requested):
Just the point about initialization

libraries/mpxe-store/src/store.c
JarvisWeng: Sounds good. I'll just change it back to 64

copperium (commented):
Just this one thing I saw

copperium:
Also (@g-jessmuir / @JarvisWeng), forgive me but I don't really understand the purpose of `s_stores` in stores.c. Its data doesn't seem to ever be accessed - `store_register` just stores new information in it and `store_get` is never called. Dispatching functions are taken care of by `s_store_funcs`. What's the point of `s_stores`?

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -20,36 +19,41 @@ static I2CPort s_i2c_port = NUM_I2C_PORTS;
 static Pca9539rGpioSettings s_pin_settings[MAX_I2C_ADDRESSES][NUM_PCA9539R_GPIO_PINS];
 
 #ifdef MPXE
+
+static MxPca9539rStore s_stores[MAX_I2C_ADDRESSES];
+
 #define PCA9539_I2C_ADDRESS 0x74  // PCA9539 address used in the smoke test
```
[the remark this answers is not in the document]
JarvisWeng: Yes thank you. Forgot about that.

JarvisWeng:
> Also (@g-jessmuir / @JarvisWeng), forgive me but I don't really understand the purpose of `s_stores` in stores.c. Its data doesn't seem to ever be accessed - `store_register` just stores new information in it and `store_get` is never called. Dispatching functions are taken care of by `s_store_funcs`. Do we need `s_stores`? (If not, the max store count is irrelevant.)

The `s_stores` should hold all the stores being used in `mpxe`. So if we had a test that had `gpio` and `pca9395r` stores, then `s_stores` will hold all the stores created.
`store_register` registers a new store within `s_stores`
`store_get` should be called if we ever write a C test that we need to call a specific store. Maybe useful for the future.
`s_stores` is needed to actually hold the space where the data is being stored.
So in all the `prv_init_store` we start with a `store_config` which should create a `prv_poll_update` and that should call the `prv_handle_store_update`. And the `prv_handle_store_update` needs the s_store, in order to get the specific store to update.

@g-jessmuir correct me if I'm wrong

g-jessmuir (approved):

g-jessmuir:
@copperium In theory for something like solar where we need to check the state of the gpio pins to determine what needs to happen in solar mpxe, we need to be able to access the other stores via `s_stores`. I had to think about it for a sec, almost said yeah wipe it all out lol

copperium:
@g-jessmuir Ahh I see. Still, I wonder if we could roll a linked list instead of using a fixed array so we don't have to deal with updating the maximum.

PR #262, opened 2021-01-23
FW102 Module

Cookieth:
Firmware 102: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1268121797/Firmware+102+Notes

### DIFF
--- a/projects/FW_102/src/main.c
+++ b/projects/FW_102/src/main.c
@@ -7,8 +7,7 @@
 #include "soft_timer.h"
 #include "wait.h"
 
-#define A_COUNTER_MS 500
-#define B_COUNTER_MS 1000
+#define COUNTER_MS 500
 
 typedef struct {
   uint8_t counter_a;
@@ -28,7 +27,7 @@ int main(void) {
   };
 
   // Increment counter a every 0.5s, counter b ever 1s.
-  soft_timer_start_millis(500, prv_inc_counter, &loop_counter, 0);
+  soft_timer_start_millis(COUNTER_MS, prv_inc_counter, &loop_counter, 0);
 
   // Wait for an interrupt
   while (true) {
@@ -50,5 +49,5 @@ static void prv_inc_counter(SoftTimerId timer_id, void *context) {
     LOG_DEBUG("Counter B: %d\n", counter->counter_b);
   }
 
-  soft_timer_start_millis(500, prv_inc_counter, context, 0);
+  soft_timer_start_millis(COUNTER_MS, prv_inc_counter, context, 0);
 }
--- a/projects/FW_102/src/main.c
+++ b/projects/FW_102/src/main.c
@@ -1,0 +1,57 @@
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "delay.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void inc_counter_a(SoftTimerId timer_id, void *context);
+static void inc_counter_b(SoftTimerId timer_id, void *context);
+
+int main(void) {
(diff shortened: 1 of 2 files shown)


projects/FW_102/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+  soft_timer_init();
+
+  Counters loop_counter = {
+    .counter_a = 0,
+    .counter_b = 0,
+  };
+
+  // Increment counter a every 0.5s, counter b ever 1s.
+  soft_timer_start_millis(500, inc_counter_a, &loop_counter, 0);
+  soft_timer_start_millis(1000, inc_counter_b, &loop_counter, 0);
```
copperium, line 31: It's good practice to #define constants for arbitrary numbers like these durations instead of just using them. So please add #defines for these durations with a descriptive name that ends with the unit, like `_MS`. Also use them in the `soft_timer_start_millis` call in the functions below.

projects/FW_102/src/main.c
```diff
@@ -0,0 +1,57 @@
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "delay.h"
```
copperium, line 5: I don't think this include is necessary? Please remove

projects/FW_102/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+    .counter_b = 0,
+  };
+
+  // Increment counter a every 0.5s, counter b ever 1s.
+  soft_timer_start_millis(500, inc_counter_a, &loop_counter, 0);
+  soft_timer_start_millis(1000, inc_counter_b, &loop_counter, 0);
+
+  // Wait for an interrupt
+  while (true) {
+    wait();
```
copperium, line 35: Awesome job using `wait()`!

copperium (changes requested):
Awesome, a few changes before moving on to the FW 103 homework

projects/FW_102/src/main.c
Cookieth: Thank you! For reviews, would you like me to respond to the comments that I have made changes on, or should I let the `outdated` tag speak for itself?

copperium: If you have questions / comments then definitely comment, otherwise the `outdated` tag is fine; resolving conversations for comments you've resolved is also good.

Cookieth: Should I let the reviewer resolve it as a standard practice, or is this more of a flexible rule dependent on the reviewer?

copperium: Eh it's not too strict, I know I appreciate it when people resolve my comments :)

projects/FW_102/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+  interrupt_init();
+  soft_timer_init();
+
+  Counters loop_counter = {
+    .counter_a = 0,
+    .counter_b = 0,
+  };
+
+  // Increment counter a every 0.5s, counter b ever 1s.
+  soft_timer_start_millis(500, prv_inc_counter, &loop_counter, 0);
```
copperium, line 31: I think you forgot to use the `X_COUNTER_MS` macros here and in the other `soft_timer_start_millis` call

projects/FW_102/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+#include <stddef.h>
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define A_COUNTER_MS 500
+#define B_COUNTER_MS 1000
```
copperium, line 11: You only need the 500ms macro here since you removed the second loop, and the `A_` prefix should be removed

copperium (changes requested):
Just a couple of things that I think you forgot

projects/FW_102/src/main.c
Cookieth: My bad! Must have rushed through everything

copperium (approved):

copperium:
Closing this since homework is complete.

PR #264, opened 2021-01-24
FW103 Module

Cookieth:
Without the ability to test, I am completely unsure whether this code works or not (It builds, though!). I pieced my knowledge through this by looking at the documentation in the `libraries/libcore/inc` and `libraries/mc-common/inc` directories, as well as various confluence files. Please let me know if I had assumed anything wrong, or have missed anything crucial!

Also note, see the README.md for my answers to the scavenger hunt.

### DIFF
--- a/projects/FW_103/README.md
+++ b/projects/FW_103/README.md
@@ -1,0 +1,62 @@
+Datasheet scavenger hunt. Hint: ctrl + f is your friend!
+
+# What bit in what register should be set on the MCP2515 CAN controller to enable one-shot mode?
+ Bit 3, Register 10-1 (Address 0xXF), 1 = enabled, 0 = disabled. Source: https://ww1.microchip.com/downloads/en/DeviceDoc/MCP2515-Stand-Alone-CAN-Controller-with-SPI-20001801J.pdf
+
+
+# In the MCP2515, what are the 3 bytes you’d send over spi if you wanted to write 0x17 to register 0x0f?
+Byte 1: 0000 0010 (instruction)
+Byte 2: 0001 0111 (0x17)
+Byte 3: 0000 1111 (0x0f)
+Source: https://ww1.microchip.com/downloads/en/DeviceDoc/MCP2515-Stand-Alone-CAN-Controller-with-SPI-20001801J.pdf Page 68, Figure 12-4
+
+
+# What is the digital output code (16 bits) for the MCP3427 when the input voltage is 0
+Input Voltage (CHn+ - CHn-)*PGA = 0
+Digital Output Code = 0000000000000000
+
+
+# What is the first Opcode byte for to start a conversion on the  ADS1259
+Byte: 0000 100x (where x is any)
+Hex Representation: 0x08 or 0x09 (since x is any)
+Source: https://www.ti.com/lit/ds/symlink/ads1259.pdf?ts=1611514736643&ref_url=https%253A%252F%252Fwww.ti.com%252Fproduct%252FADS1259 page 32
+
+
+# What is the address (hex) for GPIO in the MCP23008?
+Address: 0x09
+Source: https://ww1.microchip.com/downloads/en/DeviceDoc/21919e.pdf. page 6
+
+
+# What are the fault codes for the when a fault is detected in bits 15-12 LTC6811?
+ Fault code (binary): 0b1xxx (where x is any)
+
+===========================================================================================================================
+
+## Schematic scavenger hunt. Note: MSXII schematics are used because MSXIV schematics are undergoing constant revision.
+
+# On the MSXII AFE board temperature mux, how many inputs are there? How many pins are needed to talk to the controller board?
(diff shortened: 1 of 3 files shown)


projects/FW_103/src/main.c
```diff
@@ -0,0 +1,88 @@
...
+    GPIO_DIR_IN, GPIO_STATE_HIGH, GPIO_RES_NONE,
+    GPIO_ALTFN_3,  // I believe Pin B2 can only have the 3rd alternate function according to
+                   // datasheet (see above)
+  };
+
+  gpio_init_pin(&button_addr, &button_settings);
+
+  // Get ADC channel
+
+  AdcChannel ADC_channel = NUM_ADC_CHANNELS;
```
copperium, line 59: Please use all lowercase! `adc_channel`

projects/FW_103/src/main.c
```diff
@@ -0,0 +1,88 @@
...
+
+  // Was told to not "worry about details", not sure what this does.
+  adc_get_channel(ADC_addr, &ADC_channel);
+  adc_set_channel(ADC_channel, true);
+  // adc_set_channel(ADC_CHANNEL_TEMP, true);
```
copperium, line 64: So what this does is the STM32 ADC doesn't speak in GPIO pins, but rather in ADC "channels", not all of which are associated with a GPIO pin (e.g. `ADC_CHANNEL_TEMP` is the channel for the STM32's internal thermistor). `adc_get_channel` retrieves the channel for a GPIO pin, and `adc_set_channel(adc_channel, true)` turns it on.

(FYI: if you look at adc.h we also have fancy new functions like `adc_set_channel_pin` and `adc_read_converted_pin` which take GPIO addresses instead of ADC channels and do the `adc_get_channel` lookup internally.)

projects/FW_103/src/main.c
```diff
@@ -0,0 +1,88 @@
...
+
+  LOG_DEBUG("ADC Channel: %d\n", ADC_channel);
+
+  // Taking documentation from gpio_it.h and interrupt_def.h, and making guesses as to settings
+  InterruptSettings buttonInterruptSettings = {
+    INTERRUPT_TYPE_INTERRUPT,
+    INTERRUPT_PRIORITY_NORMAL,
+  };
```
copperium, line 72: Please use `snake_case` and the GNU struct initialization style! And yes the settings are correct.

projects/FW_103/src/main.c
```diff
@@ -0,0 +1,88 @@
...
+  gpio_it_register_interrupt(&button_addr, &buttonInterruptSettings, INTERRUPT_EDGE_RISING,
+                             prv_get_adc_callback, &ADC_channel);
+
+  while (true) {
+    wait();
+  }
+}
+
+void prv_get_adc_callback(const GpioAddress *address, void *context) {
+  AdcChannel *channelData = (AdcChannel *)context;
```
copperium, line 84: Please use `snake_case`!

projects/FW_103/src/main.c
```diff
@@ -0,0 +1,88 @@
...
+  uint16_t data = 0;
+  adc_read_raw(*channelData, &data);
```
copperium, line 86: The homework description does say to use an ADC _converted_ read - the difference is that converted reads, with `adc_read_converted`, return a result scaled to millivolts, whereas raw reads with `adc_read_raw` just return whatever raw number the ADC spits out.

projects/FW_103/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# What are the fault codes for the when a fault is detected in bits 15-12 LTC6811?
+ Fault code (binary): 0b1xxx (where x is any)
+
+===========================================================================================================================
+
+## Schematic scavenger hunt. Note: MSXII schematics are used because MSXIV schematics are undergoing constant revision.
+
+# On the MSXII AFE board temperature mux, how many inputs are there? How many pins are needed to talk to the controller board?
+Weirdly, I could only find the MSXIV schematic (thus contradicting the above statement: https://university-of-waterloo-solar-car-team.365.altium.com/designs/9D181E70-999D-4D1E-8C98-3B5285CC6775?variant=%5BNo%20Variations%5D#design)
```
copperium, line 38: The scavenger hunt is probably outdated, sorry for the confusion there

copperium (changes requested):
Awesome job! For the scavenger hunt, I don't have the answers handy but I'll send them to you when I get them - it's not really meant to be marked anyways. And you're right that this project isn't really testable without hardware, it does look like it'll work though.

projects/FW_103/src/main.c
Cookieth: Ah, so we set `adc_channel` to `NUM_ADC_CHANNELS` just to instantiate it, knowing that `adc_get_channel()` will modify it?

copperium: Yep, the idea is that if `adc_get_channel` fails then we end up with a definitely-invalid `adc_channel` and not a possibly-valid-but-incorrect one which could cause worse problems.

projects/FW_103/README.md
Cookieth: No problem, it was a lot of fun, actually! I think it really helped me gain confidence in navigating everything.

copperium:
Great! Closing since homework is complete.

PR #266, opened 2021-01-25
FW 103 Homework

hzhou0:
Code review please.

### DIFF
--- a/projects/hello_adc/src/hello_adc.c
+++ b/projects/hello_adc/src/hello_adc.c
@@ -1,0 +1,4 @@
+//
+// Created by henry on 2021-01-24.
+//
+
--- a/projects/hello_adc/src/hello_adc.c
+++ b/projects/hello_adc/src/hello_adc.c
@@ -6,23 +6,22 @@
 #include "log.h"
 #include "soft_timer.h"
 
-#define BUTTON_PIN 178
-#define READ_FROM_PIN 166
+static void prv_log_pin_data(GpioAddress *button_addr, void *context);
 
-GpioItCallback prv_log_pin_data(GpioAddress *button_addr, void *context);
-
-int main() {
+int main(void) {
   interrupt_init();
   soft_timer_init();
   gpio_init();
-  GpioAddress button_addr = { GPIO_PORT_A, BUTTON_PIN };
-  GpioAddress read_addr = { GPIO_PORT_A, READ_FROM_PIN };
+  GpioAddress button_addr = { GPIO_PORT_A, 6 };
+  GpioAddress read_addr = { GPIO_PORT_B, 2 };
   // Setting pin value is not necessary
-  GpioSettings null_settings = {};
-  gpio_init_pin(&button_addr, &null_settings);
-  gpio_init_pin(&read_addr, &null_settings);
+  GpioSettings pin_settings = { .direction = GPIO_DIR_IN, .alt_function = GPIO_ALTFN_ANALOG };
+  gpio_init_pin(&button_addr, &pin_settings);
+  gpio_init_pin(&read_addr, &pin_settings);
   adc_init(ADC_MODE_SINGLE);
-  InterruptSettings button_interrupt = { INTERRUPT_TYPE_EVENT, INTERRUPT_PRIORITY_NORMAL };
+  adc_set_channel_pin(read_addr, true);
+  InterruptSettings button_interrupt = { .type = INTERRUPT_TYPE_EVENT,
+                                         .priority = INTERRUPT_PRIORITY_NORMAL };
   gpio_it_register_interrupt(&button_addr, &button_interrupt, INTERRUPT_EDGE_FALLING,
                              (GpioItCallback)prv_log_pin_data, &read_addr);
   while (true) {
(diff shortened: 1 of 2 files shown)


projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+#define BUTTON_PIN 178
+#define READ_FROM_PIN 166
+
+GpioItCallback prv_log_pin_data(GpioAddress *button_addr, void *context);
```
copperium, line 12: This declares a function that *returns* a `GpioItCallback`, not one that *is* a `GpioItCallback`! Any function that matches the `GpioItCallback` can be used as one, you don't have to declare it implicitly; `GpioItCallback` is just a typedef for `void (*)(GpioAddress *, void *)` which is C-speak for "a pointer to a function that returns void and takes a GpioAddress pointer and a void pointer". So this needs to return `void` and not `GpioItCallback`.

Also, it should be `static` since it's a private function - that makes sure it's not visible to other files.

projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
...
+
+int main() {
```
copperium, line 14: This parameter list should be `(void)` rather than `()` - in C, [weird things](https://stackoverflow.com/questions/13950642/why-does-a-function-with-no-parameters-compared-to-the-actual-function-definiti) can happen with functions taking empty parameter lists like in other languages rather than explicitly declaring them as `(void)`.

projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+#define BUTTON_PIN 178
+#define READ_FROM_PIN 166
```
copperium, line 10: "Pin A6" and "Pin B2" doesn't refer to hex lol, the letter is the port (`GPIO_PORT_A`, `GPIO_PORT_B`, ...) and the number is the pin (0 through 15). So you don't need these defines; remove them and specify the addresses in the struct initializations in `main`.

(FYI: if in the future you need to write numbers which are best expressed in hex (which you don't need to for this homework!), use the `0x` prefix: `0x100 == 256`. `0b` also works for binary: `0b100 == 4`.)

projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
...
+GpioItCallback prv_log_pin_data(GpioAddress *button_addr, void *context);
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+  GpioAddress button_addr = { GPIO_PORT_A, BUTTON_PIN };
+  GpioAddress read_addr = { GPIO_PORT_A, READ_FROM_PIN };
+  // Setting pin value is not necessary
+  GpioSettings null_settings = {};
```
copperium, line 21: The GPIO settings aren't just for setting the initial pin value: you'll need to specify the pin direction (should be `GPIO_DIR_IN`) and for any address you're going to do an ADC read on, you'll need to specify the alt function as `GPIO_ALTFN_ANALOG`.

projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
...
+  gpio_init_pin(&button_addr, &null_settings);
+  gpio_init_pin(&read_addr, &null_settings);
+  adc_init(ADC_MODE_SINGLE);
+  InterruptSettings button_interrupt = { INTERRUPT_TYPE_EVENT, INTERRUPT_PRIORITY_NORMAL };
```
copperium, line 25: FYI: it's alright for small one liners, but we generally prefer the GNU struct initialization syntax like:
```c
InterruptSettings button_interrupt = {
  .type = INTERRUPT_TYPE_EVENT,
  .priority = INTERRUPT_PRIORITY_NORMAL,
};
```

projects/hello_adc/src/hello_adc.c
```diff
@@ -0,0 +1,39 @@
...
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+  GpioAddress button_addr = { GPIO_PORT_A, BUTTON_PIN };
+  GpioAddress read_addr = { GPIO_PORT_A, READ_FROM_PIN };
+  // Setting pin value is not necessary
+  GpioSettings null_settings = {};
+  gpio_init_pin(&button_addr, &null_settings);
+  gpio_init_pin(&read_addr, &null_settings);
+  adc_init(ADC_MODE_SINGLE);
```
copperium, line 24: You will need to call `adc_set_channel_pin(read_addr, true)` to enable ADC reading on the `read_addr` pin here (after initializing ADC).

copperium (changes requested):
Nice work!

copperium:
Looks good! You can see me for a task and/or move on to FW 104.

PR #267, opened 2021-01-30
Soft 377 spi exchange firmware

Kyle02:


projects/baby_driver/test/test_spi_exchange.c
```diff
@@ -0,0 +1,354 @@
...
+                            1, 0, 0, 0, 0, 0, 0);
+  else if (bytes == 3)
+    CAN_TRANSMIT_BABYDRIVER(BABYDRIVER_MESSAGE_SPI_EXCHANGE_TX_DATA,  // id
+                            1, 1, 1, 0, 0, 0, 0);
+  else if (bytes == 7)
+    CAN_TRANSMIT_BABYDRIVER(BABYDRIVER_MESSAGE_SPI_EXCHANGE_TX_DATA,  // id
+                            1, 1, 1, 1, 1, 1, 1);
+
+  MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+}
```
copperium, line 139: It seems a bit sketchy for this function to only work for three specific values of `bytes`; it would definitely be better for it to work for 0-7. So that you don't have to use a big huge if statement, try using the plain old `can_transmit` function from can.h: construct a `CanMessage` (from can_msg.h), and use `NULL` for the ack request. (Take a look at the implementation of the `CAN_TRANSMIT_MACRO` for hints.)

Also, please use {curly brackets} for if statements that aren't super short on one line.

projects/baby_driver/test/test_spi_exchange.c
```diff
@@ -0,0 +1,354 @@
...
+  TEST_CAN_EVENT_RX,
+  TEST_CAN_EVENT_FAULT,
+  NUM_TEST_CAN_EVENTS,
+} TestCanEvent;
+
+static CanStorage s_can_storage;
+
+static uint8_t s_callback_counter_status;
+static uint8_t s_num_messages;
+static uint8_t s_received_data_status[8];
```
copperium, line 36: You don't really need to store all 8 bytes of the status message, eh? You already know that the 0th byte is `BABYDRIVER_MESSAGE_STATUS` (since the callback wouldn't be called otherwise!), so you don't need to test it. You only really need to store the status in byte 1.

projects/baby_driver/test/test_spi_exchange.c
```diff
@@ -0,0 +1,354 @@
...
+      MS_TEST_HELPER_CAN_TX_RX(TEST_CAN_EVENT_TX, TEST_CAN_EVENT_RX);
+    }
+    // right now there is no way to receive this many messages without the helper issue
+
+    TEST_ASSERT_EQUAL(292, s_num_messages);
+    test_equal_multiple_messages();
+
+    clear_received_data_array();
+    s_num_messages = 0;
+  */
```
copperium, line 238: I think it's alright to omit this test with just a comment saying what it is and why it was omitted.

projects/baby_driver/test/test_spi_exchange.c
```diff
@@ -0,0 +1,354 @@
...
+
+StatusCode TEST_MOCK(spi_exchange)(SpiPort spi, uint8_t *tx_data, size_t tx_len, uint8_t *rx_data,
+                                   size_t rx_len) {
+  for (uint8_t i = 0; i < rx_len; i++) rx_data[i] = 1;
+
+  return STATUS_CODE_OK;
+}
+
+StatusCode TEST_MOCK(spi_init)(SpiPort spi, const SpiSettings *settings) {
+  // still need to add tests making sure the settings and port are the same
```
copperium, line 349: Do you still?

projects/baby_driver/inc/spi_exchange.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+// This module provides the function spi_exchange_init()
+// for spi_exchange CAN messages from the babydriver. Once the message is received,
+// the spi_init is run with the inputs and the spi_exchange function is run
```
copperium, line 5: You need a `// Requires x, y, and z to be initialized.` comment

copperium (changes requested):
Great job Kyle! General feedback: add more comments, some of the logic is pretty complex and it will help future readers to explain it.

projects/baby_driver/src/spi_exchange.c
```diff
@@ -0,0 +1,220 @@
...
+    case 1:
+      spi_mode = (SpiMode)SPI_MODE_1;
+      break;
+    case 2:
+      spi_mode = (SpiMode)SPI_MODE_2;
+      break;
+    case 3:
+      spi_mode = (SpiMode)SPI_MODE_3;
+      break;
+  }
```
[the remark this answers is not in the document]
Kyle02: They don't need to be of type SpiMode?

projects/baby_driver/test/test_spi_exchange.c
Kyle02: Oh I thought I deleted this comment, I finished that.

copperium (changes requested):
Just one thing: the coding standards prescribe against /\* \*/ style comments (and in fact there's a ticket going to remove any instances of /\* \*/ from the projects/ directory). Please change them to // line comments and then we can merge this!

copperium (approved):

PR #268, opened 2021-02-01
Daniel Zeng's FM 102 HW

PurpleRhythms:


projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,47 @@
...
+#include "wait.h"        // for wait function
+
+#define COUNTER_A_INCREMENT_INTERVAL_MS 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void timer_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 15: Since this is a private function, it should be `static` and its name should be prefixed with `prv_`. `static` means that the function can't be seen by any other file; without `static`, if another file tried to define a `timer_callback` function, it might cause issues. The `prv_` is just a convention ([and you can see the full list here!](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards)).

copperium (changes requested):
Awesome, just one thing to fix and then you'll be good to go!

projects/timer_challenge/.project
```diff
@@ -0,0 +1,11 @@
...
+<projectDescription>
+	<name>timer_challenge</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
```
copperium, line 11: I think this was committed accidentally, what IDE are you using? We can add it to the .gitignore so it doesn't get committed in the future.

copperium:
Look good! Closing since homework is complete, feel free to move on to FW 103.

PR #270, opened 2021-02-02
FW 104 Module

Cookieth:
Here's what I have initially for Onboarding FW 104. I'm expecting a lot of issues, since I had to do a lot of guesswork. I believe my CAN functionality works, but I wasn't too sure what was meant by:
> Then, register callbacks for both that print the data, but only ACK the 0xA message.  

So, I just registered and printed both data since I'm a bit unsure what it means to ACK a message (acknowledge, i believe). Only the message for the CAN message with ID `0xA` is received, with the value `5`, which is the correct behaviour (I believe) since it is of higher priority than `0xB` (which has the message with value `10`.

I'm a bit confused on how the I2C library works, in particular with the distinction between `PORT`, `REGISTER`, and `ADDRESS`. In my mind, these are all sort of the same thing, so I was very lost. A bit help/explanation would be much appreciated! I did a bit of guesswork with the `PORT`, so correction on that part is a bit expected.

Further, I did a bit of guesswork with getting the SPI baud rate, mode, GPIO ports, and port. I looked through the datasheets and I feel like what I have written is correct, but I can't help but feel like I'm looking in the wrong place. Guidance on this part would also be much appreciated!


Cookieth:
Though this builds on my machine, there seems to be a build error on Travis. I will look into this locally tomorrow.

Cookieth:
Fixed the build issue, it turned out to be an inconsistency with capitalization in my files. My system seemed to be able to build with the inconsistency. Git was unable to detect the fix in capitalization (for some odd reason), so I just changed the name altogether (`i2c` -> `iic`).

projects/FW_104/src/fw_104_can.c
```diff
@@ -0,0 +1,102 @@
...
+  soft_timer_start_millis(CAN_SEND_TIME_MS, prv_can_transmit_A, NULL, NULL);
+
+  // The prv_can_callback did not trigger unless I put this event queue stuff... (Copied from
+  // smoke_can)
+  Event e = { 0 };
+  while (true) {
+    while (event_process(&e) == STATUS_CODE_OK) {
+      can_process_event(&e);
+    }
+  }
```
copperium, line 70: Yep, this is necessary - the tl;dr is the CAN library uses the event queue to drive its operation, so you need this event queue loop so its events continuously get processed. Since it's an infinite loop, you really should only have one of these in a program,  usually it's in main.c - currently this function never returns.

projects/FW_104/src/fw_104_can.c
```diff
@@ -0,0 +1,102 @@
...
+  printf("\n");
+  return STATUS_CODE_OK;
+}
+
+void prv_can_transmit_A(SoftTimerId timer_id, void *context) {
+  uint8_t messageData = 5;
+  CanMessage message = {
+    .source_id = 0x1, .msg_id = 0x1, .data_u8 = { messageData }, .type = CAN_MSG_TYPE_DATA, .dlc = 8
+  };
+  can_transmit(&message, NULL);
```
copperium, line 27: The 0xA and 0xB in the homework refers to the message ID; the source ID should just match the device ID you initialized with. DLC is the data-length-code, it's the length of the body in bytes (so if you only want to transmit 1 uint8, set it to 1). And to request an ACK, pass a CanAckRequest * (defined in can_ack.h) instead of NULL to can_transmit. For the CanAckRequest, the callback is called when enough ACK messages are received, and the `expected_bitset` is a bitset of the device IDs you want acknowledgement messages from.

20-second rundown on ACKs: when a message is received that is considered "critical" (message id < 13), the CAN library will send back an "ACK" message which has the same message ID but has the ACK bit in the arbitration ID toggled on. The `ack_request` parameter to `can_transmit` tells the CAN library to watch for ACKs from certain device IDs (the expected bitset) and call a callback when those device IDs ACK it. This lets us make sure that certain boards receive certain critical messages.

projects/FW_104/src/fw_104_can.c
```diff
@@ -0,0 +1,102 @@
...
+    .device_id = 0xA,
+    .bitrate = CAN_HW_BITRATE_500KBPS,
+    // Not really sure what to put here, following smoke_can
+    .rx_event = 0,
+    .tx_event = 1,
+    .fault_event = 2,
+    .tx = { GPIO_PORT_A, 12 },
+    .rx = { GPIO_PORT_A, 11 },
+    .loopback = true,
+  };
```
copperium, line 55: A few points:
* you only really need to initialize CAN once, it's usually done in main but for homework purposes, whatever
* `loopback = true` tells the CAN library to receive its own messages, I think the idea with the "run it on two terminals" thing was for loopback to be false so you see the messages and the ACKs travelling across the virtual CAN bus between the two instances of your program
* the tx/rx/fault_event are event IDs. I'll talk about the event queue on Saturday but those events are used for driving the CAN library (tx is raised when a message is being tx'd (transmitted), rx is raised when a message was rx'd (received), and fault is raised on a fault). Usually you should define them in an enum kinda like this:
```c
typedef enum {
  FW_104_CAN_RX_EVENT = 0,
  FW_104_CAN_TX_EVENT,
  FW_104_CAN_FAULT_EVENT,
  NUM_FW_104_CAN_EVENTS,
} Fw104CanEvent;
```
You can see examples of these event enums in e.g. solar_events.h.

copperium (changes requested):
Nice job! If you want, you can see me for a task when you're done with this homework (or before).

projects/FW_104/src/fw_104_iic.c
```diff
@@ -0,0 +1,26 @@
...
+  StatusCode status = NUM_STATUS_CODES;
+  uint8_t message[2] = { 0b00100100, 0b10100110 };
+
+  // SDA PIN Address = 0b1001010
+  // Conversion Register = 0b0000000
+  // Configuration Register = 0b0000001
+  status = i2c_write_reg(I2C_PORT_1, 0b1001010, 0b0000001, message, ARRAY_SIZE(message));
+
+  uint8_t data[2];
+  status = i2c_read_reg(I2C_PORT_1, 0b1001010, 0b0000000, data, ARRAY_SIZE(data));
```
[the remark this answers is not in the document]
Cookieth: Could I collect questions I have regarding the difference between addresses and registers, and have a short chat with you about them? I'm a bit confused about all of it, I sort of assumed they were the same thing, but it's quite obvious that that's a big gap in my knowledge. That would be much appreciated, thanks!

projects/FW_104/src/fw_104_can.c
Cookieth: I'm a bit confused with the function of the device id, source id, and message id. For the sake of this module, should the instance sending message id 0xA have the same device id/source id as the instance sending message id 0xB?

Cookieth:
I've made an initial commit to try to fix the gaps with my CAN implementation, but there's a lot of guesswork, and I'm not sure of everything. I feel like perhaps I need to do some more research and try to figure out the CAN protocol so that I have better context as to everything that is happening. What are your thoughts?

Cookieth:
`test_bms_carrier` is failing my build... I'm trying to diagnose what it is, but I can't quite pin it down. I believe it's due to the segmentation fault that occasionally returns from `make run PROJECT=FW_104 PLATFORM=x86`.

copperium:
I'm going to close this since it's been like 4 months and I think the code is fine, and you've moved on anyways.

PR #271, opened 2021-02-03
[SOFT-419] Enable parameterized tests

copperium:
Our testing framework Unity gives us a way to use parameterized tests! See test_parameterized.c for an example. (Apparently it breaks github syntax highlighting though, oh well.)

Highlights:
* `TEST_CASE` macro required for it to work defined in test_helpers.h
* parameterized tests enabled with a flag to the test runner generator script in the makefile
* example in test_parameterized.c
* monkey patch to the test runner generator script to remove `(RUN_TEST_NO_ARGS)` spam in the test logs

### DIFF
--- a/libraries/libcore/inc/test_helpers.h
+++ b/libraries/libcore/inc/test_helpers.h
@@ -11,3 +11,6 @@
 
 // Mocking
 #define TEST_MOCK(func) __attribute__((used)) __wrap_##func
+
+// Parameterized tests, see test_parameterized.c for usage examples
+#define TEST_CASE(...)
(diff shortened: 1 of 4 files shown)


g-jessmuir (approved):
I think it wouldn't hurt to write out an example test for a real module to prove out the use case, but seems like a neat and harmless change.

copperium:
I did partially convert some of charger's test_control_pilot.c for [the confluence page](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2107539612/Parameterized+Unit+Tests). The whole thing will get a lot less awkward if we update Unity, which we apparently haven't done since 2014 - I made a ticket for this, SOFT-420.

PR #272, opened 2021-02-04
Soft 327 integrate python tests

shahhet14:
Created 2 targets that can run tests from a specified project or run all tests for all projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,9 @@ before_script:
   # Install pylint for python3 used by miniconda
   - pip install --user pylint
 
+  # Install python requirements for every project
+  - make install_requirements
+
   # force ${PATH} to update
   - hash -r
 
@@ -107,5 +110,5 @@ script:
   - make clean
   - make build_all PLATFORM=x86 DEFINE="${DEFINES}"
   - make test_all PLATFORM=x86 DEFINE="${DEFINES}"
-  - make test_all_python
+  - make pytest_all
   - make build_all PLATFORM=x86 COMPILER=clang DEFINE="${DEFINES}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,4 +107,5 @@ script:
   - make clean
   - make build_all PLATFORM=x86 DEFINE="${DEFINES}"
   - make test_all PLATFORM=x86 DEFINE="${DEFINES}"
+  - make test_all_python
   - make build_all PLATFORM=x86 COMPILER=clang DEFINE="${DEFINES}"
--- a/Makefile
+++ b/Makefile
@@ -248,11 +248,15 @@ update_codegen:
 
 .PHONY: test_python
 test_python:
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -107,4 +107,5 @@ script:
   - make clean
   - make build_all PLATFORM=x86 DEFINE="${DEFINES}"
   - make test_all PLATFORM=x86 DEFINE="${DEFINES}"
+  - make test_all_python
```
copperium, line 110: Travis is failing because it can't find various requirements from the libraries - I'd suggest fixing this by adding something in the `before_script` section which finds all `requirements.txt` files in the repo (they contain python requirements for each project) and runs `pip install -r <path/to/requirements.txt>` on them.

Makefile
```diff
@@ -246,5 +246,17 @@ socketcan:
 update_codegen:
 	@python make/git_fetch.py -folder=libraries/codegen-tooling -user=uw-midsun -repo=codegen-tooling-msxiv -tag=latest -file=codegen-tooling-out.zip
 
+.PHONY: test_python
+test_python:
+	@python3 -m unittest discover -t $(PROJ_DIR)/$(PROJECT)/scripts -s $(PROJ_DIR)/$(PROJECT)/scripts -p "test_*$(TEST).py"
+
+#find projects -name "test_*.py" -print0 | xargs -0 python -m unittest
```
copperium, line 253: Please remove this comment, it's not necessary anymore

Makefile
```diff
@@ -246,5 +246,17 @@ socketcan:
...
+.PHONY: test_all_python
+test_all_python:
+	for i in $$(find projects -name "test_*.py"); \
+	do											\
+		python -m unittest discover -t $$(dirname $$i) -s $$(dirname $$i) -p $$(basename $$i);							\
+	done
```
copperium, line 259: You need an `@` before the `for` so that it doesn't print the command. Also the backslashes are in super inconsistent positions, please either have all of them one space after the command or align them or something.

Makefile
```diff
@@ -246,5 +246,17 @@ socketcan:
 update_codegen:
 	@python make/git_fetch.py -folder=libraries/codegen-tooling -user=uw-midsun -repo=codegen-tooling-msxiv -tag=latest -file=codegen-tooling-out.zip
 
+.PHONY: test_python
+test_python:
```
copperium, line 250: `test_python` is a little bit long for a thing people will be typing over and over again, maybe `pytest` would be better (in the same vein as `pylint`). If you change it to that also change `test_all_python` -> `pytest_all`.

copperium (changes requested):
Sorry the review took so long, but nice job! Also please put the `test_python` and `test_all_python` (or what you change the names to) in both the comments at the top of the Makefile and in the README so people can find them without trawling through the makefile.

shahhet14:
Changed the names for the targets used for testing and created a new target to install all python requirements. Documented all of the targets in the readme and at the top of the makefile.

copperium (approved):

PR #273, opened 2021-02-06
[SOFT-416] Minimal bootloader with linker scripts

copperium:
_Now actually works!_

This adds a minimal bootstrapped version of the bootloader. The bootloader itself does nothing except jump to the application code. The bulk of the complexity is in the linker scripts for organizing the bootloader and application code in memory.

There's also a provision for selecting stm32 linker scripts per project from rules.mk (mostly useful for the bootloader project), and a temporary makefile target, `temp-bootloader-write`, for building the bootloader and application code and flashing both together.


mitchellostler (approved):
So I had a look through, and it all seems to make sense to me. 
I definitely am not very familiar with the linker scripts, but I read through the zero to main articles, and based on what I understand they should be good to go? I think as we progress with this project I will try to learn more about them and hopefully be able to give better code reviews. 
The c code makes sense to me, I checked over each line (and had to do some learning as well), there doesn't seem to be any issues.

platform/stm32f0xx/ldscripts/bootloader/memory_spaces.ld
```diff
@@ -0,0 +1,39 @@
...
+
+_flash_start = 0x08000000;
+_flash_size = 128K;
+_flash_page_size = 2K;
+
+_ram_start = 0x20000000;
+_ram_size = 16K;
+
+_bootloader_start = _flash_start;
+_config_page1_start = _bootloader_start + _bootloader_size;
```
AveryChiu64, line 19: What are these config pages? I'm looking at the memory map but not sure I see it?
image

projects/bootloader/src/x86/jump_to_application.c
```diff
@@ -0,0 +1,5 @@
+#include "jump_to_application.h"
+
+void jump_to_application(void) {
+  // unless we try simulating the bootloader, this doesn't have much meaning on x86: do nothing
```
AveryChiu64, line 4: Maybe we could add a LOG_DEBUG here to warn someone if they try and run the bootloader on x86

platform/stm32f0xx/ldscripts/default/stm32f0discovery_def.ld
```diff
@@ -6,14 +6,16 @@ Linker subscript for STM32F072 definitions with 128K Flash and 16K RAM
...
+
 MEMORY
 {
   FLASH (rx) : ORIGIN = 0x08000000, LENGTH = 128K
   RAM  (xrw) : ORIGIN = 0x20000000, LENGTH = 16K
 }
 
 /* highest address of the user mode stack */
-_estack = 0x20002000;
+_estack = 0x20004000;
```
AveryChiu64, line 18: Where is this from? Can we just increase it like this?

projects/bootloader/src/x86/jump_to_application.c
AveryChiu64: Actually,  I'm wondering if we even want to keep this? Would there be a point of allowing people to specify that they want PLATFORM=x86 for the stm32 CAN bootloader? Maybe we could just remove this functionality? For our other code it makes sense for testing though I'm not sure if its as applicable to have an x86 version of CAN bootloader.

platform/stm32f0xx/ldscripts/bootloader/memory_spaces.ld
AveryChiu64: https://www.st.com/resource/en/datasheet/stm32f072c8.pdf

AveryChiu64 (approved):
I think this is a great start. I don't have too many comments since I know this is just the minimal bootloader and there are a lot of things to be implemented but I think this is a good starting point.

platform/stm32f0xx/ldscripts/bootloader/memory_spaces.ld
copperium: These are designations for the flash pages where we'll store the bootloader config - see https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2030895112/Bootloader+Flash+over+CAN#Memory-Layout-and-Config. It's not an intrinsic part of the stm32 memory layout, just a section of flash I set aside so we don't overwrite it with the bootloader code or other such silliness.

platform/stm32f0xx/ldscripts/default/stm32f0discovery_def.ld
copperium: This was done on the `mockup` branch and I added it here to get it in master - we actually have 16KB of SRAM, addresses 0x20000000 to 0x20004000, but we previously only used half of it. It seemed to be a mistake left over from some MSXII (maybe even MSXI)-era migration from a smaller chip; they updated the rest of the linker script but didn't touch _estack for some reason.

projects/bootloader/src/x86/jump_to_application.c
copperium: I added the x86 implementation because the stm32-specific version doesn't compile on x86, so I had to put it in an stm32f0xx directory. Then if we don't have the x86 version we get a linker error from the `jump_to_application()` call in main.c, and we can't put main.c itself in stm32f0xx/ since then we get an linker error complaining about a missing `main`. So it looks like the x86 version is here to stay.

I guess we could limit the bootloader in rules.mk or somewhere to compile only on stm32, but there are parts of the bootloader we'll want to unit test on x86 (e.g. the protocol implementation), so I think this way is good. The `LOG_DEBUG` is a good idea though.

PR #275, opened 2021-02-06
SOFT 422: Fix status_ok_or_return to only evaluate argument once

Cookieth:
See Ticket: https://uwmidsun.atlassian.net/browse/SOFT-422

Implemented statement expression to set function return to a variable, and return that variable if not equal to `STATUS_CODE_OK` (0).

A test has been added as well.

### DIFF
--- a/libraries/libcore/inc/status.h
+++ b/libraries/libcore/inc/status.h
@@ -84,6 +84,6 @@ void status_register_callback(StatusCallback callback);
 // Use to forward failures or continue on success.
 #define status_ok_or_return(code)          \
   ({                                       \
-    __typeof__(code) status_expr = code;   \
-    if (status_expr) return (status_expr); \
+    __typeof__(code) status_expr = (code); \
+    if (status_expr) return status_expr;   \
   })
--- a/libraries/libcore/inc/status.h
+++ b/libraries/libcore/inc/status.h
@@ -82,5 +82,8 @@ void status_register_callback(StatusCallback callback);
 #define status_clear() status_msg(STATUS_CODE_OK, "Clear")
 
 // Use to forward failures or continue on success.
-#define status_ok_or_return(code) \
-  if (code) return (code)
+#define status_ok_or_return(code)          \
+  ({                                       \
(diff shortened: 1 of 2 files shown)


Cookieth:
Latest commit doesn't build, but I don't think that's because of my variable change. It seems to be an issue with a `sudo apt-get` command line call

libraries/libcore/inc/status.h
```diff
@@ -82,5 +82,8 @@ void status_register_callback(StatusCallback callback);
 #define status_clear() status_msg(STATUS_CODE_OK, "Clear")
 
 // Use to forward failures or continue on success.
-#define status_ok_or_return(code) \
-  if (code) return (code)
+#define status_ok_or_return(code)          \
+  ({                                       \
+    __typeof__(code) status_expr = code;   \
+    if (status_expr) return (status_expr); \
```
copperium, line 88: You don't need the parentheses around `status_expr` in the return statement, but you do need parentheses around `code` in the assignment since it's a macro argument; due to precedence issues with the literal copy/pasting the preprocessor does it's good practice to always put parentheses whenever you use a macro argument.

Also, nice job with the `__typeof__`!

copperium (changes requested):
Awesome, thanks for getting started on it right away! Just a couple things to fix.

libraries/libcore/test/test_status.c
```diff
@@ -59,13 +59,29 @@ static StatusCode prv_ok_or_return() {
   return status_code(STATUS_CODE_UNKNOWN);
 }
 
+// Increments an external count to check how many times the function is called
+static StatusCode prv_status_num_calls(uint16_t *count) {
+  (*count)++;
+  return STATUS_CODE_OK;
```
copperium, line 65: Since this function returns `STATUS_CODE_OK`, this will actually pass with the old `status_ok_or_return` function - have this return some non-ok code just to make sure it actually tests the thing.

copperium (changes requested):
Sorry, one thing I forgot last review round.

copperium (approved):

PR #276, opened 2021-02-08
Added autopep8 format to build process

hzhou0:
Autopep8 needs to be installed through apt.


copperium (changes requested):
Great job Henry! To stop Travis from failing you'll want to run your `make format` over the repo - this will make your PR huge but that's alright.

Makefile
```diff
@@ -155,11 +155,16 @@ $(foreach lib,$(VALID_LIBRARIES),$(call include_lib,$(lib)))
 $(foreach proj,$(VALID_PROJECTS),$(call include_proj,$(proj)))
 
 IGNORE_CLEANUP_LIBS := CMSIS FreeRTOS STM32F0xx_StdPeriph_Driver unity FatFs
+IGNORE_PY_FILES := lint.py
+# Find all python files excluding library files in project env (./venv)
+FIND_PY_FILES:= $(shell printf "! -name %s " $(IGNORE_PY_FILES) | xargs find -path ./venv -prune -o -name '*.py')
+AUTOPEP8_CONFIG:= -a --max-line-length 100 -r
```
copperium, line 161: Could you also add the whole `unity` directory as ignored? It contains a bunch of python files which aren't *really* owned by us (they're part of the unity test library) and formatting them would probably cause merge conflicts.

(Also, please put a space before the `:=`)

copperium (changes requested):
I'm not sure what version of `make format` you formatted with, but it seems to have caused a *whole lot* of changes to headers and some changes that the current `make format` actually corrects, e.g. solar_config.c. (I'm pretty sure that autopep8 isn't supported to touch C files...) Could you use git to reset all files other than the Makefile, then (after adding `unity` as ignored) `make format` again? That way the whole repo is formatted like the current `make format` wants.

hzhou0:
I checked the Makefile history and can confirm I have not made any changes to the original c formatting code. It still seems to be formatting c files.

copperium:
Are you doing `make format` inside the vagrant box? I wonder if you have a different version of clang-format on your machine.

copperium:
You also need to add a `pip install --user autopep8` to the travis script in .travis.yml (by the pip install --user pylint) so that travis doesn't error because it can't find autopep8.

copperium (approved):

PR #278, opened 2021-02-12
for firmeware 102 hw

KevinLee5656:
Not sure how I would implement this code with the struct inside main because type Counters in the callback would show up as undefined for me.


projects/fw_102_hw/Midnight Sun.code-workspace
```diff
@@ -0,0 +1,11 @@
...
+	"folders": [
+		{
+			"path": "..\\.."
+		},
+		{
+			"path": "..\\..\\..\\..\\..\\..\\Sheertex\\iot-firmware"
+		}
+	],
+	"settings": {}
+}
```
copperium, line 11: Please don't commit the .code-workspace files VS Code generates; if it's a pain to avoid accidentally committing them then make a PR adding `*.code-workspace` to the .gitignore.

projects/fw_102_hw/Workspace.code-workspace
```diff
@@ -0,0 +1,14 @@
...
+		},
+		{
+			"path": "..\\..\\..\\..\\..\\..\\Sheertex\\iot-firmware"
+		},
+		{
+			"path": "..\\..\\..\\..\\..\\..\\..\\..\\Downloads\\MTE 140"
+		}
+	],
+	"settings": {}
+}
```
copperium, line 14: (Also don't commit this .code-workspace file)

projects/fw_102_hw/src/main.c
```diff
@@ -0,0 +1,36 @@
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+// not sure how to make it work with struct inside main()
+// would appreciate feedback!
```
copperium, line 7: Ah, by `In your main function, create a counters struct`, the homework just means to create a variable of your struct type inside main (which you've done with `count`), not put the struct definition inside, which indeed wouldn't work. So you've done it correctly!

projects/fw_102_hw/src/main.c
```diff
@@ -0,0 +1,36 @@
...
+#include "soft_timer.h"
+#include "wait.h"
+// not sure how to make it work with struct inside main()
+// would appreciate feedback!
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void counter_timer_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 13: Since this is a private function, please make it `static` and prefix its name with `prv_`. (The `static` makes it not visible from other files; otherwise there would be issues if another file tried to define a `counter_timer_callback`. The `prv_` is just a convention.)

projects/fw_102_hw/src/main.c
```diff
@@ -0,0 +1,36 @@
...
+  Counters *count = context;
+  count->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", count->counter_a);
+  if (count->counter_a % 2 == 0) {
+    count->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", count->counter_b);
+  }
+
+  soft_timer_start_millis(500, counter_timer_callback, count, NULL);
```
copperium, line 22: It's good practice to use a `#define`'d constant in place of the 500 here and below. That way, if we want to change the timer delay in the future, we only have to change it in one place. Make sure your constant has a descriptive name and ends with the unit (e.g. `_MS`) so we know what it is.

projects/hello_world/src/main.c
```diff
@@ -0,0 +1,7 @@
+#include "log.h"
+
+int main(void) {
+  LOG_DEBUG("Hello World!\n");
+
+  return 0;
+}
```
copperium, line 7: The `hello_world` project isn't part of the homework, so please don't commit it.

copperium (changes requested):
Nice job Kevin! Most of the comments are git issues, there's just a couple things to fix with the homework itself.

copperium:
LGTM, closing this since homework is complete

PR #279, opened 2021-02-13
SOFT-280: Thermistor reading modification to celsius

Cookieth:
I've talked to Liam Hawkins about the thermistor types, and here is the summary:

5 MPPT Board:
- TEMP 1 = RTD
- TEMP 2 = RTD
- TEMP 3 = NTC
- TEMP 4 = NTC
- TEMP 5 = NTC
- TEMP 6 = BJT

6 MPPT Board:
- TEMP 1 = RTD
- TEMP 2 = RTD
- TEMP 3 = NTC
- TEMP 4 = NTC
- TEMP 5 = NTC
- TEMP 6 = NTC

So, I've implemented changing the type of TEMP 6 to be dependent on `mppt_count` read from `Pin A7`.

Side note: I don't seem to be able to selectively commit things (I think the linting and formatting git hooks do a `git add --all` or something), so my commit messages don't match up very well with what is actually being done. I'm not sure if descriptive commits are desired in Midnight Sun, and so if they are please let me know so I can rebase.

**Tests**
As of now, `make test PLATFORM=x86 PROJECT=solar TEST=sense_temperature` fails:
```Running test_sense_temperature_runner
[0] libraries/x86/src/x86_interrupt.c:92: Main Thread (id:140041589992320)
projects/solar/test/test_sense_temperature.c:72:test_single_thermistor_cycle():FAIL: Expected 0 Was 19
[0] libraries/x86/src/x86_interrupt.c:92: Main Thread (id:140041589992320)
projects/solar/test/test_sense_temperature.c:107:test_max_thermistors_cycle():FAIL: Expected 2048 Was 0
[0] libraries/x86/src/x86_interrupt.c:92: Main Thread (id:140041589992320)
[0] projects/solar/test/test_sense_temperature.c:125: Testing ADC fault return value: there should be exactly one warning...
projects/solar/test/test_sense_temperature.c:128:test_adc_read_fail():FAIL: Expected FALSE Was TRUE
[0] libraries/x86/src/x86_interrupt.c:92: Main Thread (id:140041589992320)
projects/solar/test/test_sense_temperature.c:141:test_invalid_settings():PASS```

### CODE REVIEW

**Comment by Cookieth** [comment]
Looking into why tests are failing, and I can't figure it out. I may need some help there.

**Comment by copperium** [comment]
You need to change the mock in the test file to mock `adc_read_converted` instead of `adc_read_raw` - the `TEST_MOCK(adc_read_raw)` construction is basically saying "within the tests, any calls to `adc_read_raw` instead will call this function". Since you changed `adc_read_raw` to `adc_read_converted` in the module, you need to change the mock too.

**Comment by Cookieth** [comment]
I changed the test file to mock converted, but the same exact error is coming up. The first error `Expected 0 Was 19` seems to be an issue with getting ADC. I tried looking into it and it kind of hinted to me that it could perhaps be an issue with the way I created my struct/enum, but I don't know how it could be.

**Comment by copperium** [comment]
Oh sorry, you also need to change `adc_read_raw` to `adc_read_converted` in the `$(T)_test_sense_temperature_MOCKS` line in rules.mk. That line is how the build system knows you're trying to mock a function.

**Comment by Cookieth** [comment]
Made the suggested change, and I still get the error.

**Comment by copperium** [comment]
Now it looks like the tests are just failing because they check that the value stored in the data store is equal to the value returned by `adc_read_converted` (i.e. `TEST_ADC_RAW_READING`), but of course your change made it so that that assumption is invalid. You'll have to change the tests so that you check for the new correct values.

**Comment by Cookieth** [comment]
Ah, you're right! My bad, I must have been looking at the wrong test error result. I will look into this and modify it.

**Comment by Cookieth** [comment]
Hello! I am back from midterms. I've added signed/unsigned helper functions, and fixed the tests to account for the thermistor types. All should be good now. I just need to rebase this onto the newest master, and it should be all good to review.

File: projects/solar/test/test_sense_temperature.c [hunk 1]
```diff
@@ -12,6 +12,11 @@
 
 #define TEST_ADC_RAW_READING 0x800
 
+// The two test variables below are calculated from inputting 0x800 into the relative conversion
+// formulae (see sense_temperature.c)
+#define TEST_RTD_CONVERTED_READING -59
```
copperium, line 17: Nitpick: it's probably a good idea to wrap this in parentheses like `(-59)` so that e.g. `foo TEST_RTD_CONVERTED_READING` isn't accidentally valid (expanding to `foo - 59`).

projects/solar/test/test_sense_temperature.c
```diff
@@ -12,6 +12,11 @@
 
 #define TEST_ADC_RAW_READING 0x800
```
copperium, line 13: This is actually the `TEST_ADC_CONVERTED_READING` now eh? And for converted readings it's more idiomatic to give it in decimal since it's an actual value in mV (in this case 2048mV).

projects/solar/test/test_sense_temperature.c
```diff
@@ -72,21 +77,23 @@ void test_single_thermistor_cycle(void) {
...
     .num_thermistors = MAX_THERMISTORS,
   };
   // take the pins sequentially from PA0 on
   for (uint8_t thermistor = 0; thermistor < MAX_THERMISTORS; thermistor++) {
-    settings.thermistor_pins[thermistor].port = thermistor / GPIO_PINS_PER_PORT;
-    settings.thermistor_pins[thermistor].pin = thermistor % GPIO_PINS_PER_PORT;
+    settings.thermistor_settings[thermistor].thermistor_address.port =
+        thermistor / GPIO_PINS_PER_PORT;
+    settings.thermistor_settings[thermistor].thermistor_address.pin =
+        thermistor % GPIO_PINS_PER_PORT;
```
copperium, line 96: You should also set the `thermistor_type` to various values, maybe with a `thermistor % NUM_THERMISTOR_TYPES` expression.

projects/solar/src/data_store.c
```diff
@@ -22,6 +22,15 @@ StatusCode data_store_set(DataPoint data_point, uint32_t value) {
   return STATUS_CODE_OK;
 }
 
+StatusCode data_store_set_signed(DataPoint data_point, int32_t value) {
```
copperium, line 25: You should add tests for these in test_data_store.c that just make sure they work correctly - we want to make sure there are no weird conversion errors.

copperium (changes requested):
Great job Keith! There seems to be something strange going on with the unity library; it should go away when you rebase/merge master though.

projects/solar/src/sense_temperature.c
```diff
@@ -9,21 +10,54 @@
...
+      // TODO(SOFT-280): Continuation, T = ΔV * q / (n * k * ln(10)), get two readings from fan and
+      // convert readings.
+      break;
+    case NUM_THERMISTOR_TYPES:
+      break;
+  }
+
   if (status_ok(status)) {
-    data_store_set(data->data_point, raw_reading);
+    data_store_set_signed(data->data_point, converted_reading);
```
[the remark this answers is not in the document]
Cookieth: To return in deciCelsius, wouldn't I need to divide by 10?

projects/solar/test/test_sense_temperature.c
Cookieth: I'm a bit confused on what the change requested here is.

projects/solar/src/data_store.c
Cookieth: How is that different from the tests already put in place that converts `0x800` to `-59` and `28` respective to the thermistor type?

projects/solar/test/test_sense_temperature.c
copperium: Ah sorry, what I meant was change this variable to `TEST_ADC_CONVERTED_READING` since it's returned from `adc_read_converted` now, not `adc_read_raw`, and specify its value in decimal rather than hex because it's an actual value in mV.

projects/solar/src/data_store.c
copperium: The tests in test_sense_temperature.c would indirectly test this function through converting 0x800 to -59 or 28 yes, but we shouldn't rely on the sense_temperature tests to test data_store - the data_store tests should test all the functionality in that module, and other modules' tests should then just assume that data_store works. Though it looks like you've removed the signed functions so I suppose it's a moot point.

copperium (approved):
Looks good! If you want you can push a commit to change the one macro name, otherwise I'll just merge it.

Cookieth:
One last thing- there seems to be a merge conflict with the platform.mk, and I just wanted to check what change I should be making:
```
<<<<<<< soft_280_thermistor_reading
# Linker flags
LDFLAGS := -L$(LDSCRIPT_DIR) -Tstm32f0.ld -Wl,--gc-sections -Wl,--undefined=uxTopUsedPriority \
           --specs=nosys.specs --specs=nano.specs -lm
=======
# Linker flags - linker script set per target
LDFLAGS := -L$(LDSCRIPT_DIR) -Wl,--gc-sections -Wl,--undefined=uxTopUsedPriority \
           --specs=nosys.specs --specs=nano.specs
>>>>>>> master
```

Since I do need the -lm flag for math.c, should this look like:
```
LDFLAGS := -L$(LDSCRIPT_DIR) -Wl,--gc-sections -Wl,--undefined=uxTopUsedPriority \
           --specs=nosys.specs --specs=nano.specs -lm
```

I forget if -Tstm32f0.ld was added for a specific reason or not. Thanks!

copperium:
Yep, what you wrote looks like. I believe `-Tstm32f0.ld` causes it to read the `stm32f0.ld` linker script, which was at the least renamed on master - I don't think you need it.

PR #280, opened 2021-02-13
FM 103 HW

PurpleRhythms:
Adc reading with button interrupt.

### DIFF
--- a/projects/adc_reading/src/main.c
+++ b/projects/adc_reading/src/main.c
@@ -7,30 +7,14 @@
 #include "log.h"
 #include "soft_timer.h"
 
-// Packages context for callback function outside main
-typedef struct Context_package {
-  GpioAddress button_addr;
-  InterruptSettings button_interrupt_settings;
-  AdcChannel adc_reading_channel;
-} Context_package;
-
 static void prv_interrupt_callback(const GpioAddress *address, void *context) {
   // casting void* back for use outside main()
-  Context_package *p_package = context;
-
-  // Unpacking context package structure
-  GpioAddress button_addr = p_package->button_addr;
-  InterruptSettings button_interrupt_settings = p_package->button_interrupt_settings;
-  AdcChannel adc_reading_channel = p_package->adc_reading_channel;
+  AdcChannel *adc_reading_channel = context;
 
   // read and print adc reading data
   uint16_t adc_reading_data = 0;
   adc_read_converted(adc_reading_channel, &adc_reading_data);
   LOG_DEBUG("ADC Reading Data: %d\n", adc_reading_data);
-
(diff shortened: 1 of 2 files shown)


projects/adc_reading/src/main.c
```diff
@@ -0,0 +1,87 @@
...
+  AdcChannel adc_reading_channel = p_package->adc_reading_channel;
+
+  // read and print adc reading data
+  uint16_t adc_reading_data = 0;
+  adc_read_converted(adc_reading_channel, &adc_reading_data);
+  LOG_DEBUG("ADC Reading Data: %d\n", adc_reading_data);
+
+  // restart interrupt watch
+  gpio_it_register_interrupt(&button_addr, &button_interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             prv_interrupt_callback, p_package);
```
copperium, line 33: You don't actually need to do this, the gpio interrupt never gets cleared. Just the `gpio_it_register_interrupt` in main will call this function whenever the interrupt occurs for the rest of the program (ie until the board is reflashed). (It's not like soft timers where if you want a loop you have to start a timer again.)

This means you don't need to pass around the `button_interrupt_settings` or the `button_addr` either.

projects/adc_reading/src/main.c
```diff
@@ -0,0 +1,87 @@
...
+#include "interrupt_def.h"
+#include "log.h"
+#include "soft_timer.h"
+
+// Packages context for callback function outside main
+typedef struct Context_package {
+  GpioAddress button_addr;
+  InterruptSettings button_interrupt_settings;
+  AdcChannel adc_reading_channel;
+} Context_package;
```
copperium, line 15: Please use UpperCamelCase for struct names: `ContextPackage`.

projects/adc_reading/src/main.c
```diff
@@ -0,0 +1,87 @@
...
+    GPIO_DIR_IN,
+    GPIO_STATE_HIGH,
+    GPIO_RES_PULLDOWN,
+    GPIO_ALTFN_ANALOG,
+  };
+
+  InterruptSettings button_interrupt_settings = {
+    INTERRUPT_TYPE_INTERRUPT,
+    INTERRUPT_PRIORITY_NORMAL,
+  };
```
copperium, line 62: The gnu struct initialization style is preferred so you don't have to worry about the order of parameters: `.direction = GPIO_DIR_IN`, `.resistor = GPIO_RES_PULLDOWN`, etc.

copperium (changes requested):
Looks great, just one conceptual comment and a couple of style nits.

copperium:
Looks great! Closing since homework is complete.

PR #281, opened 2021-02-14
[SOFT-392] Babydriver: gpio interrupts python implementation

sudhish3120:


projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+        port = data[1]
+        pin = data[2]
+        edge = data[3]
+        # Calling function from callback_dict based on (port, pin)
+        try:
+            # ret is only used for testing purposes
+            ret = callback_dict[(port, pin)]((port, pin, edge))
+        except TypeError as type_error:
+            raise "Callback function parameters are of incorrect format (use a named-tuple called \
+                    info to store port, pin and edge" from type_error
```
copperium, line 45: You can't raise strings in Python (that's only JS afaik), so you'll need to wrap this in an exception type like `ValueError` or `TypeError`. (The `raise ... from` syntax only sets the cause of the new error to the first error.)

Also when breaking Python strings over multiple lines, it's good practice to do it like this:
```py
a = ("some really really long string"
     " which is really really long")
```
Your current method with a backslash escape actually includes all the indentation in the second line in the string, so the string actually has a ton of whitespace in the middle.

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+        unfiltered_msg.data[0] == message_defs.BabydriverMessageId.GPIO_IT_INTERRUPT):
+
+        data = unfiltered_msg.data[:]
+        port = data[1]
+        pin = data[2]
+        edge = data[3]
+        # Calling function from callback_dict based on (port, pin)
+        try:
+            # ret is only used for testing purposes
+            ret = callback_dict[(port, pin)]((port, pin, edge))
```
copperium, line 42: By a named tuple I actually meant Python's `namedtuple` feature from the `collections` library: https://docs.python.org/3/library/collections.html#collections.namedtuple. If you define a namedtuple at the top level like `GpioItInfo = namedtuple("GpioItInfo", ["port", "pin", "edge"])` then you can instantiate it like `GpioItInfo(port, pin, edge)` and access it like `info.port`, `info.pin`, `info.edge`.

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+    with the interrupt
+    """
+
+    unfiltered_msg = can_util.Message.from_msg(can_message)
+    # print(f"{unfiltered_msg.message_id}, {unfiltered_msg.data[0]},
+    # {unfiltered_msg.data[1]}, {unfiltered_msg.data[1]}")
+    if (unfiltered_msg.message_id == message_defs.BABYDRIVER_CAN_MESSAGE_ID and
+        unfiltered_msg.data[0] == message_defs.BabydriverMessageId.GPIO_IT_INTERRUPT):
+
+        data = unfiltered_msg.data[:]
```
copperium, line 35: Is there a reason you copy the data with `[:]` instead of just assigning `data` as an alias of `unfiltered_msg.data` like `data = unfiltered_msg.data`?

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+        # Calling function from callback_dict based on (port, pin)
+        try:
+            # ret is only used for testing purposes
+            ret = callback_dict[(port, pin)]((port, pin, edge))
+        except TypeError as type_error:
+            raise "Callback function parameters are of incorrect format (use a named-tuple called \
+                    info to store port, pin and edge" from type_error
+
+
+    return ret
```
copperium, line 48: When the if statement isn't hit, `ret` isn't defined, so this line will throw a `NameError`.

However, I don't think this function should return a value at all - there's no need to, and you shouldn't return random things from your functions for the sake of your tests. See my comments on the tests.

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+    status_message = can_util.next_message(babydriver_id=message_defs.BabydriverMessageId.STATUS)
+    received_status = status_message.data[1]
+
+    # Check if status is invalid (0 refers to STATUS_CODE_OK)
+    if received_status != 0:
+        raise Exception("Received a nonzero STATUS_CODE: {}".format(received_status))
+
+    # Clearing callback related to the interrupt that was unregistered
+    if (port, pin) not in callback_dict:
+        raise KeyError(f"No interrupt registered on given port and pin {callback_dict}")
```
copperium, line 180: I can't guarantee that all the Python versions we'll run this on are 3.6+ and support f-strings, so could you replace all your f-strings with `.format` instead?

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,182 @@
...
+
+#  Setting up notifier to listen for all CAN messages
+def init_notifier_gpio_it():
+    """ Imports notifier from can_util and adds callback_listener
+        to list of listeners the notifier calls upon receiving a can message """
+    notifier = can_util.get_bus_notifier()
+    notifier.add_listener(callback_listener)
+
+
+def register_gpio_interrupt(port, pin, edge = InterruptEdge.RISING, callback = None):
```
copperium, line 67: I think `RISING_AND_FALLING` should be the default edge, since it gives the most information (and it's also the slowest to type).

projects/baby_driver/scripts/test_gpio_interrupts.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        Tests callback_listener by checking if it can trigger default/user-defined callbacks
+        upon receving can messages and tests fail conditions
+        """
+
+        def test_user_callback(info):
+            return f"Test callback, port:{info[0]}, pin:{info[1]}, edge:{info[2]}"
+
+        def incorrect_test_user_callback(port, pin, edge):
+            return f"Incorrect Test callback, {port}, {pin}, {edge}"
```
copperium, line 151: Instead of returning values from the callbacks to test, why don't you set some properties on `self` like you do with the `parameter_test` mock above? That way you don't have to bend over backwards in your module to accommodate the tests.

projects/baby_driver/scripts/test_gpio_interrupts.py
```diff
@@ -0,0 +1,342 @@
...
+                                       device_id = BABYDRIVER_DEVICE_ID)
+            return ret_msg
+
+        mock_convert_can_msg.side_effect = test_msg_converter
+
+        callback_dict.clear()
+        callback_dict[(0,0)] = test_user_callback
+        callback_dict[(5,3)] = test_user_callback
+        callback_dict[(2,15)] = test_user_callback
+        callback_dict[(5,15)] = test_user_callback
```
copperium, line 164: The fact that a global `callback_dict` is used internally to store the callbacks is an implementation detail, and you shouldn't be testing implementation details. Instead, register these callbacks through `register_gpio_interrupt`. 

(This also goes for the part below marked `Testing fail condition`.)

copperium (changes requested):
This is great Sudhish, I love how extensive your tests are!

If you merge master then `make format` will run autopep8 over your code (and this'll be needed to pass CI once it gets reenabled).

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,186 @@
...
+            callback_dict[(port, pin)](info)
+        except TypeError as type_err:
+            raise TypeError ("Callback function parameters are of incorrect format (use a "
+                             "named tuple called info to store port, pin and edge)") from type_err
+
+
+def default_callback(info):
+    """ The function called after occurence of interrupt if
+       no user-defined callback function exists"""
+    info.port = chr(info.port + ord('A'))
```
copperium, line 59: When this line executes, you'll get the error:
> AttributeError: can't set attribute

This is because you can't update the fields of a namedtuple like this; instead, just use a new variable.

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,186 @@
...
+    print("Interrupt on P{}{}: {}".format(info.port, info.pin, info.edge))
```
copperium, line 60: `info.edge` here will be printed as a number, like `Interrupt on PA0: 1`, which isn't terribly helpful. I think it would be better to print `rising` or `falling` as appropriate.

projects/baby_driver/scripts/gpio_interrupts.py
```diff
@@ -0,0 +1,186 @@
...
+    if pin < 0 or pin >= NUM_PINS_PER_PORT:
+        raise ValueError("Invalid GPIO pin number (Range: 0 - 15)")
+
+    if isinstance(edge, str):
+        if edge.upper() not in InterruptEdge.__members__:
+            raise AttributeError("Enter 'RISING', 'FALLING' or 'RISING_AND_FALLING' for "
+                                 "interrupt edge")
+        edge = InterruptEdge[edge.upper()]
+
+    if (edge < 0 or edge >= InterruptEdge.NUM_INTERRUPT_EDGES):
```
copperium, line 114: Nit: you don't need the parentheses here

copperium (changes requested):
Just a couple things to fix, then we should be good to merge!

If you want to test this, I encountered a few bugs, so do the following:
* make a new branch like `git checkout -b gpio-interrupts-test`
* git merge `soft_440_watchdog_bug` and `soft_441_x86_adc_bug` to fix some issues I found while trying to test it
* in projects/baby_driver/src/main.c, move the `interrupt_init` call to the top of `main()` - it has to be initialized before `gpio_it`. (Actually, could you do this on this branch to include it in this PR? It's a bug introduced in the firmware gpio interrupts side.)

Then testing in `make babydriver PLATFORM=x86` should work.

copperium (approved):

PR #282, opened 2021-02-14
SOFT-429: Added auto-generated README.md

Cookieth:
The README.md file comes with the following template:
```
<!--
        General guidelines
        These are just guidelines, not strict rules - document however seems best.
        A README for a firmware-only project (e.g. Babydriver, MPXE, bootloader, CAN explorer) should answer the following questions:
            - What is it?
            - What problem does it solve?
            - How do I use it? (with usage examples / example commands, etc)
            - How does it work? (architectural overview)
        A README for a board project (powering a hardware board, e.g. power distribution, centre console, charger, BMS carrier) should answer the following questions:
            - What is the purpose of the board?
            - What are all the things that the firmware needs to do?
            - How does it fit into the overall system?
            - How does it work? (architectural overview, e.g. what each module's purpose is or how data flows through the firmware)
    -->
    # $name
```

Anything enclosed between the  tags won't be shown by most markdown readers, and the $name is a variable passed in which is the project's name.


copperium (approved):

copperium:
I'm going to hold off on merging until we get CI reenabled (since last time I merged something without CI `make test_format` started complaining), but this is amazing!

copperium:
Looks like autopep8 (which just recently got enabled) is complaining about something to do with the docstring, could you run `make format`?

PR #285, opened 2021-02-17
Soft 999 {kevin lee} fw 103

KevinLee5656:
25823590th time's the charm :'


projects/buttons_and_adcs/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+// your lead)
+#include <stdbool.h>
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "interrupt_def.h"
+#include "log.h"
+
+void prv_button_interrupt_handler(const GpioAddress *address, void *context) {
```
copperium, line 18: This function should be `static` since it's private to this file.

projects/buttons_and_adcs/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+  uint16_t adc_pin_data = 0;
+  GpioAddress *adc_pin_addr = (GpioAddress *)context;
+  adc_read_raw_pin(*adc_pin_addr, &adc_pin_data);
```
copperium, line 21: The homework does specify a *converted* reading: that means use `adc_read_converted_pin` instead of `adc_read_raw_pin`. The difference is that `adc_read_converted_pin` returns a value in millivolts, whereas `adc_read_raw_pin` returns whatever raw value the ADC spits out - you almost always want a converted read.

copperium (changes requested):
Really sorry for the delay, just a couple of nitpicks.

copperium:
LGTM, closing since homework is complete.

PR #287, opened 2021-02-17
[SOFT-345] Add mpxe project startup conditions

mitchellostler:
Ok, init conditions are finally working lol. For some reason the build env was being really finicky and causing the tests to fail, but now with the changes, if you pull from this branch and run `make clean` and then `make mpxe` they should all pass. (Although if you could check that would be great for my sanity)

Here's what the changes are:

- pm.start() can take tuple containing necessary information to write to store
- Init_cond.py stores functions used to create initial conditions, mcp23008 can be seen as an example, this format should be able to be used by any driver
- upon store init in driver, stdin will be checked for input of initial conditions once
-  Polling thread in store.c will block until SIGRTMIN received from pm

In addition:
- Makefile has been updated to take an argument for init conditions. An implementation similar to that in test_mcp23008.py will be needed for each driver needing init conditions
- Driver specific implementations will be needed for processing each store

Once PR is merged I will add to the mpxe wiki for how it works and what will be needed for driver specific versions.


Makefile
```diff
@@ -11,6 +11,7 @@
 #   PB: [PROBE=] - Specifies which debug probe to use on STM32F0xx. Defaults to cmsis-dap [cmsis-dap | stlink-v2].
 #   DF: [DEFINE=] - Specifies space-separated preprocessor symbols to define.
 #   CH: [CHANNEL=] - Specifies the default CAN channel for Babydriver. Defaults to vcan0 on x86 and can0 on stm32f0xx.
+#   IC: [INIT_COND=] - Specifies whether initial conditions to be used for MPXE tests
```
mitchellostler, line 14: I will add that TRUE needs to be passed as an argument here

libraries/ms-drivers/src/adt7476a_fan_controller.c
```diff
@@ -114,6 +116,7 @@ StatusCode adt7476a_get_status(I2CPort port, uint8_t adt7476a_i2c_read_address,
 StatusCode adt7476a_init(Adt7476aStorage *storage, Adt7476aSettings *settings) {
 #ifdef MPXE
   prv_init_store();
+  mpxe_initial_conditions = read_init_conditions();
```
mitchellostler, line 119: This call to read_init_conditions is needed to keep tests from failing

mpxe/integration_tests/init_cond.py
```diff
@@ -0,0 +1,15 @@
...
+    mcp23008_msg             = mpxe.protogen.mcp23008_pb2.MxMcp23008Store()
+    mcp23008_msg.state.extend([0] * 8)
+    for i in range(NUM_MCP_PINS):
+        mcp23008_msg.state[i]       =  i % 2
+        mcp23008_mask            = mpxe.protogen.mcp23008_pb2.MxMcp23008Store()
+        mcp23008_mask.state.extend([0] * 8)
+    for i in range(NUM_MCP_PINS):
+        mcp23008_mask.state[i]      = 1
+    startup_cond = (mcp23008_msg, mcp23008_mask, stores_pb2.MxStoreType.MCP23008)
+    return startup_cond
```
copperium, line 15: Not sure what's up with the formatting here, could you fix it? (though it will be fixed when we run autopep8 on it after merging to master anyways)

copperium (changes requested):
I'll just summarize my thoughts here instead of leaving comments.

The locking sequencing looks like it works great! I'm not too sure about doing it per driver though; that seems sketchy because it a) requires updating every driver and b) is super order dependent. With the system of relying on the `read_init_conditions` return value, we'd only be able to add initial values to the first initial conditions-enabled driver that's initialized in the library (I think that's how it works).

I also notice that the pm.py `startup_messages` API only lets us send one initial condition - we should have it take a list so that we can send multiple initial conditions, in case we want to initialize multiple drivers.

I think we can make this more user friendly by moving the initial conditions logic more to store.c. Essentially, I'm thinking that in `store_config`, we receive all the iniital conditions and store them in a map-type data structure mapping (store type, key) => initial store/mask/etc (or whatever structure is more appropriate). Then, in `store_register`, we check if we have an initial condition for the type/key and if so we call the update function. Then the only change we'd have to make in the drivers is to move `init_store`/`store_config`/`store_register` calls to after the driver's default settings are initialized, which I think is okay if documented. (The mapping stuff might need `malloc`/`free`, but that's alright on x86.)

This way we wouldn't have to expose any of the `read_init_conditions` stuff to the driver and we wouldn't have to change every single driver. Let me know if you think this is a good idea! (also @g-jessmuir)

Also, I'm not really sure what the purpose of the `INIT_COND` makefile variable is. Wouldn't some (real) integration tests always need initial conditions? I can't really imagine a scenario other than your test mcp23008 example where you'd need to switch between initial conditions and no initial conditions in a test, and if you wanted to you could just not pass initial conditions to `start()`. I think you can remove it.

libraries/mpxe-store/src/store.c
```diff
@@ -45,6 +48,10 @@ static void prv_sigusr2(int signo) {
   pthread_mutex_unlock(&s_log_lock);
 }
 
+static void prv_sigrtmin(int signo) {
```
g-jessmuir, line 51: lol isn't this gonna conflict with the gpio interrupts since they also use sigrt and use the full range of 1-32? maybe just bump this to SIGIO, what do you think?

g-jessmuir (commented):
I agree with Ryan's points.

One of the main design goals was to minimize how much actual driver / smoke test / project code we have to change (ideally zero non-driver code), but it does seem like we should be able to move things back into store.c, I'm open to looking at this together if you want to set up some time

PR #288, opened 2021-02-19
[SOFT-434] Add interrupt function to PCA9539R driver

hzhou0:


libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -81,3 +81,14 @@ StatusCode pca9539r_gpio_get_state(const Pca9539rGpioAddress *address,
   *input_state = s_pin_settings[address->i2c_address][address->pin].state;
   return STATUS_CODE_OK;
 }
+
+StatusCode pca9539r_gpio_subscribe_interrupts(const GpioAddress *interrupt_pin,
+                                              const Pca9539rInterruptCallback callback,
+                                              void *context) {
+  gpio_it_init();
```
copperium, line 88: Libraries like `gpio_it` can generally only be initialized once per program, or else they do nasty things like resetting all the registered interrupts. Because of this, we leave initializations to the `main()` function in each project, and we tell projects to initialize the relevant libraries using a comment at the top of the header file (like `// Requires ... to be initialized.`).

So remove this line, and add a line at the top of the header underneath the `// Requires I2C to be initialized.` line that says something about how if you want to subscribe to interrupts, you must initialize `gpio_it`.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -81,3 +81,14 @@ StatusCode pca9539r_gpio_get_state(const Pca9539rGpioAddress *address,
   *input_state = s_pin_settings[address->i2c_address][address->pin].state;
   return STATUS_CODE_OK;
 }
+
+StatusCode pca9539r_gpio_subscribe_interrupts(const GpioAddress *interrupt_pin,
+                                              const Pca9539rInterruptCallback callback,
```
copperium, line 86: The callback is declared as `const` here but not in the header file - it doesn't really matter whether it's const or not, but please be consistent.

copperium (changes requested):
Nice job! Please add tests to the `test_pca9539r_gpio_expander.c` file in libraries/ms-common/test to demonstrate that your function works. You can find examples of tests in any other file in a test/ directory.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
hzhou0: My IDE is telling me when I qualify the callback as const: 
Clang-Tidy: Parameter 'callback' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions

mitchellostler: I looked into this and there is a thread that talks more about this [here](https://stackoverflow.com/questions/51353648/const-qualification-of-parameters-in-function-declaration) if you're interested. You shouldn't need the const qualifier on the callback param though you can just get rid of it

libraries/ms-drivers/inc/pca9539r_gpio_expander.h
```diff
@@ -1,8 +1,9 @@
 #pragma once
 // GPIO HAL interface for the PCA9539RPW/Q900J GPIO expander.
-// Requires I2C to be initialized.
+// Requires I2C , GPIO_IT to be initialized.
```
copperium, line 3: We have a bunch of projects which use the PCA9539R driver already, but they don't all need to be updated to initialize GPIO interrupts. So instead of adding to the `Requires I2C to be initialized` line, add another comment below it and say that *if* you want to subscribe to interrupts, you must initialize gpio interrupts.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -1,5 +1,5 @@
 #include "pca9539r_gpio_expander.h"
-
+#include "status.h"
```
copperium, line 2: Nitpick: the "include order" item in our [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) dictates that there should be a blank line between the #include of this C file's header file (here `#include "pca9539r_gpio_expander.h"`) and the rest of the includes, and there should be a blank line between the last include and the rest of the file.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -81,3 +81,13 @@ StatusCode pca9539r_gpio_get_state(const Pca9539rGpioAddress *address,
...
 }
+#include "log.h"
+StatusCode pca9539r_gpio_subscribe_interrupts(const GpioAddress *interrupt_pin,
+                                              Pca9539rInterruptCallback callback, void *context) {
+  InterruptSettings interrupt_settings = { .type = INTERRUPT_TYPE_INTERRUPT,
+                                           .priority = INTERRUPT_PRIORITY_NORMAL };
+  gpio_it_register_interrupt(interrupt_pin, &interrupt_settings, INTERRUPT_EDGE_FALLING, *callback,
+                             context);
+  LOG_DEBUG("%s", status_get().message);
+  return status_get().code;
```
copperium, line 92: I'm not sure why you're logging `status_get().message`? Just returning the `gpio_it_register_interrupt` call is enough to forward any errors.

libraries/ms-drivers/src/x86/pca9539r_gpio_expander.c
```diff
@@ -81,3 +81,13 @@ StatusCode pca9539r_gpio_get_state(const Pca9539rGpioAddress *address,
   *input_state = s_pin_settings[address->i2c_address][address->pin].state;
   return STATUS_CODE_OK;
 }
+#include "log.h"
```
copperium, line 84: You shouldn't need log.h, but if you do, its include should be at the top with the status.h include, not down here.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
...
+void test_pca9539r_gpio_subscribe_interrupts(void) {
+  GpioAddress address = { .pin = VALID_PORT_0_PIN, .port = 0 };
+  TEST_ASSERT_OK(pca9539r_gpio_subscribe_interrupts(&address, &prv_test, NULL));
+}
+
+void test_pca9539r_gpio_subscribe_interrupts_invalid(void) {
+  GpioAddress invalid_address = { .pin = INVALID_GPIO_PIN, .port = 1 };
+  TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS,
+                    pca9539r_gpio_subscribe_interrupts(&invalid_address, NULL, NULL));
+}
```
copperium, line 326: General style nitpicks: there should be a blank line between every pair of functions and a blank line at the end of the file.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
   Pca9539rGpioState state;
   TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, pca9539r_gpio_get_state(&address, &state));
 }
+void prv_test(const struct GpioAddress *address, void *context) {
+  GpioState state = GPIO_STATE_LOW;
+  gpio_get_state(address, &state);
+}
+void test_pca9539r_gpio_subscribe_interrupts(void) {
+  GpioAddress address = { .pin = VALID_PORT_0_PIN, .port = 0 };
```
copperium, line 318: `VALID_PORT_0_PIN` is #define'd to refer to a pin on the PCA9539R gpio expander, rather than the normal GPIO of the STM32 that GpioAddresses refer to. I'd just use `0` as a generic "valid gpio pin". Also, you should be using `GPIO_PORT_A`, `GPIO_PORT_B`, etc (from gpio_mcu.h) as port values instead of literals like `0`.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
   Pca9539rGpioState state;
   TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, pca9539r_gpio_get_state(&address, &state));
 }
+void prv_test(const struct GpioAddress *address, void *context) {
+  GpioState state = GPIO_STATE_LOW;
+  gpio_get_state(address, &state);
+}
+void test_pca9539r_gpio_subscribe_interrupts(void) {
```
copperium, line 317: This test is a good start, but you should do more than just test that `pca9539r_gpio_subscribe_interrupts` returns OK; you should test that it actually does everything it's supposed to. I think a good way to do that would be to check that the callback you register is called when you trigger an interrupt on the pin with `gpio_it_trigger_interrupt`.

A way to do that would be to have a static variable (maybe a `uint8_t s_times_interrupt_callback_called`), reset it to 0 in `setup_test`, and increment it in your callback. Then in this test you can call `gpio_it_trigger_interrupt` and test that the callback was called by testing that your static variable was incremented.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
   Pca9539rGpioState state;
   TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, pca9539r_gpio_get_state(&address, &state));
 }
+void prv_test(const struct GpioAddress *address, void *context) {
+  GpioState state = GPIO_STATE_LOW;
+  gpio_get_state(address, &state);
+}
```
copperium, line 316: A few points about this function:
* The name `prv_test` doesn't really describe what its purpose is, which is to be a callback for subscribing to interrupts. Please change its name to reflect its purpose.
* It's a private function, so it should be `static`.
* You don't need the `struct` keyword before `GpioAddress`; we do the `typedef struct GpioAddress { ... } GpioAddress` construct so you don't have to say `struct` everywhere.
* The contents of the function aren't really doing anything useful; see my previous comment for what to replace them with.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
   Pca9539rGpioState state;
   TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, pca9539r_gpio_get_state(&address, &state));
 }
+void prv_test(const struct GpioAddress *address, void *context) {
```
copperium, line 313: Nitpick: the convention in this file (not everywhere, just in this file) is to separate tests for each function with a comment like `// pca9539r_gpio_get_state` at the start of the `pca9539r_gpio_get_state` tests; please conform to this.

libraries/ms-drivers/test/test_pca9539r_gpio_expander.c
```diff
@@ -308,3 +310,17 @@ void test_pca9539r_gpio_get_state_invalid_address(void) {
...
+  GpioState state = GPIO_STATE_LOW;
+  gpio_get_state(address, &state);
+}
+void test_pca9539r_gpio_subscribe_interrupts(void) {
+  GpioAddress address = { .pin = VALID_PORT_0_PIN, .port = 0 };
+  TEST_ASSERT_OK(pca9539r_gpio_subscribe_interrupts(&address, &prv_test, NULL));
+}
+
+void test_pca9539r_gpio_subscribe_interrupts_invalid(void) {
+  GpioAddress invalid_address = { .pin = INVALID_GPIO_PIN, .port = 1 };
```
copperium, line 323: Like above, `INVALID_GPIO_PIN` is defined for an invalid _pca9539r_ pin (I admit it's kind of confusing); instead use `GPIO_PINS_PER_PORT` for an invalid pin, or use `NUM_GPIO_PORTS` for an invalid port.

copperium (changes requested):
Nice start on the tests, see comments for ideas on how to improve them.

Eddy-M-K (commented):
Are all the changes to the Python files normal, or did I do something wrong?

copperium:
> Are all the changes to the Python files normal, or did I do something wrong?

The Python changes should be fine (I guess autopep8 wasn't run over some of the files), but it did cause one merge conflict; use `git pull origin master && git merge master` and then resolve the conflict and commit and push it. The conflict should be in make/new_target.py, choose the nicely aligned version.

copperium:
Now if you look in the checks, CI caught a compile error - please fix that and make sure CI passes.

Frank-Yan:
Some parts may fail because I did force the commits and pushes with --no-verify since I kept running into errors with the commit and push commands.

ryandancy (approved):

ryandancy:
Looks like CI caught a syntax error again, could you fix? @Frank-Yan

ryandancy:
Still failing @Frank-Yan, for a different reason this time it looks like

PR #290, opened 2021-02-21
[SOFT-301] Fix race condition in x86 CAN library

copperium:
Required rewriting power_distribution's publish_data, but hopefully tests should stand half a chance of passing now.


projects/power_distribution/src/publish_data.c
```diff
@@ -19,17 +30,42 @@ StatusCode power_distribution_publish_data_init(PowerDistributionPublishConfig c
   return STATUS_CODE_OK;
 }
 
+static void prv_partially_publish(SoftTimerId timer_id, void *context) {
+  if (s_current_measurements == NULL) {
```
mitchellostler, line 34: This shouldn't be needed right, since there will always be a sized array created in power_distribution_publish_data_publish()? Is there a reason there would be a NULL condition that I am missing?

projects/power_distribution/src/publish_data.c
```diff
@@ -19,17 +30,42 @@ StatusCode power_distribution_publish_data_init(PowerDistributionPublishConfig c
...
+  }
+
+  // We use a uintptr_t for the index so we can pass it through the context
+  uintptr_t index = (uintptr_t)context;
+
+  uint16_t transmit_up_to = MIN(index + NUM_MSG_PER_ITERATION, s_config.num_currents_to_publish);
+  while (index < transmit_up_to) {
+    PowerDistributionCurrent current_id = s_config.currents_to_publish[index];
+    StatusCode tx_code = s_config.transmitter(current_id, s_current_measurements[current_id]);
+    if (!status_ok(tx_code)) {
```
mitchellostler, line 46: Is there any other error handling needed here? The previous method called status_ok_or_return on each tx. There isn't any handling in main.c set up right now i guess so this shouldn't be an issue, but would it be a good idea to maybe re-transmit?

projects/power_distribution/src/publish_data.c
```diff
@@ -19,17 +30,42 @@ StatusCode power_distribution_publish_data_init(PowerDistributionPublishConfig c
...
   }
 
-  // transmit each current from the config
-  for (uint16_t c = 0; c < s_config.num_currents_to_publish; c++) {
-    PowerDistributionCurrent current_id = s_config.currents_to_publish[c];
-    status_ok_or_return(s_config.transmitter(current_id, current_measurements[current_id]));
-  }
-
+  s_current_measurements = current_measurements;
+  uintptr_t index = 0;
```
mitchellostler, line 68: What are the benefits of using a uintptr as opposed to a uint8_t by reference? Is it just so you don't have to cast it to use it?

mitchellostler (approved):
Thanks for resolving this Ryan, just some general questions but looks good to me. I see that the fix for power_distro_publish was making it so that it can only tx 8 messages at a time with a delay in between. Just out of curiousity, did you figure out why they were getting lost?

projects/power_distribution/src/publish_data.c
copperium: If we used a uint8_t by reference we'd have to either malloc it (possible memory leaks) or make it a static variable, since it can't live on the stack since any function might return before the soft timer is called. If it's a static variable, then if we called `power_distribution_publish_data_publish` twice before the all the currents were tx'd, the counter might get messed up and messages might not get sent. With a uintptr_t we can just stuff the counter into the pointer itself and avoid these issues.

projects/power_distribution/src/publish_data.c
copperium: You're right it shouldn't be null normally, but I like to defensively guard against it anyways - e.g. what if somebody passes NULL for `current_measurements` into `power_distribution_publish_data`, or somebody decides to reset the pointer to NULL at the end of TXing everything down the line. My reasoning was mostly that dereferencing a pointer that's null at some point in the program makes me nervous 😰

projects/power_distribution/src/publish_data.c
copperium: The messages aren't critical and it'll be retransmitted in the next cycle in 500ms anyways, so I figure it's not that big a deal. I prefer this version to the old one anyways - the previous version silently ignored errors and just didn't tx the rest of the currents, whereas this one at least tries and alerts us.

copperium:
@mitchellostler It turns out what was happening was that publish_data is txing 22 messages each call, which exhausted the event queue (size 20) but not the CAN TX queue (size 32). The behaviour that we patched out in can.c was saving it by regenerating the lost events, so when it was patched out `MS_TEST_HELPER_CAN_TX` would spin endlessly waiting for the 21st TX event.

PR #291, opened 2021-02-21
[SOFT-381] mcp3427 mpxe support

anthony-chang:


libraries/ms-drivers/src/x86/mcp3427_adc.c
```diff
@@ -64,8 +106,18 @@ static void prv_raise_ready(SoftTimerId timer_id, void *context) {
 
 static void prv_channel_ready(struct Fsm *fsm, const Event *e, void *context) {
   Mcp3427Storage *storage = (Mcp3427Storage *)context;
+#ifdef MPXE
+  s_store.readings[storage->current_channel] = FIXED_RESULT;
```
mitchellostler, line 110: Setting the store here isn't necessary. Since you are updating the store from the mpxe harness, you don't need to change the values going into the harness from the c side, they only need to go python -> c.  After you get rid of this line, the smoke tests might fail because they are expecting the fixed value. In your integration tests on the python side, set the mcp3427 value to the expected (20) and then the smoke test will read this value. Should be a minor change 🙂

mitchellostler (changes requested):
Hey Anthony, great work getting this done! 
I've done an MPXE driver support ticket before so I know there's a big learning curve and it's no cakewalk lol. 
It all looks good, just some minor changes to be made. Let me know if you have any questions (in the comments or over slack)  🚀 🚀 🚀

mpxe/integration_tests/test_mcp3427.py
```diff
@@ -0,0 +1,21 @@
...
+from mpxe.sims.mcp3427 import Mcp3427
+
+
+class TestMcp3427(int_test.IntTest):
+    def setUp(self):
+        super().setUp()
+        self.mcp3427 = self.manager.start('smoke_mcp3427', Mcp3427())
+
+    def test_mcp3427(self):
+        time.sleep(1)
```
mitchellostler, line 14: I mentioned it above, but you should have a update_mcp3427_reading() function in the sim, which you can call with 20 as the value here and then the assert_store_value line below should work the same

PR #292, opened 2021-02-22
Firmware Homework 103 Part 2

apardawa:


projects/soft_999_akbar_pardawalla_fw_103/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+  adc_register_callback(*channel, adc_callback, address);
+}
+
+void adc_callback(AdcChannel addr, void *context) {
+  uint16_t reading = 0;
+  GpioAddress *address = context;
+
+  adc_read_converted(addr, &reading);
+
+  LOG_DEBUG("Converted Logic: %d", reading);
```
copperium, line 20: You need a `\n` at the end here.

projects/soft_999_akbar_pardawalla_fw_103/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+#include "log.h"
+
+void adc_callback(AdcChannel addr, void *context);
+
+void bttn_callback(const GpioAddress *address, void *context) {
+  AdcChannel *channel = context;
+  adc_register_callback(*channel, adc_callback, address);
+}
+
+void adc_callback(AdcChannel addr, void *context) {
```
copperium, line 14: Since these are private functions, they should be static and prefixed with `prv_`. (Also, `btn` is the standard abbreviation for `button`.)

projects/soft_999_akbar_pardawalla_fw_103/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+  adc_set_channel(channel, true);
+
+  GpioAddress bttn_addr = { .port = GPIO_PORT_B, .pin = 2 };
+
+  InterruptSettings event_settings = {
+    .type = INTERRUPT_TYPE_EVENT,           //
+    .priority = INTERRUPT_PRIORITY_NORMAL,  //
+  };
+
+  gpio_it_register_interrupt(&bttn_addr, &event_settings, INTERRUPT_EDGE_RISING, bttn_callback,
```
copperium, line 56: The homework does say that the interrupt should trigger when the button is released, not when it's initially pressed. The button is active-high, so it's high when pressed and low when not pressed. So I don't think `INTERRUPT_EDGE_RISING` is the correct interrupt edge here.

projects/soft_999_akbar_pardawalla_fw_103/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+                             &channel);
+
+  gpio_it_trigger_interrupt(&bttn_addr);
```
copperium, line 59: `gpio_it_trigger_interrupt` simulates an interrupt in software; this is alright for testing with `make run`.

copperium (changes requested):
Nice job! If you want to test it on x86 (which I assume is what you were doing with the `gpio_it_trigger_interrupt` calls), you can do `make run PLATFORM=x86 PROJECT=soft_999_akbar_pardawalla_fw_103`. (See the readme or the top of the `Makefile` for more `make` commands.)

copperium:
Closing for inactivity, feel free to reopen if you want to fix the comments.

PR #293, opened 2021-02-27
[SOFT-426] Bootloader config

Kyle02:


projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+
+  uint32_t config_1_check_crc = crc32_arr((uint8_t *)&s_config_1_blob, sizeof(s_config_1_blob));
+  uint32_t config_2_check_crc = crc32_arr((uint8_t *)&s_config_2_blob, sizeof(s_config_2_blob));
+
+  bool is_config_1_equal = config_1_crc == config_1_check_crc;
+  bool is_config_2_equal = config_2_crc == config_2_check_crc;
+
+  if (is_config_1_equal && is_config_2_equal) {
+    s_config_1_blob.CRC32 = config_1_crc;
+    s_config_2_blob.CRC32 = config_2_crc;
```
copperium, line 40: If you take my suggestion of factoring the crc32 calculation out into its own function, then this part will be done automatically, so you won't need this case (nor the `s_config_X_blob.crc32 = config_X_crc` lines in other cases). So remove it and restructure the if/else if chain.

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+  s_config_1_blob.CRC32 = 0;
+
+  uint32_t config_2_crc = s_config_2_blob.CRC32;
+  s_config_2_blob.CRC32 = 0;
+
+  uint32_t config_1_check_crc = crc32_arr((uint8_t *)&s_config_1_blob, sizeof(s_config_1_blob));
+  uint32_t config_2_check_crc = crc32_arr((uint8_t *)&s_config_2_blob, sizeof(s_config_2_blob));
+
+  bool is_config_1_equal = config_1_crc == config_1_check_crc;
+  bool is_config_2_equal = config_2_crc == config_2_check_crc;
```
copperium, line 36: `is_config_X_equal` isn't a particularly descriptive name - if I just see the name, what does it mean for one of the configs to be "equal"? Change these names to be something about whether it's ok or corrupted; IMO the most readable form would be to invert them and name them something like `is_config_X_corrupted`.

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+static PersistStorage s_config_1_persist = { 0 };
+static BootloaderConfig s_config_1_blob = { 0 };
+
+static PersistStorage s_config_2_persist = { 0 };
+static BootloaderConfig s_config_2_blob = { 0 };
+
+#define BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_1_START))
+#define BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_2_START))
+
+StatusCode config_init(void) {
```
copperium, line 17: The safety features going on in this function are rather involved, so you should have a big comment explaining what it's doing (and why), and small comments each time it does something non-self-explanatory (like loading from flash or restoring a corrupt page).

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+
+  return STATUS_CODE_OK;
+}
+
+StatusCode config_get(BootloaderConfig *config) {
+  // Please persist any changes
+  memcpy(config, &s_config_1_blob, sizeof(s_config_1_blob));
+
+  return STATUS_CODE_OK;
+}
```
copperium, line 62: Since this always returns `STATUS_CODE_OK`, does having it just return `void` make sense?

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+
+StatusCode config_commit(BootloaderConfig *input_config) {
```
copperium, line 64: Like the init function, there's some complex safety stuff going on here, so make sure to have a big comment explaining it and then comment when the function does something non-trivial.

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,87 @@
...
+  uint32_t input_config_crc = crc32_arr((uint8_t *)input_config, sizeof(input_config));
+  input_config->CRC32 = input_config_crc;
+
+  memcpy(&s_config_1_blob, input_config, sizeof(*input_config));
+  persist_commit(&s_config_1_persist);
+
+  status_ok_or_return(persist_init(&s_config_1_persist, BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE,
+                                   &s_config_1_blob, sizeof(s_config_1_blob), false));
+
+  bool input_equals_config_1 = memcmp(&s_config_1_blob, input_config, sizeof(*input_config)) == 0;
```
copperium, line 75: This is also kind of a cryptic name eh? IMO something like `config_1_written_correctly` would be better, or even invert the polarity and call it `config_1_corrupted`.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+static PersistStorage s_test_persist_storage_2 = { 0 };
+static BootloaderConfig s_test_config_2 = { 0 };
+
+#define BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_1_START))
+#define BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_2_START))
+
+static void (*s_function_pointer_hook)() = NULL;
+
+static void prv_change_config_page(void) {
+  // This function chnages the values in config page 1
```
copperium, line 26: Typo: `chnages`

projects/bootloader/src/main.c
```diff
@@ -1,8 +1,18 @@
...
+#include "flash.h"
 #include "jump_to_application.h"
 #include "log.h"
+#include "soft_timer.h"
 
 int main(void) {
   LOG_DEBUG("Hello from the bootloader!\n");
+  // initialize all the modules
+  flash_init();
+  soft_timer_init();
```
copperium, line 12: Soft timers require interrupts to be initialized eh?

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+#include "crc32.h"
+#include "flash.h"
+#include "log.h"
+#include "persist.h"
+#include "soft_timer.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+extern StatusCode __real_persist_init(PersistStorage *persist, FlashPage page, void *blob,
+                                      size_t blob_size, bool overwrite);
```
copperium, line 12: This is great, but you should definitely comment on it since it's a super obscure feature that most people won't know unless they've dug into how mocks work. Also, I'd like to keep this, `s_function_pointer_hook`, and the `persist_init` mock itself together since they form one sort of logical unit.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+static PersistStorage s_test_persist_storage_1 = { 0 };
+static BootloaderConfig s_test_config_1 = { 0 };
+
+static PersistStorage s_test_persist_storage_2 = { 0 };
+static BootloaderConfig s_test_config_2 = { 0 };
+
+#define BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_1_START))
+#define BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE (FLASH_ADDR_TO_PAGE(BOOTLOADER_CONFIG_PAGE_2_START))
+
+static void (*s_function_pointer_hook)() = NULL;
```
copperium, line 23: The parameter list here (which is the `()` on the end) should be `(void)` instead of `()`.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  persist_init(&s_test_persist_storage_2, BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE, &s_test_config_2,
+               sizeof(s_test_config_2), false);
+  persist_ctrl_periodic(&s_test_persist_storage_2, false);
+
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_2, &s_test_config_1, sizeof(s_test_config_2));
+}
+
+StatusCode TEST_MOCK(persist_init)(PersistStorage *persist, FlashPage page, void *blob,
+                                   size_t blob_size, bool overwrite) {
+  if (s_function_pointer_hook != NULL) (*s_function_pointer_hook)();
```
copperium, line 167: You don't need to dereference the function pointer manually, `s_function_pointer_hook()` will do it automatically.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  __real_persist_init(persist, page, blob, blob_size, overwrite);
+
+  return STATUS_CODE_OK;
```
copperium, line 170: You should just return `__real_persist_init` instead of calling it and then always returning OK eh?

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  // during config_commit
+  s_test_config_1.controller_board_ID = 6;
+  s_test_config_1.app_code = 6;
+  s_test_config_1.app_size = 6;
+  persist_commit(&s_test_persist_storage_1);
+}
+
+void setup_test(void) {
+  flash_init();
+  soft_timer_init();
```
copperium, line 37: Need to initialize interrupts here too

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+               sizeof(s_test_config_1), false);
+  persist_ctrl_periodic(&s_test_persist_storage_1, false);
+
+  persist_init(&s_test_persist_storage_2, BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE, &s_test_config_2,
+               sizeof(s_test_config_2), false);
+  persist_ctrl_periodic(&s_test_persist_storage_2, false);
+
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_1, &s_test_config_2, sizeof(s_test_config_1));
+
+  // Test 2, testing if config_init works for a faulty page (in this case, page 2)
```
copperium, line 76: If these separate tests within the test function are fully separate then you should put them into separate test functions, that way they'll be shown independently in test output. There's no rule that you need one test function per function, you can have more than one.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  // Below, I'm purposefully changing config 2 so that it becomes faulty
```
copperium, line 77: Love the comments here!

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  persist_init(&s_test_persist_storage_1, BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE, &s_test_config_1,
+               sizeof(s_test_config_1), false);
+  persist_ctrl_periodic(&s_test_persist_storage_1, false);
+
+  persist_init(&s_test_persist_storage_2, BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE, &s_test_config_2,
+               sizeof(s_test_config_2), false);
+  persist_ctrl_periodic(&s_test_persist_storage_2, false);
+
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_1, &s_test_config_2, sizeof(s_test_config_1));
+
```
copperium, line 94: You should also test the case where config 1 is corrupted but config 2 is ok

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+void test_config_commit(void) {
+  // I need to create an input config to give config_commit();
+  // I use random values for the input_config
+  BootloaderConfig test_input_config = { .CRC32 = 1,
+                                         .controller_board_ID = 1,
+                                         .controller_board_name = { 'a', '\0' },
+                                         .project_present = true,
+                                         .project_name = { 'a', '\0' },
+                                         .project_info = { 'a', '\0' },
+                                         .git_version = { 'a', '\0' },
```
copperium, line 131: You can actually just use strings here instead of manually inputting char arrays! (i.e. `.controller_board_name = "abcd"` will work). All strings are in C are char arrays, so they're valid as char array literals.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+
+  persist_init(&s_test_persist_storage_2, BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE, &s_test_config_2,
+               sizeof(s_test_config_2), false);
+  persist_ctrl_periodic(&s_test_persist_storage_2, false);
+
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_1, &test_input_config, sizeof(s_test_config_1));
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_2, &s_test_config_1, sizeof(s_test_config_2));
+
+  // Test 2, testing if page 1 is reset when the input config was corrupted
+  s_function_pointer_hook = &prv_change_config_page;
```
copperium, line 150: You don't need the `&` (and we don't use it in the rest of the codebase), just setting it to the function itself will auto convert to a function pointer.

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+  TEST_ASSERT_EQUAL(STATUS_CODE_INTERNAL_ERROR, config_commit(&test_input_config));
+  s_function_pointer_hook = NULL;
```
copperium, line 152: You should probably be doing this in `setup_test`

projects/bootloader/test/test_config.c
```diff
@@ -0,0 +1,171 @@
...
+
+  persist_init(&s_test_persist_storage_1, BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE, &s_test_config_1,
+               sizeof(s_test_config_1), false);
+  persist_ctrl_periodic(&s_test_persist_storage_1, false);
+
+  persist_init(&s_test_persist_storage_2, BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE, &s_test_config_2,
+               sizeof(s_test_config_2), false);
+  persist_ctrl_periodic(&s_test_persist_storage_2, false);
+
+  TEST_ASSERT_EQUAL_MEMORY(&s_test_config_2, &s_test_config_1, sizeof(s_test_config_2));
```
copperium, line 162: Should probably also check that it's NOT equal to `test_input_config`.

copperium (changes requested):
Great job Kyle!

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,103 @@
...
+  if (is_config_1_corrupted && is_config_2_corrupted) {
+    LOG_CRITICAL("ERROR: Both config pages corrupted \n");
+    return STATUS_CODE_INTERNAL_ERROR;
+  } else if (is_config_2_corrupted) {
+    memcpy(&s_config_2_blob, &s_config_1_blob, sizeof(BootloaderConfig));
+    persist_commit(&s_config_2_persist);
+  } else {
+    memcpy(&s_config_1_blob, &s_config_2_blob, sizeof(BootloaderConfig));
+    persist_commit(&s_config_1_persist);
+  }
```
copperium, line 62: This will be run even if neither page is corrupted, which will increase wear on the flash (because we'll be writing to flash every time the board starts up). So instead of a plain `else` clause, I think you should guard this with `else if (config_1_corrupted || config_1_check_crc != config_2_check_crc)`. We'd want to do this if config 1 is corrupted (obviously), and the `config_1_check_crc != config_2_check_crc` clause is for the edge case where the configs are inconsistent but both valid - we'll just default to the second config.

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,103 @@
...
+  memcpy(&s_config_1_blob, input_config, sizeof(BootloaderConfig));
+  persist_commit(&s_config_1_persist);
+
+  // This persist_init updates the s_config_1_blob by pulling from the flash
+  status_ok_or_return(persist_init(&s_config_1_persist, BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE,
+                                   &s_config_1_blob, sizeof(BootloaderConfig), false));
+
+  // checks to see if blob 1 and input config are different to detect corruption
+  bool config_1_corrupted =
+      !(memcmp(&s_config_1_blob, input_config, sizeof(BootloaderConfig)) == 0);
```
copperium, line 91: nitpick: you can just do `(xyz != 0)` instead of `!(xyz == 0)`, might be clearer

projects/bootloader/inc/config.h
```diff
@@ -0,0 +1,38 @@
...
+StatusCode config_init(void);
+
+// Gets the config for page 1, and memcpys it to the input config pointer
+// Please persist any changes you make
+void config_get(BootloaderConfig *config);
+
+// Will add the input config as the new config
+// If the config is corrupted during transfer, then the
+// module will use the old redundant page to "reset" the config
+StatusCode config_commit(BootloaderConfig *input_config);
```
copperium, line 38: I think you should mention that if `config_commit` fails it'll return `STATUS_CODE_INTERNAL_ERROR`, and also that the user doesn't have to fill in the crc32 field.

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,103 @@
...
+  }
+  return STATUS_CODE_OK;
+}
+
+void config_get(BootloaderConfig *config) {
+  // Please persist any changes
+  memcpy(config, &s_config_1_blob, sizeof(BootloaderConfig));
+}
+
+StatusCode config_commit(BootloaderConfig *input_config) {
```
copperium, line 71: nitpick: just to be super safe can you check that the input config isn't NULL at the start of these two functions - it would really suck to have a segfault in the bootloader code

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,103 @@
...
+  config->crc32 = 0;
+  uint32_t calculated_crc = crc32_arr((uint8_t *)config, sizeof(BootloaderConfig));
+  config->crc32 = temp_crc;
+
+  return calculated_crc;
+}
+
+StatusCode config_init(void) {
+  // This function makes sure that both config pages will not end up corrupted
+  // The saftey functions are that if a page is corrupted but the other is not then
```
copperium, line 32: typo: `saftey`

projects/bootloader/src/config.c
```diff
@@ -0,0 +1,103 @@
...
+
+  // This memcpy copies over the input config onto the config page 1
+  // This copy is commited into the storage, to change s_config_1_blob
+  // a persist_init will be called
+  memcpy(&s_config_1_blob, input_config, sizeof(BootloaderConfig));
+  persist_commit(&s_config_1_persist);
+
+  // This persist_init updates the s_config_1_blob by pulling from the flash
+  status_ok_or_return(persist_init(&s_config_1_persist, BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE,
+                                   &s_config_1_blob, sizeof(BootloaderConfig), false));
```
copperium, line 87: You need to do the `persist_ctrl_periodic` thing here right?

copperium (changes requested):
Just a few functional changes that I didn't catch the first time around

copperium (approved):
Looks good, once you fix the formatting I'll merge

PR #298, opened 2021-02-28
Completed 102 HW timer_challenge

markorenic:
My second attempt at a pull request, still could not figure out why the make format and make lint are failing tho.

### DIFF
--- a/projects/timer_challenge/src/main.c
+++ b/projects/timer_challenge/src/main.c
@@ -5,14 +5,14 @@
 #include "soft_timer.h"
 #include "wait.h"
 
-#define TIMER_PERIOD 500  // timer in ms
+#define TIMER_PERIOD_MS 500  // timer in ms
 
 typedef struct Counters {  // stores counters
   uint8_t counter_a;
   uint8_t counter_b;
 } Counters;
 
-void prv_timer_callback(SoftTimerId timer_id, void *context) {
+static void prv_timer_callback(SoftTimerId timer_id, void *context) {
   Counters *counters = context;
 
   counters->counter_a++;  // increments a every 0.5s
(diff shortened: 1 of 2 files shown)


projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,43 @@
...
+#include "wait.h"
+
+#define TIMER_PERIOD 500  // timer in ms
+
+typedef struct Counters {  // stores counters
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 15: This is a private function, so it should be `static` so that it's local to the file; otherwise if some other file tried to define a `prv_timer_callback` there might be an issue.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,43 @@
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define TIMER_PERIOD 500  // timer in ms
```
copperium, line 8: Our usual convention is to suffix variables with the unit they're in, so this should be `TIMER_PERIOD_MS`. Great job using a #define though, most people don't think to!

copperium (changes requested):
Awesome, just a couple minor formatting things to fix up before FW 103.

projects/timer_challenge/src/main.c
markorenic: Fixed tnx

copperium:
LGTM, feel free to move on to FW 103. Closing this since homework is complete.

PR #299, opened 2021-02-28
[SOFT-440] Fix watchdog cancelling random soft timers

copperium:
The `watchdog` library wasn't setting the `timer_id` to `SOFT_TIMER_INVALID_TIMER` before calling `watchdog_kick` which cancels the timer with id `timer_id`, so if we're unlucky, the default `timer_id` would be a valid timer id used by some other soft timer which would get cancelled, causing *super* hard to debug issues. This appeared when trying to debug babydriver.

Fixed by setting the timer ID in `watchdog_start`. Also changed all the empty parameter lists in test_watchdog.c to `(void)` instead of `()` to comply with conventions.


mitchellostler (approved):
Nice find, can't imagine how much debugging that must have taken 😬

PR #300, opened 2021-02-28
[SOFT-441] Don't initialize interrupts and soft timers in x86 adc.c

copperium:
The x86 `adc` library was initializing interrupts and soft timers in `adc_init`, which was causing problems since reinitializing those libraries does nasty stuff like clearing all currently registered interrupts, cancelling soft timers, etc. Fixed by not initializing them.


mitchellostler (approved):
LGTM, looks like lint is failing though?

copperium:
that's what I get for skipping the git hooks

PR #301, opened 2021-03-01
[SOFT-437] integrate voltage regulator into pd

mitchellostler:
Made the change in power_distribution/main.c, didn't think it would be worth it to make another module.
I tested locally, and the function is getting called if the monitor GPIO goes low, and transmitting can message as expected. Figure this should be enough for now?

I updated the pd fan control fault message to unified FRONT_PD_FAULT and REAR_PD_FAULT messages, as we were using less than half of the 16 bits allotted for error flags. will open a pr to the codegen repo as well. I updated the tests to reflect this


copperium (approved):
LGTM, you can fix the nitpick if you want

copperium (approved):
LGTM, looks like something weird went on with git though, I'll try merging and hopefully it goes okay

PR #302, opened 2021-03-01
FW_103 part 2

markorenic:
I did guess some of the settings i was unsure about. Also not sure how to test? I know I can run it with x86 but the output isn't very useful.


copperium:
Looks good! Closing since the homework is complete, feel free to do FW 104 and/or see me for a task.

(It is hard to test yeah, that's a flaw with the homework description.)

PR #304, opened 2021-03-02
Firmware Onboarding 102

simcard0000:
Here's the homework submission!

### DIFF
--- a/projects/mod_2_homework/src/main.c
+++ b/projects/mod_2_homework/src/main.c
@@ -1,0 +1,37 @@
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void timer_callback(SoftTimerId timer_id, void *ctx) {
+  Counters *ctr = ctx;
+  ctr->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", ctr->counter_a);
+  if (ctr->counter_a % 2 == 0) {
+    ctr->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", ctr->counter_b);
+  }
+  soft_timer_start_seconds(0.5, timer_callback, ctr, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters ctr = { 0 };
+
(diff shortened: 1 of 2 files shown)


projects/mod_2_homework/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+
+static void timer_callback(SoftTimerId timer_id, void *ctx) {
+  Counters *ctr = ctx;
+  ctr->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", ctr->counter_a);
+  if (ctr->counter_a % 2 == 0) {
+    ctr->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", ctr->counter_b);
+  }
+  soft_timer_start_seconds(0.5, timer_callback, ctr, NULL);
```
copperium, line 21: We usually ask that you factor out the timeout from here and the `soft_timer_start_seconds` call later into a #define (named descriptively and with the unit as a suffix of course). Also, it's more usual to use `soft_timer_start_millis` for timeouts that aren't a whole number of seconds due to the potential for floating point nonsense.

projects/mod_2_homework/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void timer_callback(SoftTimerId timer_id, void *ctx) {
```
copperium, line 13: This is a private function so it should be prefixed with `prv_`.

projects/mod_2_homework/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct {
```
copperium, line 8: The usual convention in our codebase is to put the name of the struct both here (i.e. `typedef struct Counters {`) and in the typedef, it helps with IDE name resolution.

copperium (changes requested):
Nice job Sims, love the gif! Just a couple style points.

projects/mod_2_homework/src/main.c
simcard0000: Ah that makes sense; I'll change it 👍🏾

simcard0000:
Let me know if I covered all the changes!

copperium:
LGTM! Closing this since homework is complete, feel free to move on to FW 103.

PR #305, opened 2021-03-03
Fixing MPXE pylint errors

selinahsu:
Formatting / whitespace / typo fixes throughout mpxe. 

The following pylint messages are now being disabled in Makefile: 
- missing-module-docstring
- missing-class-docstring
- missing-function-docstring
- import-error 
- invalid-name
- redefined-outer-name
- unused-argument
- too-few-public-methods
- duplicate-code
- no-self-use


mpxe/harness/pm.py
```diff
@@ -52,12 +51,12 @@ def stop_all(self):
...
         def prep_poll():
             poll = select.poll()
-            for fd in self.fd_to_proj.keys():
+            for fd in self.fd_to_proj:
                 poll.register(fd, select.POLLIN)
             return poll
 
         def handle_poll_res(fd, event):
-            if not (event & select.POLLIN):
+            if not event & select.POLLIN:
```
copperium, line 59: IMO this change leaves the line ambiguous; I'd have to look up the Python precedence chart to make sure it's `not (event & select.POLLIN)` and not `(not event) & select.POLLIN`. Perhaps change it to `if (event & select.POLLIN) == 0:` - that version is unambiguous and accepted by pylint.

mpxe/harness/pm.py
```diff
@@ -81,7 +80,7 @@ def handle_poll_res(fd, event):
                 for fd, event in res_list:
                     handle_poll_res(fd, event)
         except InvalidPollError as e:
-            return
+            return e
```
copperium, line 83: Returning `e` here is rather confusing since `poll` runs in a thread and its return value isn't used. IMO a better way to fix the pylint warning would be to remove the `as e` part entirely: `except InvalidPollError:` will catch InvalidPollErrors but won't assign a name.

mpxe/integration_tests/test_pca9539r.py
```diff
@@ -15,7 +15,7 @@ def test_pca9539r(self):
         time.sleep(0.1)
         self.pca9539r.sim.assert_store_values(self, False)
         value = True
-        for x in range(3):
+        for _x in range(3):
```
copperium, line 18: I think the usual convention for iterating n times without using the iteration variable is to use `_` like `for _ in range(3):`

mpxe/sims/adt7476a.py
```diff
@@ -1,9 +1,12 @@
 from mpxe.protogen import stores_pb2
-from mpxe.protogen import adt7476a_pb2
 
 from mpxe.sims import sim
 
 class Adt7476a(sim.Sim):
+    def __init__(self):
+        self.speed = None
+        self.status = None
```
copperium, line 8: IMO, since these properties are set to/used as arrays of 3 ints elsewhere, they should be initialized as `[0, 0, 0]` rather than None.

mpxe/sims/mcp23008.py
```diff
@@ -23,11 +25,11 @@ def update_pin_state(self, proj, pin, state):
...
         proj.write_store(mcp23008_msg, mcp23008_mask, stores_pb2.MxStoreType.MCP23008)
 
     # Compares pin state against store
+    # pylint: disable=unused-argument
     def assert_store_value_reading(self, proj, pin, state):
-        for i in range(NUM_MCP_PINS):
-            assert(self.states[pin] == state)
-
+        for _i in range(NUM_MCP_PINS):
+            assert self.states[pin] == state
```
copperium, line 35: I'm not quite sure what this function was intended to do (@mitchellostler ?) but to me it looks safe to remove the `for` line entirely and just leave the assert.

Makefile
```diff
@@ -202,8 +202,12 @@ lint_quick:
 	@echo "Quick linting on ONLY changed/new files"
 	@$(FIND_MOD_NEW) | xargs -r python2 lint.py
 
-# Disable import error
-PYLINT_DISABLE := F0401 duplicate-code
+# Disable all the following pylint messages:
+PYLINT_DISABLE := \
+	missing-module-docstring missing-class-docstring missing-function-docstring \
+	import-error invalid-name redefined-outer-name unused-argument \
+	too-few-public-methods duplicate-code no-self-use
```
copperium, line 209: I'd appreciate if some of these could be disabled only for MPXE (the `missing-X-docstring` messages I've found have been really good for improving documentation in babydriver, and `invalid-name` probably shouldn't be disabled outside MPXE). The rest look good to disable everywhere though.

Also, thanks for expanding `F0401` into the more readable `import-error` :)

copperium (changes requested):
Left a few comments, mostly LGTM. The unity changes definitely shouldn't be there though - since we upgraded our unity version recently, I think `make format_quick` running in the git hooks is detecting the difference between the version on master and the (old) version on mpxe and trying to format it (yet another reason we need to merge the mpxe branch), which will cause a ton of merge conflicts when we merge the `mpxe` branch - please reset those files so they aren't changed and just skip the hooks with `--no-verify`.

mpxe/sims/mcp23008.py
mitchellostler: Yeah should be good to get rid of, function should be just checking state of pin idk why i put a for loop there lol

libraries/unity/auto/parse_output.rb
```diff
@@ -0,0 +1,322 @@
+#============================================================
```
copperium, line 1: Please remove this file, it got missed when resetting unity to mpxe!

libraries/unity/auto/run_test.erb
```diff
@@ -0,0 +1,37 @@
+/*=======Test Runner Used To Run Each Test=====*/
```
copperium, line 1: Also remove this one!

copperium (changes requested):

copperium:
@selinahsu I think you reset unity back to master...the goal is for the PR not to show any changes in the unity library.

copperium:
@selinahsu The unity dir looks good but I'm not sure why you reverted the "small pylint and readability changes" commit, now a bunch of the original comments are unresolved...

copperium (approved):

PR #306, opened 2021-03-06
[SOFT-414] Update BPS fault sequencing

JarvisWeng:


copperium (changes requested):
Here are my thoughts, would like your input @g-jessmuir as well.

My reading of the [VDR](https://docs.google.com/document/d/19CH-fSSRmeMvJ70s5vPP6m2Das0xHQHGyqE-kpt_QsY/edit) / the [meeting notes](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2041315679/2021-01-16+BPS+strobe+sequencing) suggests we should add another power state, "BPS fault", where most things are turned off (possibly by PD responding to the BPS fault message rather than by centre console sending a "turn off things for BPS fault" message). That state transitions only to off, which it does when the power button is pressed, at which point the whole power off sequence goes (turning off telemetry pi, discharging precharge, etc) and the strobe light is turned off.

The other option is to use our current strategy of immediately turning the car off but turning on the strobe light. We then set a "BPS fault" flag that stops us from transitioning out of off, so the only way to get out of the off state (or even clear the strobe) is to disconnect the aux battery and power cycle the whole system (or at least centre console and PD).

I'm actually not sure which option is best / intended - @g-jessmuir do you have additional context?

g-jessmuir:
I talked briefly with Aashmika and there's really not much to do in firmware to gracefully transition to off. So, I'm in favor of telling PD to strobe, then throwing a critical section infinite loop in center console to lock up firmware until we're hard reset by toggling the aux switch. This should keep it reasonably simple. Don't try to write a unit test for this.

Thoughts @copperium ?

JarvisWeng:
> Here are my thoughts, would like your input @g-jessmuir as well.
> 
> My reading of the [VDR](https://docs.google.com/document/d/19CH-fSSRmeMvJ70s5vPP6m2Das0xHQHGyqE-kpt_QsY/edit) / the [meeting notes](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2041315679/2021-01-16+BPS+strobe+sequencing) suggests we should add another power state, "BPS fault", where most things are turned off (possibly by PD responding to the BPS fault message rather than by centre console sending a "turn off things for BPS fault" message). That state transitions only to off, which it does when the power button is pressed, at which point the whole power off sequence goes (turning off telemetry pi, discharging precharge, etc) and the strobe light is turned off.
> 
> The other option is to use our current strategy of immediately turning the car off but turning on the strobe light. We then set a "BPS fault" flag that stops us from transitioning out of off, so the only way to get out of the off state (or even clear the strobe) is to disconnect the aux battery and power cycle the whole system (or at least centre console and PD).
> 
> I'm actually not sure which option is best / intended - @g-jessmuir do you have additional context?

I had thought that the BPS fault could only be resolved if the car is turn off. So if we add another state, we have to power off the car ourselves? And power distribution would need to send a SET_RELAY message?

projects/centre_console/src/power_fsm.c
```diff
@@ -58,10 +58,15 @@ static void prv_state_fault_output(Fsm *fsm, const Event *e, void *context) {
   power_fsm->destination_state = power_fsm->previous_state;
   FaultReason fault = { .raw = e->data };
   power_fsm->current_state = POWER_STATE_FAULT;
+
   if (fault.fields.area != EE_CONSOLE_FAULT_AREA_BPS_HEARTBEAT) {
+    // set the strobe light off
+    CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_STROBE, EE_LIGHT_STATE_OFF);
```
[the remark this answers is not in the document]
JarvisWeng: Ah you're right. I think I had put this to test something, but yes I should remove this

copperium:
@g-jessmuir That would probably work although it seems super janky - shouldn't we at least tell MCI to discharge precharge though? (Not really sure how that works actually.) And maybe tell PD to turn off a bunch of stuff? It could intercept the bps fault message though.

@JarvisWeng yeah we would have to turn it off manually. Don't think we'd need PD to send a SET_RELAY message though. The "off" state doesn't actually turn off all power, it just switches to the aux battery and turns off most systems.

g-jessmuir:
@copperium You're right, +1 for intercepting the message from PD and just setting the config to whatever's powered on aux. This should automatically discharge mci. 

We love jank around here, don't we?

JarvisWeng:
> @g-jessmuir That would probably work although it seems super janky - shouldn't we at least tell MCI to discharge precharge though? (Not really sure how that works actually.) And maybe tell PD to turn off a bunch of stuff? It could intercept the bps fault message though.
> 
> @JarvisWeng yeah we would have to turn it off manually. Don't think we'd need PD to send a SET_RELAY message though. The "off" state doesn't actually turn off all power, it just switches to the aux battery and turns off most systems.

@copperium So I should receive the message in Power Distribution by adding another can receiver. And when I receive the message, I should transition the PD state to POWER_DISTRIBUTION_POWER_SEQUENCE_EVENT_TURN_OFF_EVERYTHING and turn on the strobe light?
Wouldn't there be some timing issues then? Since centre console would still send the can message to power off. Or would I get rid of that?

jessm:
@copperium Center console doesn't control heartbeat failures, bms does. In the case of a false positive heartbeat failure, bms will open relays and then tell center console about it, so there's no point in center console telling bms to again. 

I don't think it's accurate that center console can unilaterally declare a bps fault. 

All things considered, the power sequencing is already pretty screwed if you consider failure modes that I think requiring a hard reset keeps things in a more consistent state than trying to gracefully fail.

copperium:
@jessm It looks to me that fault_monitor has a 4000ms watchdog timeout which if it expires declares a BPS fault - my concern is that if that happens and bms doesn't know, how big of a deal is it to manually open the relays to do a hard reset from main power?

I suppose this is probably a moot point because the bps heartbeat has the highest priority on the bus, so pretty much the only way the watchdog would expire would be if bms was truly disconnected from CAN or was otherwise screwed, in which case the open-relays message wouldn't do much anyways. So we can probably just skip sending the message.

And hm, the infinite loop is probably fine then. Really the only critical thing we would have wanted to keep centre console running for would be the ebrake but that's gone now so it doesn't matter.

projects/centre_console/src/power_fsm.c
```diff
@@ -64,6 +65,8 @@ static void prv_state_fault_output(Fsm *fsm, const Event *e, void *context) {
   } else {
     power_fsm->previous_state = POWER_STATE_OFF;
     event_raise(CENTRE_CONSOLE_POWER_EVENT_OFF, 0);
```
copperium, line 67: We don't need to do this anymore eh? Since we're locking up centre console and letting power distro turn off whatever needs turned off.

projects/centre_console/src/power_fsm.c
```diff
@@ -64,6 +65,8 @@ static void prv_state_fault_output(Fsm *fsm, const Event *e, void *context) {
...
+    CAN_TRANSMIT_DISCHARGE_PRECHARGE();
+    critical_section_start();
```
copperium, line 69: You also need an infinite loop (like a literal `while (1) {}`). Also, this is such a bonkers, out-of-place move that I'd give the BPS fault handling its own function, and it should be heavily commented - say why we're discharging precharge and why we're locking up centre console.

projects/power_distribution/inc/bps_watcher.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+#include "can.h"
```
copperium, line 2: nitpick: blank line here please

projects/power_distribution/inc/bps_watcher.h
```diff
@@ -0,0 +1,7 @@
...
+#include "can_msg_defs.h"
+#include "can_transmit.h"
+#include "can_unpack.h"
+
+StatusCode bps_watcher_init();
```
copperium, line 7: Nitpick: the parameter list should be `(void)` not `()` (also in the .c file). Also please comment on what the module does!

projects/power_distribution/src/bps_watcher.c
```diff
@@ -0,0 +1,16 @@
+#include "bps_watcher.h"
+#include "event_queue.h"
+#include "pd_events.h"
+
+StatusCode prv_bps_watcher_callback_handler(const CanMessage *msg, void *context,
+                                            CanAckStatus *ack_reply) {
+  uint8_t data = 0;
+  CAN_UNPACK_BPS_HEARTBEAT(msg, &data);
+  if (data) event_raise(POWER_DISTRIBUTION_STROBE_EVENT, 1);
```
copperium, line 9: We also need to turn off to whatever's powered on aux, right? So raise a `POWER_DISTRIBUTION_POWER_SEQUENCE_EVENT_TURN_ON_EVERYTHING_AUX` as well (with a comment, bc it sounds like you're turning things on).

Also, I don't think a raw `if (data)` is recommended, better would be `if (data != 0)`, or even better to be self-documenting, `#define NO_FAULT 0` and then `if (data != NO_FAULT)`.

projects/power_distribution/src/bps_watcher.c
```diff
@@ -0,0 +1,16 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode bps_watcher_init() {
+  can_register_rx_handler(SYSTEM_CAN_DEVICE_BMS_CARRIER, prv_bps_watcher_callback_handler, NULL);
+  return STATUS_CODE_OK;
```
copperium, line 15: nitpick: you could just return the `can_register_rx_handler` call eh?

projects/power_distribution/src/bps_watcher.c
```diff
@@ -0,0 +1,16 @@
...
+StatusCode prv_bps_watcher_callback_handler(const CanMessage *msg, void *context,
+                                            CanAckStatus *ack_reply) {
+  uint8_t data = 0;
+  CAN_UNPACK_BPS_HEARTBEAT(msg, &data);
+  if (data) event_raise(POWER_DISTRIBUTION_STROBE_EVENT, 1);
+  return STATUS_CODE_OK;
+}
+
+StatusCode bps_watcher_init() {
+  can_register_rx_handler(SYSTEM_CAN_DEVICE_BMS_CARRIER, prv_bps_watcher_callback_handler, NULL);
```
copperium, line 14: `SYSTEM_CAN_DEVICE_BMS_CARRIER` isn't a CAN message ID, you need to register on `SYSTEM_CAN_MESSAGE_BPS_HEARTBEAT`.

copperium (changes requested):
Nice job, a few comments. Please also write (very simple) tests for `bps_watcher` just to verify that it works and catch anything like the `SYSTEM_CAN_DEVICE_BMS_CARRIER` bug :)

projects/power_distribution/src/bps_watcher.c
JarvisWeng: My bad, looked at the wrong enum

projects/centre_console/src/power_fsm.c
JarvisWeng: I remember why I kept them now. It was to passed a test. Should I just get rid of part of the test?

projects/centre_console/src/power_fsm.c
JarvisWeng: Oh if I do that, a test will never complete. Should I just rid of that part then?

copperium: Yeah don't try testing it, it'll be more of a pain than it's worth

projects/centre_console/src/power_fsm.c
copperium: Yeah just get rid of that part

projects/centre_console/test/test_power_fsm.c
```diff
@@ -146,13 +146,15 @@ void test_bps_faults_to_off(void) {
...
+
+  // FaultReason fault = { .fields = { .area = EE_CONSOLE_FAULT_AREA_BPS_HEARTBEAT, .reason = 0 } };
+  // e.id = CENTRE_CONSOLE_POWER_EVENT_FAULT;
+  // e.data = fault.raw;
+  // TEST_ASSERT_TRUE(power_fsm_process_event(&s_power_fsm_storage, &e));
+  // MS_TEST_HELPER_ASSERT_NEXT_EVENT_ID(e, CENTRE_CONSOLE_POWER_EVENT_OFF);
+  // prv_assert_current_state(POWER_STATE_FAULT);
+  // TEST_ASSERT_TRUE(power_fsm_process_event(&s_power_fsm_storage, &e));
+  // prv_assert_current_state(POWER_STATE_TRANSITIONING);
+  // TEST_ASSERT_TRUE(s_power_fsm_storage.destination_state == POWER_STATE_OFF);
```
copperium, line 159: We shouldn't just leave commented-out code lying around the repo - delete this, and if you want you can add a comment saying that why we aren't testing the BPS fault case.

projects/centre_console/src/power_fsm.c
```diff
@@ -52,6 +53,17 @@ FSM_STATE_TRANSITION(power_state_fault) {
   FSM_ADD_TRANSITION(CENTRE_CONSOLE_POWER_EVENT_OFF, power_state_transitioning);
 }
 
+// This function is called when we catch a bps fault
+// we discharge main and switch to aux
+// we the lockup firmware with an infinite loop
+// until the fault is resolved and the car's been power cycled
+static void prv_bps_fault() {
```
copperium, line 60: Nitpick: please use `(void)` for the parameter list instead of `()`

projects/power_distribution/test/test_bps_watcher.c
```diff
@@ -0,0 +1,93 @@
+#include "bps_watcher.h"
+#include "can.h"
```
copperium, line 2: nitpick: please put a blank line here

projects/power_distribution/src/bps_watcher.c
```diff
@@ -0,0 +1,24 @@
+#include "bps_watcher.h"
+#include "event_queue.h"
```
copperium, line 2: nitpick: blank line here

projects/power_distribution/inc/bps_watcher.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "can.h"
+#include "can_msg_defs.h"
+#include "can_transmit.h"
+#include "can_unpack.h"
+
+// This module simply watches for the BPS_HEARTBEAT
+// to make sure that no fault has occured
```
copperium, line 9: Please add the `// Requires X, Y, and Z to be initialized` comment for easier debugging if something's not initialized

copperium (changes requested):
Looks good to me, just a few small nitpicks (mostly just the commented-out block)

copperium (approved):

PR #307, opened 2021-03-06
Soft 351 race normal switch implementation

CheranMahalingam:


projects/centre_console/rules.mk
```diff
@@ -9,7 +9,9 @@
...
 
 $(T)_test_main_event_generator_MOCKS := get_pedal_state \
-																				drive_fsm_get_global_state \
-																				power_fsm_get_current_state \
-																				get_global_speed_state \
-																				get_global_charging_state
+										drive_fsm_get_global_state \
+										power_fsm_get_current_state \
+										get_global_speed_state \
+										get_global_charging_state
```
copperium, line 15: The indentation messiness seems to be because Arshan used tabs, could you replace them with spaces? (Unless doing so breaks Make, which is entirely possible, in which case never mind.)

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,125 @@
...
+static GpioSettings s_gpio_settings = { .state = GPIO_STATE_LOW,
+                                        .direction = GPIO_DIR_IN,
+                                        .resistor = GPIO_RES_NONE,
+                                        .alt_function = GPIO_ALTFN_NONE };
+
+static InterruptSettings s_interrupt_settings = { .type = INTERRUPT_TYPE_INTERRUPT,
+                                                  .priority = INTERRUPT_PRIORITY_NORMAL };
+
+static void prv_voltage_monitor_error_callback(VoltageRegulatorError error, void *context) {
+  LOG_WARN("Voltage Regulator Error %d\n", error);
```
copperium, line 34: Maybe to save debugging time in the future you could print "off when should be on" or "on when should be off" as appropriate according to the value of `error` instead of the raw number

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,125 @@
...
+  // switch into race or normal mode
+  status_ok_or_return(gpio_it_register_interrupt(&s_race_switch_address, &s_interrupt_settings,
+                                                 INTERRUPT_EDGE_RISING_FALLING,
+                                                 prv_gpio_interrupt_handler, storage));
+
+  status_ok_or_return(
+      voltage_regulator_init(&(storage->voltage_storage), &s_voltage_regulator_settings));
+
+  // Since car begins in normal mode, the 5V regulator must be enabled
+  status_ok_or_return(voltage_regulator_set_enabled(&(storage->voltage_storage), true));
```
copperium, line 114: Hm, I think it's actually possible that the switch could be in any state when we start up the car, and it would be kind of annoying to have to toggle it to turn infotainment off if the switch was in the 'race' position at the start. So instead perhaps let's get the initial state from the value of the gpio pin.

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,125 @@
...
+
+  // Initially the car starts in normal mode
+  storage->current_state = RACE_STATE_OFF;
+  fsm_init(&(storage->race_switch_fsm), "Race Switch FSM", &race_switch_off, storage);
```
copperium, line 118: Nitpick: the arrow operator `->` has a higher precedence than `&`, so you don't need the parentheses in `&(storage->race_switch_fsm)` (or elsewhere in the file)

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,125 @@
...
+static void prv_voltage_monitor_error_callback(VoltageRegulatorError error, void *context) {
+  LOG_WARN("Voltage Regulator Error %d\n", error);
+}
+
+static VoltageRegulatorSettings s_voltage_regulator_settings = {
+  .enable_pin = VOLTAGE_REGULATOR_ENABLE_ADDRESS,
+  .monitor_pin = VOLTAGE_REGULATOR_MONITOR_ADDRESS,
+  .timer_callback_delay_ms = VOLTAGE_REGULATOR_DELAY_MS,
+  .error_callback = prv_voltage_monitor_error_callback,
+  .error_callback_context = NULL
```
copperium, line 42: Nitpick: would prefer a comma at the end of this line too

copperium (changes requested):
This is amazing Cheran, thanks for getting this done so quickly! There's just one or two functionality related changes, the rest are nitpicks.

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,146 @@
...
+  Event e = { 0 };
+
+  // Rising edge indicates the gpio state has become high and the car should go into race mode
+  if (input_state == GPIO_STATE_HIGH) {
+    e.id = s_event_lookup[RACE_SWITCH_EVENT_ON];
+  } else {
+    e.id = s_event_lookup[RACE_SWITCH_EVENT_OFF];
+  }
+
+  race_switch_fsm_process_event(storage, &e);
```
copperium, line 102: Sorry, didn't catch this in the earlier review: What we usually do in module code is raise the event with `event_raise` (or `event_raise_no_data`); then the fsm will transition when the event is processed with `race_switch_fsm_process_event` in the main event loop. This way the transitioning etc happens out of the interrupt, and it gives other events time to process.

To fix your tests to comply with this, you can declare an `Event e`, use `event_process(&e)` after triggering the interrupt to get the event, and then call `race_switch_fsm_process_event` (even tie it up in a helper function if you want).

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,146 @@
...
+
+FSM_DECLARE_STATE(race_switch_off);
+FSM_DECLARE_STATE(race_switch_on);
+
+FSM_STATE_TRANSITION(race_switch_off) {
+  FSM_ADD_TRANSITION(RACE_STATE_ON, race_switch_on);
+}
+
+FSM_STATE_TRANSITION(race_switch_on) {
+  FSM_ADD_TRANSITION(RACE_STATE_OFF, race_switch_off);
```
copperium, line 62: You should be transitioning on the events (i.e. `RACE_SWITCH_EVENT_ON` and `RACE_SWITCH_EVENT_OFF`), rather than the race states. Currently, this will transition if anyone raises an event with id 0 or 1 (currently the CAN RX and TX events!), which isn't what you want. Instead transition on the event IDs themselves and raise events with IDs `RACE_SWITCH_EVENT_ON` etc. You don't need the `s_event_lookup` lookup array at all I don't think.

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,146 @@
...
+
+  status_ok_or_return(
+      voltage_regulator_init(&storage->voltage_storage, &s_voltage_regulator_settings));
+
+  // Determine whether the car starts in normal or race mode
+  GpioState race_switch_initial_state;
+  status_ok_or_return(gpio_get_state(&s_race_switch_address, &race_switch_initial_state));
+
+  // If in race mode (gpio state high or 1) voltage regulator must be disabled, otherwise enable
+  bool is_voltage_regulator_enabled = race_switch_initial_state == GPIO_STATE_HIGH ? false : true;
```
copperium, line 128: Instead of checking `race_switch_initial_state == GPIO_STATE_HIGH` everywhere, I'd prefer to just have a boolean `is_race_mode` (or even a `RaceState` variable here) and check it. (Also, in general, `? false : true` can just be replaced with a boolean NOT `!`.)

As well, I'd prefer that we add a #define for the gpio state that means "race mode" (`GPIO_STATE_HIGH`), just so that in the case where we want to invert it and make `GPIO_STATE_LOW` mean "race mode", we only have to change it in one place.

projects/centre_console/src/race_switch.c
```diff
@@ -0,0 +1,146 @@
...
+  status_ok_or_return(
+      voltage_regulator_set_enabled(&storage->voltage_storage, is_voltage_regulator_enabled));
+
+  storage->current_state =
+      race_switch_initial_state == GPIO_STATE_HIGH ? RACE_STATE_ON : RACE_STATE_OFF;
+
+  if (race_switch_initial_state == GPIO_STATE_HIGH) {
+    fsm_init(&storage->race_switch_fsm, "Race Switch FSM", &race_switch_on, storage);
+  } else {
+    fsm_init(&storage->race_switch_fsm, "Race Switch FSM", &race_switch_off, storage);
```
copperium, line 138: Nit: maybe add a #define for the fsm name just so you're not repeating it?

copperium (changes requested):
Just a couple things I didn't catch earlier

copperium (approved):

PR #309, opened 2021-03-10
FW102 commit to join midsun firmware

KarimAlatrash:


projects/soft_999_karim_alatrash_fw_102/src/main.c
```diff
@@ -0,0 +1,48 @@
+//
+// Created by Karim Alatrash on 2021-03-08.
+//
+#include <stdbool.h>     // for booleans
+#include <stdint.h>      // for integer types
+#include "interrupt.h"   // interrupts are required for soft timers
```
copperium, line 6: Technically according to the [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) you should have a newline between the <standard> imports and the "midsun-specific" imports. Also I don't think you use `<stdbool.h>` so you can remove it, and you can remove the eclipse autogenerated header comments.

projects/soft_999_karim_alatrash_fw_102/src/main.c
```diff
@@ -0,0 +1,48 @@
...
+
+  if (count->counter_a == count->counter_b) {
+    count->counter_a++;
+    LOG_DEBUG("Counter A: %i\n", count->counter_a);
+  } else {
+    count->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", count->counter_b);
+  }
+
+  soft_timer_start_millis(500, increment_counter, count, NULL);
```
copperium, line 27: It's generally a good idea to factor out the 500ms here and below into a `#define` constant. Make sure it has a good descriptive name that ends with the unit as a suffix (in this case `_MS`).

projects/hellow_world/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
```
copperium, line 1: The `hellow_world` project isn't part of the FW102 homework, please remove.

copperium (changes requested):
I love the number of comments! It might even be a bit much for production code, but for homework it shows that you understand each piece. Just a few things to fix before moving onto FW 103.

copperium:
Closing for inactivity, feel free to reopen if you decide to address the comments.

PR #311, opened 2021-03-12
[SOFT-445] Centralized pip dependencies

raiyansayeed:


Makefile
```diff
@@ -195,7 +195,7 @@ pylint:
 format_quick:
 	@echo "Quick format on ONlY changed/new files"
 	@$(FIND_MOD_NEW) | xargs -r clang-format -i -style=file
-	@$(FIND_MOD_NEW_PY) | xargs autopep8 $(AUTOPEP8_CONFIG) -i
+	@$(FIND_MOD_NEW_PY) | xargs -r autopep8 $(AUTOPEP8_CONFIG) -i
```
raiyansayeed, line 198: if no python files were changed, this make command would fail (and you wouldn't be able to commit). The -r flag is added so the following commands run if there isn't any empty input from the pipe

copperium: Thanks for fixing this!

Makefile
```diff
@@ -161,14 +161,14 @@ IGNORE_CLEANUP_LIBS := CMSIS FreeRTOS STM32F0xx_StdPeriph_Driver unity FatFs
 # This uses regex
 IGNORE_PY_FILES := ./lint.py ./libraries/unity.*
 # Find all python files excluding library files in project env (./venv)
-FIND_PY_FILES:= $(shell printf "! -regex %s " $(IGNORE_PY_FILES) | xargs find -path ./venv -prune -o -name '*.py')
+FIND_PY_FILES:= $(shell printf "! -regex %s " $(IGNORE_PY_FILES) | xargs find . \( -path ./.venv -o -path ./libraries -o -path ./lint.py \) -prune -o -name '*.py' -print)
```
copperium, line 164: The point of `IGNORE_PY_FILES` is to filter out ./libraries/unity/* and ./lint.py, is there a reason you added those to the -prune too? At least you should use `IGNORE_PY_FILES` somehow (perhaps with a gnu make function) instead of duplicating.

Makefile
```diff
@@ -161,14 +161,14 @@ IGNORE_CLEANUP_LIBS := CMSIS FreeRTOS STM32F0xx_StdPeriph_Driver unity FatFs
...
 AUTOPEP8_CONFIG:= -a --max-line-length 100 -r
 FIND_PATHS := $(addprefix -o -path $(LIB_DIR)/,$(IGNORE_CLEANUP_LIBS))
 FIND := find $(PROJECT_DIR) $(LIBRARY_DIR) \
 			  \( $(wordlist 2,$(words $(FIND_PATHS)),$(FIND_PATHS)) \) -prune -o \
 				-iname "*.[ch]" -print
-FIND_MOD_NEW := git diff origin/master --name-only --diff-filter=ACMRT -- '*.c' '*.h'
-FIND_MOD_NEW_PY := git diff origin/master --name-only --diff-filter=ACMRT -- '*.py'
+FIND_MOD_NEW := git diff origin/master --name-only --diff-filter=ACMRT -- '*.c' '*.h' 
+FIND_MOD_NEW_PY := git diff origin/master --name-only --diff-filter=ACMRT -- '*.py' ':(exclude)./.venv/*'
```
copperium, line 171: Could you exclude all ignored files (`IGNORE_PY_FILES`/`IGNORE_CLEANUP_LIBS`) here too? That should solve the issue where if we modify an ignored file on master then it gets formatted on every branch made before the change.

Makefile
```diff
@@ -269,14 +270,13 @@ pytest_all:
 	@for i in $$(find projects -name "test_*.py"); 													\
 	do																								\
 		python -m unittest discover -t $$(dirname $$i) -s $$(dirname $$i) -p $$(basename $$i);		\
-	done			
+	done
```
copperium, line 273: maybe could you remove the rest of the trailing whitespace on this line? :)

projects/baby_driver/README.md
```diff
@@ -0,0 +1,9 @@
+# Babydriver 
+
+## Usage
+
+```bash
+# Run python tests
+python test_insert_module_name.py
```
copperium, line 7: You can actually just do `make pytest PROJECT=baby_driver [TEST=module_name]` now. Also if you want to flesh this file out more in a future PR that would be greatly appreciated!

Makefile
```diff
@@ -269,14 +270,13 @@ pytest_all:
...
+	done		
 
 .PHONY: install_requirements
 install_requirements:
-	@for i in $$(find projects -name "requirements.txt"); 		\
-	do															\
-		pip install -r $$i;										\
-	done								
+	@rm -rf .venv
+	@virtualenv .venv
```
copperium, line 278: When I run `make install_requirements` on my system I get `virtualenv: Command not found` - maybe use `python3 -m venv` as it seems more portable.

Makefile
```diff
@@ -243,6 +243,7 @@ new:
 .PHONY: clean
 clean:
 	@rm -rf $(BUILD_DIR)
+	@rm -rf .venv
```
copperium, line 246: `make clean` is just supposed to clean up the build files, I'm not sure about adding this here

Makefile
```diff
@@ -269,14 +270,13 @@ pytest_all:
...
 
 .PHONY: install_requirements
 install_requirements:
-	@for i in $$(find projects -name "requirements.txt"); 		\
-	do															\
-		pip install -r $$i;										\
-	done								
+	@rm -rf .venv
+	@virtualenv .venv
+	@pip install -r requirements.txt
```
copperium, line 279: Are you sure this is using the venv pip? (i.e. should it be `.venv/bin/pip`?) Or maybe should you automatically do `source .venv/bin/activate`? (though that might not work if every line is a separate shell)

README.md
```diff
@@ -52,15 +52,18 @@ make pylint
 # Define a symbol from the command line - for example, setting the log level
 make test LIBRARY=ms-common DEFINE="LOG_LEVEL_VERBOSITY=LOG_LEVEL_WARN"
 
+# Intall python requirements for every project
+make install_requirements
+
+# Note that for python make commands, you will likely need to be in virtualenv. After running install_requirements, run
+source .venv/bin/activate
```
copperium, line 59: Does this need to be run every time a new shell is opened, or does it work just running it once? If the first, we should add it to `~/.bashrc`, and if the second, we should just do it automatically in `make install_requirements`.

copperium (changes requested):
Awesome Raiyan!

Makefile
raiyansayeed: Interesting, I'm getting the opposite error "must install python3-venv". Should I write a note saying that you may need to run "sudo apt-get install python3-venv" or "sudo pip3 install virtualenv"?

README.md
raiyansayeed: I'll meet you halfway: we can append the virtual environment onto $(PATH) in the Makefile whenever a `make` command is run. This way it won't mess up other projects like CAN-Explorer or telemetry which have their own separate dependencies / virtualenvs.

Makefile
```diff
@@ -100,6 +100,9 @@ MPXE_PROTOS_DIR := $(MPXE_DIR)/protos
 DIRS := $(BUILD_DIR) $(BIN_DIR) $(STATIC_LIB_DIR) $(OBJ_CACHE) $(DEP_VAR_DIR)
 COMMA := ,
 
+# Virtualenv directory
+VENV_DIV := .venv
```
copperium, line 104: Is `DIV` a typo for `DIR`?

Makefile
```diff
@@ -194,16 +203,16 @@ IGNORE_CLEANUP_LIBS := CMSIS FreeRTOS STM32F0xx_StdPeriph_Driver unity FatFs mpx
 IGNORE_PY_FILES := ./lint.py ./libraries/unity ./.venv
```
copperium, line 203: This line should use `VENV_DIV` I believe (and also `LIB_DIR`)

Makefile
```diff
@@ -356,17 +365,19 @@ socketcan:
 	@sudo modprobe vcan
 	@sudo ip link add dev vcan0 type vcan || true
 	@sudo ip link set up vcan0 || true
-	@ip link show vcan0		
+	@ip link show vcan0
```
copperium, line 368: Could you remove all the tabs on the end of this line? :)

Makefile
copperium: Yeah sure, also put that in the Software 101 setup on confluence.

copperium (approved):
LGTM, just a few non-blocking nitpicks I noticed. If you want you can push a commit to fix the comments, otherwise I'll merge

PR #313, opened 2021-03-14
initial commit

KevinLee5656:
Let me know if the caching and the separation of make commands were done correctly 👍

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,16 +19,21 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # TODO: cache all the setup
-
       - name: Setup directories
         run: |
           # create directory that will be on the PATH
           mkdir -p ~/.local/bin
           echo "${HOME}/.local/bin" >> $GITHUB_PATH
           mkdir -p ~/source
+      
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
 
       - name: Install gcc, clang, clang-format
+        if: steps.cache.outputs.cache-hit != 'true'
         run: |
           sudo add-apt-repository ppa:ubuntu-toolchain-r/test
           sudo apt-get update
@@ -51,6 +56,7 @@ jobs:
           MAKE_PATH: make-4.1
           MAKE_ARCHIVE_PATH: make-4.1.tar.gz
           MAKE_URL: http://ftp.gnu.org/gnu/make/make-4.1.tar.gz
+        if: steps.cache.outputs.cache-hit != 'true'
         run: |
           wget -nv $MAKE_URL
           tar xvf $MAKE_ARCHIVE_PATH
@@ -66,6 +72,7 @@ jobs:
           python-version: '3.x'
 
       - name: Install Python tooling
+        if: steps.cache.outputs.cache-hit != 'true'
         run: |
           pip install --user pylint
           pip install --user autopep8
@@ -90,17 +97,34 @@ jobs:
           pylint --version
           autopep8 --version
 
-      - name: Format and lint
+      - name: Format
         run: |
           make test_format
+
+      - name: Lint
           make lint
+
+      - name: Pylint
           make pylint
 
-      - name: Build and test
+      - name: Build stm32f0xx
         run: |
           make build_all PLATFORM=stm32f0xx DEFINES="${DEFINES}"
           make clean
+
+      - name: Build x86
+        run: |
           make build_all PLATFORM=x86 DEFINE="${DEFINES}"
           make test_all PLATFORM=x86 DEFINE="${DEFINES}"
+
+      - name: Test x86
+        run: |
+          make test_all PLATFORM=x86 DEFINE="${DEFINES}"
+          
+      - name: Pytest
+        run: |
           make pytest_all
+
+      - name: Build x86 with clang
+        run: |
           make build_all PLATFORM=x86 COMPILER=clang DEFINE="${DEFINES}"


.github/workflows/main.yml
```diff
@@ -19,16 +19,21 @@ jobs:
...
           # create directory that will be on the PATH
           mkdir -p ~/.local/bin
           echo "${HOME}/.local/bin" >> $GITHUB_PATH
           mkdir -p ~/source
+      
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
copperium, line 33: This only caches pip dependencies eh? You'll need to use a separate cache step with a separate id for each step you want to skip based on caching. The general rule as I see it is, whatever directories need to be cached go in `path`, optionally line break separated if there's multiple of them:
```
path: |
  ~/foo
  ~/bar
```
Then the key should be created by concatenating the os (`${{ runner.os }}`), the name of the thing you're caching, and then `hashFiles` on any files that determine whether the cache should be invalidated. I.e., for `pip`, any requirements.txt files in the repo could change what dependencies get installed, so we call `hashFiles` on all the requirements.txt files.

.github/workflows/main.yml
```diff
@@ -90,17 +97,34 @@ jobs:
...
+          make test_all PLATFORM=x86 DEFINE="${DEFINES}"
+          
+      - name: Pytest
+        run: |
           make pytest_all
-          make build_all PLATFORM=x86 COMPILER=clang DEFINE="${DEFINES}"
+
+      - name: Build x86 with clang
+        run: |
+          make build_all PLATFORM=x86 COMPILER=clang DEFINE="${DEFINES}"
```
copperium, line 130: Please add a newline at the end of the file!

.github/workflows/main.yml
```diff
@@ -90,17 +97,34 @@ jobs:
...
-      - name: Build and test
+      - name: Build stm32f0xx
         run: |
           make build_all PLATFORM=stm32f0xx DEFINES="${DEFINES}"
           make clean
+
+      - name: Build x86
+        run: |
           make build_all PLATFORM=x86 DEFINE="${DEFINES}"
           make test_all PLATFORM=x86 DEFINE="${DEFINES}"
```
copperium, line 118: This line shouldn't be here eh?

copperium (changes requested):
Awesome job Kevin, the one other thing is please fix the status badge in the readme!

.github/workflows/main.yml
```diff
@@ -19,25 +19,77 @@ jobs:
...
+        run: sudo apt-fast -y install clang-format-5.0
+
+        if: steps.cache-libc6-i386.outputs.cache-hit != 'true'
+        run: sudo apt-fast -y install libc6-i386
+
+        # for vcan module
+        if: steps.cache-linux-module.outputs.cache-hit != 'true'
+        run: sudo apt-fast -y install linux-modules-extra-$(uname -r)
+
+        run: |
```
copperium, line 92: I'm fairly certain you can only have one `if` and one `run` per step - this is a limitation of the YAML format. So you'll need to split these into multiple steps, one per if/run.

.github/workflows/main.yml
```diff
@@ -19,25 +19,77 @@ jobs:
...
+        id: cache-linux-module
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-linux-module
+
+      - uses: actions/cache@v2
+        id: cache-make
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-make
```
copperium, line 63: I'm not sure where clang, clang-format, make, etc are installed, but it's definitely not in `~/.cache/pip`! You'll need to figure out where the actual clang, make, etc binaries get downloaded and installed to and use those paths. This might involve reading the source of any actions in a `uses` step (like the one that installs `arm-none-eabi-gcc`) or running the commands on your machine and figuring out where things got installed to.

Also I notice there's no cache step for the installing arm-none-eabi-gcc step (`uses: fiam-arm-none-eabi-gcc@v1`)? It installs the toolchain we need to build for STM32, and it takes almost two full minutes to do so - we'll definitely need to cache that step.

.github/workflows/main.yml
```diff
@@ -19,25 +19,77 @@ jobs:
...
         run: |
           # create directory that will be on the PATH
           mkdir -p ~/.local/bin
           echo "${HOME}/.local/bin" >> $GITHUB_PATH
           mkdir -p ~/source
+     
+      - uses: actions/cache@v2
+        id: cache-gcc
+        with:
+          path: ~/.local/bin/gcc
```
copperium, line 32: The only thing this script does in `~/.local/bin` is set up some links that take a negligible amount of time to set up.  I don't think you need to cache `~/.local/bin/gcc` - figure out where GCC is actually installed to and use that directory instead.

copperium (changes requested):

copperium:
FYI if you think the failures might be because you need to clear the caches, you can change the relevant cache keys (like add a `-v1-` or something).

copperium:
Ah, I meant the cache *key* rather than the id - the key is how it determines if it has a cache for a directory.

copperium:
Small additional request once you get the caching working: since you're updating the readme, could you update the paragraph about CI to talk about GitHub Actions instead of Travis? Thanks :)

copperium:
So I did a little bit of research, and here's a couple things you can try:
* For caching apt-get dependencies, try following https://raymii.org/s/articles/Github_Actions_cpp_boost_cmake_speedup.html - you might have to look up what `dpkg` etc do but they seem to have a good tutorial on caching apt-get. (Their stuff on using multiple cores to speed up compilation might be a good idea too but that's for another ticket.)
* But if you can't get that working but can get other things working like caching `make` and the arm-gcc-embedded toolchain, I think it would be a good idea to merge the things that work now to get faster builds and then try again with caching the rest later.

meganpark01:
Used dpkg for the apt-get dependencies

PR #314, opened 2021-03-15
102 and 103 HW

YiJie-Zhu:


projects/HW102/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+  Counters *storage = context;
+  storage->counter_b++;
+  LOG_DEBUG("%d\n", storage->counter_b);
+  soft_timer_start_millis(SECOND, increment_b, &storage, NULL);
+}
+
+int main() {
+  Counters counter = { 0 };
+  soft_timer_start_millis(HALF_SECOND, increment_a, &counter, NULL);
+  soft_timer_start_millis(SECOND, increment_b, &counter, NULL);
```
copperium, line 32: This is a nice solution, but the homework does say:
> Uses a **single soft timer** to increment `counter_a` every half a second...

projects/HW102/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+  LOG_DEBUG("%d\n", storage->counter_a);
+  soft_timer_start_millis(HALF_SECOND, increment_a, &storage, NULL);
+}
+
+void increment_b(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  storage->counter_b++;
+  LOG_DEBUG("%d\n", storage->counter_b);
+  soft_timer_start_millis(SECOND, increment_b, &storage, NULL);
+}
```
copperium, line 27: These are private functions, so according to the [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) they should be `static` and prefixed with `prv_`. The `static` marks the functions as local to the file; otherwise, if some other file declared an `increment_a` function, we might get linker errors.

projects/HW102/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define SECOND 1000
+#define HALF_SECOND 500
```
copperium, line 8: It's great that you factored these out into defines! Usually though we use names that don't just describe the constant itself but its purpose, so mentioning that it's a timeout or something would be better than just `SECOND`. The convention is also to add a suffix with the unit, so `_MS`.

projects/HW102/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+#include "interrupt.h"
```
copperium, line 2: Nitpick: the coding standards prescribe there should be a blank line between <standard> imports and "midsun-specific" imports.

projects/HW102/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+}
+
+void increment_b(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  storage->counter_b++;
+  LOG_DEBUG("%d\n", storage->counter_b);
+  soft_timer_start_millis(SECOND, increment_b, &storage, NULL);
+}
+
+int main() {
```
copperium, line 29: You'll need to initialize interrupts and soft timers here. You can also test your FW102 with `make run PLATFORM=x86 PROJECT=HW102`.

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+                                 .state = GPIO_STATE_LOW,
+                                 .alt_function = GPIO_ALTFN_NONE,
+                                 .resistor = GPIO_RES_NONE };
+
+InterruptSettings interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+void interrupt_handler(const GpioAddress *address, void *context) {
```
copperium, line 27: Again this is a private function so it should be `static` and prefixed with `prv_`.

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+  AdcChannel *reading_channel = (AdcChannel *)context;
```
copperium, line 28: You don't actually need the cast here, void pointers don't need cast when assigned to other pointer types in C (unlike in C++).

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+  uint16_t data = 0;
+  adc_read_raw(*reading_channel, &data);
```
copperium, line 30: The homework does say:
> Create a project that reads an ADC **converted** reading...

So use `adc_read_converted` instead. The difference is that `adc_read_converted` gives you output in mV, whereas `adc_read_raw` gives you whatever raw junk came out of the ADC. You almost always want `adc_read_converted`.

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+
+GpioSettings button_settings = { .direction = GPIO_DIR_IN,
+                                 .state = GPIO_STATE_LOW,
+                                 .alt_function = GPIO_ALTFN_NONE,
+                                 .resistor = GPIO_RES_NONE };
+
+InterruptSettings interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
```
copperium, line 25: These are static variables (file local), so according to the coding standards they should be `static` and prefixed with `s_`. However, I don't think any of these need to be global, so please declare them inside a function instead.

Global/static variables should generally be minimized for several reasons, one of which is that the microcontrollers we program on have only ~16KB of memory. Global variables take up some of that memory all the time, whereas local variables only take it up when the relevant function is running, so local variables are preferred.

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
+#include <stdint.h>
+#include "adc.h"
```
copperium, line 2: Again strictly there should be a blank line between these two lines.

projects/FW103/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+  AdcChannel reading_channel = NUM_ADC_CHANNELS;
+  adc_get_channel(reading_addr, &reading_channel);
+  adc_set_channel(reading_channel, true);
+  gpio_it_register_interrupt(&button_addr, &interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             interrupt_handler, &reading_channel);
+  while (true) {
+  }
+
+  return 0;
+}
```
copperium, line 52: Please put a newline at the end of your files! (`make format` should add it automatically, but if it doesn't please add it nonetheless)

copperium (changes requested):
Nice job, a few things to fix up

projects/FW103/src/main.c
copperium: Looks like you missed this comment, probably because it was collapsed by default:
> Again this is a private function so it should be static and prefixed with prv_.

copperium (changes requested):
Looks good mostly, just one comment on something you missed from previously. If you look at the checks, CI is failing because of linting errors - do `make format` to format your code and `make lint` to check for lint errors, and if any come up fix them.

copperium:
Not sure what happened with git that it's showing all kinds of changes, but the actual homework LGTM. I'll close this since the homework is complete, slack me tomorrow for a task!

PR #315, opened 2021-03-18
Fix mpxe issues 18 03 21

mitchellostler:
Just a few quick fixes. 
The one thing to mention is that `make pylint` seems to be moving the modules back up to the top of the file in pm.py, which causes the circular include error. Not sure how to add this to list of ignores?


jessm (approved):
LGTM 👍 sorry my code is like this

PR #317, opened 2021-03-19
Soft 999 {faraz khoubsriat} fw 102

farazkh80:
Completed FW102 homework.

### DIFF
--- a/projects/timer_fw102/src/main.c
+++ b/projects/timer_fw102/src/main.c
@@ -1,0 +1,44 @@
+#include <stdbool.h>
+#include "gpio.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timing_callback(SoftTimerId timer_id, void *context) {
+  Counters *counters = context;
+
+  counters->counter_a++;  // increments a
+
+  LOG_DEBUG("Counter A: %i\n", counters->counter_a);
+
+  if (counters->counter_a % 2 == 0) {  // increments b
+    counters->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", counters->counter_b);
+  }
+
+  // starts timer again after callback function completes
+  soft_timer_start_millis(500, prv_timing_callback, counters, NULL);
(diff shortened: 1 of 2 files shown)


projects/timer_fw102/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timing_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 13: This is a private function, so it should be `static` as per the [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards). `static` gives the function "internal linkage" (as opposed to regular "external linkage") which basically means that the function is local to the file only. Otherwise, if some other file defined a `prv_timing_callback` function, you might have issues.

projects/timer_fw102/src/main.c
```diff
@@ -0,0 +1,46 @@
...
+
+  LOG_DEBUG("Counter A: %i\n", counters->counter_a);
+
+  if (counters->counter_a % 2 == 0) {  // increments b
+    counters->counter_b++;
+    LOG_DEBUG("Counter B: %i\n", counters->counter_b);
+  }
+
+  // starts timer again
+  soft_timer_start_millis(500, prv_timing_callback, counters, NULL);
```
copperium, line 26: It's good practice to factor out the 500ms constant here and below into a #define'd constant so we're not repeating the constant and it's easier to change if necessary. Make sure to give it a good descriptive name and suffix it with the unit (like `_MS`) so we know what the unit is.

projects/timer_fw102/src/main.c
```diff
@@ -0,0 +1,46 @@
+#include <stdbool.h>
+#include "gpio.h"
```
copperium, line 2: Nitpick: the coding standards say there should be a blank line between <standard> includes and "midsun-specific" includes.

copperium (changes requested):
Nice job Faraz! Just a couple things to fix.

copperium:
Looks good! Feel free to move on to FW 103. Closing since homework is complete.

PR #318, opened 2021-03-20
[SOFT-447] Patch x86 CAN library to allow a message with a zero CAN ID

copperium:
On x86, the flow for receiving a CAN message is the following:
1. The RX thread (`prv_rx_thread` in can_hw.c) reads a CAN frame from the socket.
2. The RX thread calls a handler, which turns out to be `prv_rx_handler` in can.c.
3. `prv_rx_handler` calls `can_hw_receive` until it returns false. On x86, `can_hw_receive` will only rx one message (the most recently rx'd one), but it does this to support STM32 which can receive multiple in a row.
4. On the first call, `can_hw_receive` sees that it has a valid frame received, passes the data upwards, resets it, and returns true; on the second call, it sees that it doesn't have a valid frame and returns false.

Previously, we determined whether the frame was valid by assuming that a CAN arbitration ID of 0 would never happen, so the zero ID caused by memsetting the frame to 0 was checked. Now with the bootloader, an ID of 0 is possible (and likely), so we use a separate flag to determine if the frame is valid.

I don't *think* there are any race conditions caused by this change but @Max-MZ @jessm I'd appreciate other eyes as well. I have checked that babydriver `spi_exchange([1]*255, 255)` (exchanging >60 messages) works with this change, and also verified with `cansend` that this lets an ID 0 message get processed.

### DIFF
--- a/libraries/ms-common/src/x86/can_hw.c
+++ b/libraries/ms-common/src/x86/can_hw.c
@@ -89,7 +89,7 @@ static void *prv_rx_thread(void *arg) {
     if (FD_ISSET(s_socket_data.can_fd, &input_fds)) {
       int bytes =
           read(s_socket_data.can_fd, &s_socket_data.rx_frame, sizeof(s_socket_data.rx_frame));
-      s_socket_data.rx_frame_valid = true;
+      s_socket_data.rx_frame_valid = (bytes != -1);
 
       if (s_socket_data.handlers[CAN_HW_EVENT_MSG_RX].callback != NULL) {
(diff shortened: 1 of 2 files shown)


jessm (approved):
Makes sense, LGTM. Hopefully the test failure isn't too weird to sort out.

copperium:
Cause of the test failure was that I forgot that `read` can fail to read, so under rare (or not-so-rare) circumstances when restarting can_hw we'd get a read failure causing an invalid frame to be processed because I always set `rx_frame_valid` to true. Fixed by actually checking if `read` failed (it returns -1 on failure).

PR #320, opened 2021-03-24
completed FW103 HW

farazkh80:
Finished 103 HW, however, I was not sure about testing since there is no access to hardware. 

Also, not sure if the system was only supposed to register an interrupt or register and trigger an interrupt.


copperium:
Looks good to me! Technically you don't need the `gpio_it_trigger_interrupt` calls since those simulate an interrupt on the pin in software (usually used in unit tests), but they're alright for the homework since as you mentioned there's no real way to test it without hardware otherwise. Feel free to move on to FW 104 and/or ask me for a task.

PR #321, opened 2021-03-25
[Soft-345] startup conditions with unified pipes

mitchellostler:
Long time coming Mpxe init conditions PR:
Thanks to Jess's efforts in creating the unified pipes, and the design meeting we had, the init conditions and mpxe seem to both be working, and more stable. 

A store update object was created in project.py for easier passing between projects, very similar to the C-side.
The initial conditions are received via the poll thread. The main thread locks until a message of type MxCmd (newly created) is received with a command FINISH_INIT_COND. The init messages are stored as mx_update objects in a static array. When the driver calls store_register, it checks for a store with the same type and key, and if it exists it initializes with the conditions.

mcp23008 and pca9539r smoke tests have been changed to demonstrate that the changes work. They will only activate if MPXE is defined so they can stay or go


libraries/mpxe-store/src/store.c
```diff
@@ -50,8 +65,24 @@ Store *prv_get_first_empty() {
 
 static void prv_handle_store_update(uint8_t *buf, int64_t len) {
   MxStoreUpdate *update = mx_store_update__unpack(NULL, (size_t)len, buf);
-  s_func_table[update->type].update_store(update->msg, update->mask, (void *)update->key);
-  mx_store_update__free_unpacked(update, NULL);
+  if (update->type == MX_STORE_TYPE__CMD) {  // Abstract out into another s_func_table func?
+    MxCmd *cmd = mx_cmd__unpack(NULL, (size_t)update->msg.len, update->msg.data);
+    if (cmd->cmd == MX_CMD_TYPE__FINISH_INIT_CONDS) {
+      s_cmd_cb_lookup[MX_CMD_TYPE__FINISH_INIT_CONDS](NULL);
+    }
```
jessm, line 72: A few comments on this block:
1. You need to free anything you unpack, so at the end call `mx_cmd__free_unpacked(cmd, NULL)` (probably not necessary to set it to `NULL` since it goes out of scope)
2. In the second `if`, it's cleaner to do something like:
```
if (s_cmd_cb_lookup[cmd->cmd] != NULL) {
  s_cmd_cb_lookup[cmd->cmd](NULL);
}
```
Because this way it'll work for all command types rather than having a new `if` for each one. I'm not convinced we'll need context for these, but I'm ok leaving it there for now as just always `NULL`. In theory if we want to add more metadata to commands we can just pass the command instead of a `void *`, but we can worry about that later.

3. To apply comment 2 safely, you'll also need to add an `END` command type, similar to store types. This is so you can set `[END]` to `NULL` in `s_cmd_cb_lookup` to properly size the array, otherwise you could end up doing an out-of-bounds access. Also if you do this, you don't have to set all command callbacks to `NULL` because they'll default to `0` by being statically initialized.

mpxe/integration_tests/init_cond.py
```diff
@@ -0,0 +1,55 @@
+from mpxe.protogen import mcp23008_pb2, pca9539r_pb2
```
jessm, line 1: The functions in this file should be per integration test, so I don't think they should be in a separate file.

mpxe/integration_tests/test_mcp23008.py
```diff
@@ -3,12 +3,13 @@
...
 from mpxe.integration_tests import int_test
 from mpxe.sims.mcp23008 import Mcp23008, NUM_MCP_PINS
+from mpxe.integration_tests.init_cond import mcp23008_init_conditions
 
 
 class TestMcp23008(int_test.IntTest):
     def setUp(self):
         super().setUp()
-        self.mcp23008 = self.manager.start('smoke_mcp23008', Mcp23008())
+        self.mcp23008 = self.manager.start('smoke_mcp23008', Mcp23008(), mcp23008_init_conditions())
```
jessm, line 12: as mentioned, I think the init conditions should be kept locally in this file since it seems weird to have to look at other files to figure out what a test is doing. Happy to hear if you have other thoughts on this.

mpxe/integration_tests/test_pca9539r.py
```diff
@@ -3,12 +3,13 @@
...
 from mpxe.integration_tests import int_test
 from mpxe.sims.pca9539r import Pca9539r
+from mpxe.integration_tests.init_cond import pca9539r_init_conditions
 
 
 class TestPca9539r(int_test.IntTest):
     def setUp(self):
         super().setUp()
-        self.pca9539r = self.manager.start('smoke_pca9539r', Pca9539r())
+        self.pca9539r = self.manager.start('smoke_pca9539r', Pca9539r(), pca9539r_init_conditions())
```
jessm, line 12: same comment as test_mcp23008

mpxe/sims/mcp23008.py
```diff
@@ -27,7 +28,13 @@ def update_pin_state(self, proj, pin, state):
         mcp23008_mask = mcp23008_pb2.MxMcp23008Store()
         mcp23008_mask.state[pin] = 1
 
-        proj.write_store(mcp23008_msg, mcp23008_mask, stores_pb2.MxStoreType.MCP23008)
+        mcp23008_update = StoreUpdate(
```
jessm, line 31: nit: if you take out the `mcp23008_` prefix to msg mask and update, maybe this'll all fit on one line? I don't think the prefix is necessary

projects/smoke_mcp23008/src/main.c
```diff
@@ -68,6 +68,19 @@ static StatusCode prv_mcp23008_check_pin_states(Mcp23008GpioState state) {
   return STATUS_CODE_OK;
 }
 
+#ifdef MPXE
```
jessm, line 71: I'm not a fan of including these blocks here unless absolutely necessary, which I don't think these are. My philosophy with MPXE was to avoid touching non-x86-driver code as much as possible. It's fine for debugging purposes, but keeping these in prevents the smoketest from being as useful as it could be since smoketests should be straightforward to understand the code so we can clearly identify the expected behaviour.

projects/smoke_pca9539r/src/main.c
```diff
@@ -33,6 +34,27 @@ void setup_test(void) {
   };
   i2c_init(I2C_PORT, &i2c_settings);
   pca9539r_gpio_init(I2C_PORT, PCA9539_I2C_ADDRESS);
+#ifdef MPXE  // used to test double initial conditions
```
jessm, line 37: Same comment as smoke_mcp23008; I don't think this should be here.

jessm (changes requested):
Looking good! a few nits here and there, but I'm mostly just being picky, the logic all seems good. Thanks for sticking through with this!

mpxe/harness/pm.py
```diff
@@ -30,11 +32,17 @@ def __init__(self):
...
-    def start(self, name, sim=None):
+    def start(self, name, sim=None, init_conds=None):
         if name not in self.proj_name_list:
             raise ValueError('invalid project "{}": expected something from projects directory')
         proj = project.Project(name, sim or Sim())
         self.fd_to_proj[proj.popen.stdout.fileno()] = proj
+        if init_conds:
+            for update in init_conds:
+                proj.write_store(update)
+                sleep(0.1)
```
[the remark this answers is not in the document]
mitchellostler: Yeah, for whatever reason both messages won't be either sent or received? When I was testing without the sleep, the C-side would only receive the command message, and not the init conditions. Not sure why this is an issue though, I would think that it should be fine without it

projects/smoke_mcp23008/src/main.c
mitchellostler: The additions to the smoke tests were just to verify that the initial conditions were working. I was thinking it might be good to have them as a demo for how to do initial conditions when the tests actually get written. 

I have tried to make them as inert as possible here, and I can add a comment to say that these are mpxe specific, or I can just remove them and add documentation for how to configure init conds.

I could also create a new project `mpxe_init_conds` or something like that that contains the functionality that I added in the smoke tests with maybe some expanded usage, and write an mpxe integration test for it to show how to setup initial conditions. This might also be a good longer term investment because then we could always run it to verify that mpxe initial conditions is working?

libraries/mpxe-store/src/store.c
```diff
@@ -122,6 +158,18 @@ void store_register(MxStoreType type, StoreFuncs funcs, void *store, void *key)
   local_store->type = type;
   local_store->store = store;
   local_store->key = key;
+
+  for (uint16_t i = 0; i < MAX_STORE_COUNT;
+       i++) {  // Need to check each index, as freed and set to NULL once used
+    if (s_init_cond[i] != NULL) {
+      if (s_init_cond[i]->type == type && (void *)(intptr_t)s_init_cond[i]->key == key) {
+        s_func_table[s_init_cond[i]->type].update_store(s_init_cond[i]->msg, s_init_cond[i]->mask,
```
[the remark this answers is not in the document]
mitchellostler: Much cleaner lmao thanks

projects/smoke_mcp23008/src/main.c
jessm: I think docs would be sufficient, but I'm also good with an extra simple `mpxe_init_conds` project if you want to create that. I'm pretty much of the opinion that the MPXE integration tests that use init conds are the living documentation, so I'm also ok if you just don't write docs since we might change the way init conds work and then we'd have to update docs.

For better or for worse this is how most of our codebase is - we just point people to live examples rather than demos or docs.

mpxe/harness/pm.py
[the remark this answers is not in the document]
jessm: I think this is because after we removed the ctop_fifo, there's no more lock on python to wait until C starts polling. Can you add something to synchronize? i.e. after C starts the poll thread, send a message to python saying C is ready, and in python wait for that message before sending the init conds. I think it makes sense to do this via a lock in this python file, and command handling in python similar to how it's done in C in the poll thread, but I'm sure there's plenty of other solutions you could explore.

projects/smoke_mcp23008/src/main.c
mitchellostler: Yeah cool, I'll add a project so that it can be seen how it works

copperium (commented):
I agree with all of Jess's points, thanks for doing this!

Perhaps the next step is to use your `MxCmd` framework to add commands to start up/shut down other projects for PD?

libraries/mpxe-store/src/store.c
```diff
@@ -50,8 +65,24 @@ Store *prv_get_first_empty() {
 
 static void prv_handle_store_update(uint8_t *buf, int64_t len) {
   MxStoreUpdate *update = mx_store_update__unpack(NULL, (size_t)len, buf);
```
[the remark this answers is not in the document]
mitchellostler: Update objects are stored until we use them in store_register, and then they are freed after the driver receives them

mpxe/harness/pm.py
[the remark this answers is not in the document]
mitchellostler: I added a locking mechanism but it didn't resolve the issue. It's in the latest commit so you can have a look. I don't think it has to do with the polling thread not being available, I think it has to do with the behaviour of the pipe itself?

 The sleep() needs to happen between sequential messages (including more than one init cond) for all messages to be received properly. Otherwise some get lost. If I remove the sleeps, for whatever reason it will still always receive the command msg, but it will not receive any initial conditions.

libraries/mpxe-store/src/store.c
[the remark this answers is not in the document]
copperium: That's only for non-command updates though right? It doesn't look like the update object is stored in the `update->type == MX_STORE_TYPE__CMD` branch.

jessm (approved):

PR #322, opened 2021-03-25
[Soft 186] fixed block comment

YiJie-Zhu:


projects/mci/inc/motor_can.h
```diff
@@ -54,46 +51,42 @@ typedef uint32_t MotorCanFrameId;
...
-   * Scale: 1
-   * Offset: 0
-   */
+  // Desired motor velocity set point in metres/second.
+
+  // Range: -
+  // Scale: 1
+  // Offset: 0
+
   float motor_velocity;
```
copperium, line 73: IMO there shouldn't be all the blank lines in the comments, please remove the blank lines just before each of the variables/structs, and either put a `//` on the blank lines within comments or remove the blank lines.

copperium (changes requested):
Thanks for getting this out of the way so quickly :) Just one nitpick, then it should be good to go.

copperium (approved):

PR #326, opened 2021-03-29
Soft 443 fix bps heartbeat

AveryChiu64:


libraries/ms-common/inc/ms_test_helpers.h
```diff
@@ -87,3 +87,8 @@
                        .data = (ack_status) };                                                     \
     TEST_ASSERT_OK(can_ack_handle_msg(&(can_storage).ack_requests, &msg));                         \
   })
+
+// Returns STATUS_CODE_OK if the *value_to_test == value_to_compare within a certain time +- the
+// tolerance
+StatusCode time_assert(uint8_t value_to_compare, uint8_t *value_to_test, double time_ms,
+                       double tolerance_ms);
```
copperium, line 94: I think this function should be prefixed with the module name (`ms_test_helpers_time_assert`) since just seeing `time_assert` in a test you wouldn't know where it came from. Also have you thought about having a `TEST_ASSERT_EQUAL_AT_TIME` macro (or some similar name) that just does `TEST_ASSERT_OK(time_assert(...))`, since that seems to be the common pattern?

Also, since `*value_to_test` changes due to another thread, it should be marked `volatile` (`volatile uint8_t *value_to_test`) so that the compiler doesn't optimize it out.

libraries/ms-common/src/ms_test_helpers.c
```diff
@@ -0,0 +1,24 @@
...
+#include "status.h"
+
+#define CLOCK_HEADSTART 30000
+
+StatusCode time_assert(uint8_t value_to_compare, uint8_t *value_to_test, double time_ms,
+                       double tolerance_ms) {
+  double elapsed = 0;
+  double upper_bound = (double)(time_ms + tolerance_ms) / 1000.0;
+  double lower_bound = (double)(time_ms - tolerance_ms) / 1000.0;
+  clock_t begin = clock() - CLOCK_HEADSTART;
```
copperium, line 15: What's `CLOCK_HEADSTART`, is that the approx number of clock ticks that it takes to start up? Seems like something that should be part of the tolerance I think. Also perhaps the tolerance should be a constant so we don't have to specify it everywhere? I can't see it really changing too much between tests.

libraries/ms-common/src/ms_test_helpers.c
```diff
@@ -0,0 +1,24 @@
...
+  while (elapsed < upper_bound) {
+    elapsed = ((double)(clock() - begin)) / (CLOCKS_PER_SEC);
+    LOG_DEBUG("value_to_test = %d @ elapsed = %f\n", *value_to_test, elapsed);
```
copperium, line 18: This is going to spam the test logs real hard, please remove.

libraries/ms-common/src/ms_test_helpers.c
```diff
@@ -0,0 +1,24 @@
+#include "ms_test_helpers.h"
+#include <string.h>
+#include <time.h>
+#include "log.h"
+#include "ms_test_helper_can.h"
+#include "status.h"
+
+#define CLOCK_HEADSTART 30000
+
+StatusCode time_assert(uint8_t value_to_compare, uint8_t *value_to_test, double time_ms,
```
copperium, line 10: `*value_to_test` should be marked `volatile` here too

copperium (changes requested):
Seems valuable, I'm not super convinced that the problem is solved though - it works on my machine, but CI for this PR is currently failing on this test...

copperium (approved):

PR #327, opened 2021-03-31
Soft 425 solar fan control tests

PurpleRhythms:
1. Created test_fan_control_solar.c for testing fan_control_solar.c
2. Fixed merge conflicts in solar.config.c 
3. Removed redundant line in fan_control_solar.c


PurpleRhythms:
Reopening pull request for auto-build.

projects/solar/inc/fan_control_solar.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <stdbool.h>
```
copperium, line 3: By convention, all our header files start with a comment like this:
```c
// A short description of the module.
// Requires X, Y, and Z to be initialized.

// An optional extended description of the module.
```
This helps knowing what each module does, and the `// Requires ... to be initialized` comments should list all the modules/libraries it depends on (including transitively - if A requires B and B requires C then A should list C in the requirement list) to help with initializing everything that's required for tests.

(Also, there should be a blank line after the <stdbool.h> include according to the coding standards.)

projects/solar/src/solar_config.c
```diff
@@ -19,28 +19,21 @@
...
-#define SOLAR_MPPT_CURRENT_SCALING_FACTOR 15220.0f
+// Must be calibrated.
+#define SOLAR_MPPT_CURRENT_SCALING_FACTOR 1.0f
 
-// Scaling factor to convert SPV1020 input voltage values from SPI to millivolts.
-// Experimentally determined to be ~26.1mV/LSB.
-#define SOLAR_MPPT_VIN_SCALING_FACTOR 26.1f
+// Scaling factor to convert SPV1020 VIN values from SPI to millivolts.
+// Must be calibrated.
+#define SOLAR_MPPT_VIN_SCALING_FACTOR 1.0f
```
copperium, line 36: I think this was a merge conflict resolution, but please put this back to what it is on master, there's a lot of experimentally-determined parameters in there

projects/solar/src/solar_config.c
```diff
@@ -182,7 +175,6 @@ static SenseMcp3427Settings s_sense_mcp3427_settings = {
           {
               .data_point = DATA_POINT_CURRENT,
               .scaling_factor = SOLAR_MCP3427_CURRENT_SENSE_SCALING_FACTOR,
-              .bias = SOLAR_MCP3427_CURRENT_SENSE_BIAS,
```
copperium, line 185: Please put this back too

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+
+static GpioState s_gpio_state;
+static Event s_e;
+
+static GpioAddress s_dummy;
+
+static void TEST_MOCK(prv_fanfail_callback)(const GpioAddress *address, void *context) {
+  LOG_WARN("fan_control detected fanfail, raising fault event");
+  fault_handler_raise_fault(EE_SOLAR_FAULT_FAN_FAIL, 0);
+}
```
copperium, line 38: You shouldn't be mocking private functions (and it wouldn't work anyways), please remove this.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+
+  // Test Board A with 6 MPPTs uses virtual port A
+  GpioAddress s_overtemp_addr = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress s_fan_fail_addr = { .port = GPIO_PORT_A, .pin = 1 };
+  GpioAddress s_full_speed_addr = { .port = GPIO_PORT_A, .pin = 2 };
+
+  // arbitrarily selected value for testing
+  uint16_t s_full_speed_temp_threshold = 3;
+
+  SolarMpptCount s_mppt_count = SOLAR_BOARD_6_MPPTS;
```
copperium, line 54: This isn't setting the values of the static variables, it's declaring local variables of the same name that shadow the global variables - don't use the type name if you want to reset the variables. But I think these are meant to be constants, so you can just set them when you declare them at the top level, and then you don't have to worry about setting them in setup_test.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// Used to reinitialize static testing variables Event s_e & GpioState s_gpio_state
+void teardown_test(void) {
+  static GpioState s_gpio_state;
+  static Event s_e;
```
copperium, line 60: This isn't resetting the static top-level variables - instead it's declaring function-local static variables (https://stackoverflow.com/a/572550) which shadow the top-level static variables. Instead, assign to the variables to reset them.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// Test that "overtemperature" & "fan fail" faults are raised properly
```
copperium, line 63: I love the thorough commenting!

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+static void TEST_MOCK(prv_fanfail_callback)(const GpioAddress *address, void *context) {
+  LOG_WARN("fan_control detected fanfail, raising fault event");
+  fault_handler_raise_fault(EE_SOLAR_FAULT_FAN_FAIL, 0);
+}
+
+void setup_test(void) {
+  event_queue_init();
+  gpio_init();
+  gpio_it_init();
+  data_store_init();
```
copperium, line 44: You also need to call `interrupt_init` (from interrupt.h) since gpio_it depends on it

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  // Test that there are no faults initially
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+
+  // Test that fan fail fault is raised correctly
+  // The MAX6643 will call the prv_callback function when fan fail is detected
+  // Parameters have no effect on function
+  prv_fanfail_callback(&s_dummy);
```
copperium, line 81: You shouldn't touch private functions within a unit test - you should only be testing the public interface of your module. I'd just do a `gpio_it_trigger_interrupt` on the relevant pin and trust that the MAX6643 driver works.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+  event_process(&s_e);
+  MS_TEST_HELPER_ASSERT_EVENT(s_e, EE_SOLAR_FAULT_FAN_FAIL, 0);
```
copperium, line 83: You don't need to `event_process` to get the next event before passing it to `MS_TEST_HELPER_ASSERT_EVENT` - if you take a look at the definition inside ms_test_helpers.h, `MS_TEST_HELPER_ASSERT_EVENT` does `event_process` on the event you pass itself.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+    .fan_fail_addr = s_fan_fail_addr,
+    .full_speed_addr = s_full_speed_addr,
+    .full_speed_temp_threshold = s_full_speed_temp_threshold,
+    .mppt_count = s_mppt_count,
+  };
+
+  // Ensure fan_control_init works
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  for (Mppt i = 1; i <= 6; i++) {
```
copperium, line 100: The MPPTs are 0-indexed, you should instead iterate like `for (Mppt i = 0; i < 6; i++)`. This also goes for everywhere else you do this.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+    data_store_set(DATA_POINT_TEMPERATURE(i), 2);
```
copperium, line 101: This `2` should probably be abstracted into a `#define BELOW_THRESHOLD_TEMPERATURE_DC 2` (the DC is for deciCelsius), currently it's not super clear what it represents.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+  }
+  data_store_done();
```
copperium, line 103: `data_store_done()` just raises the DATA_READY_EVENT, you need to call `fan_control_process_event` with an event with ID DATA_READY_EVENT. (You don't actually need to call `data_store_done` at all, just pass an event you construct.)

This also goes for all the rest of the tests.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+static GpioAddress s_overtemp_addr;
+static GpioAddress s_fan_fail_addr;
+static GpioAddress s_full_speed_addr;
+static uint16_t s_full_speed_temp_threshold;
+static SolarMpptCount s_mppt_count;
+
+static GpioState s_gpio_state;
+static Event s_e;
+
+static GpioAddress s_dummy;
```
copperium, line 33: The usual practice for these kinds of variables is to declare them locally in each function, that way you don't have to worry about state carrying over between functions (or the weird `s_e` variable name). Also, `s_dummy` won't be needed after you make the `gpio_it_trigger_interrupt` change.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+    .full_speed_addr = s_full_speed_addr,
+    .full_speed_temp_threshold = s_full_speed_temp_threshold,
+    .mppt_count = s_mppt_count,
+  };
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  for (Mppt i = 1; i <= 5; i++) {
+    data_store_set(DATA_POINT_TEMPERATURE(i), 2);
+  }
+  data_store_set(DATA_POINT_TEMPERATURE(6), 4);
```
copperium, line 127: Again the `4` should be abstracted into a `#define ABOVE_THRESHOLD_TEMPERATURE_DC 4` to make it easier to read

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+  data_store_done();
+
+  // Overtemperature fault should be raised
+  TEST_ASSERT_OK(event_process(&s_e));  // Make sure an event exists
```
copperium, line 131: Again you don't need this, `MS_TEST_HELPER_ASSERT_EVENT` does it

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+    .overtemp_addr = s_overtemp_addr,
+    .fan_fail_addr = s_fan_fail_addr,
+    .full_speed_addr = s_full_speed_addr,
+    .full_speed_temp_threshold = s_full_speed_temp_threshold,
+    .mppt_count = s_mppt_count,
+  };
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  for (Mppt i = 1; i <= 6; i++) {
+    data_store_set(DATA_POINT_TEMPERATURE(i), 5);
```
copperium, line 152: Use the `ABOVE_THRESHOLD_TEMPERATURE_DC` macro here

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,234 @@
...
+    .overtemp_addr = s_overtemp_addr,
+    .fan_fail_addr = s_fan_fail_addr,
+    .full_speed_addr = s_full_speed_addr,
+    .full_speed_temp_threshold = s_full_speed_temp_threshold,
+    .mppt_count = s_mppt_count,
+  };
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  // Set mppt(6)'s status to overtemperature
+  data_store_set(DATA_POINT_MPPT_STATUS(6), 2);  // "0000010"
```
copperium, line 177: Aside from this needing to be zero indexed, you should also use `SPV1020_OVT_MASK` (OVT == overtemperature) from spv1020_mppt_defs.h instead of the raw number

copperium (changes requested):
Awesome job Daniel, I love the thoroughness of the tests! There's a bunch of things to fix up, both in the module itself and the tests; even though you didn't write the module, you're sort of the owner of it now, so please address those comments too.

copperium (changes requested):

copperium:
CI is failing because now there are no changes at all in solar_config.c, so the `config_get_fan_control_settings` function doesn't have a definition.

PurpleRhythms:
Reminder to self: GitHub no longer merges changes to `solar_config.h`, `solar_config.c`, `test_solar_config.c`, and `main.c` using `git pull` so you will need to do this every time before pushing.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,342 @@
...
+  TEST_ASSERT_EQUAL(STATUS_CODE_INVALID_ARGS, fan_control_init(&invalid_settings));
+}
+
+// Test that processing a non DATA_READY_EVENT returns false (fails gracefully)
+void test_not_data_ready_event(void) {
+  // NULL passed in
+  TEST_ASSERT_FALSE(fan_control_process_event(NULL));
+
+  // Non DATA_READY_EVENT passed in
+  Event e;
```
copperium, line 340: Could you initialize this to something? Like `Event e = { DATA_READY_EVENT + 1, 0 }`. Otherwise `e` has a value of uninitialized garbage, which could in theory occasionally make `e.id` by chance get initialized to `DATA_READY_EVENT`'s value.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,342 @@
...
+void test_fan_fail_fault_handling(void) {
+  FanControlSolarSettings settings = {
+    .overtemp_addr = s_overtemp_addr,
+    .fan_fail_addr = s_fan_fail_addr,
+    .full_speed_addr = s_full_speed_addr,
+    .full_speed_temp_threshold_dC = s_full_speed_temp_threshold,
+    .mppt_count = s_mppt_count,
+  };
+
+  TEST_ASSERT_OK(fan_control_init(&settings));
```
copperium, line 94: Since you do this in almost every test (and for the other two tests it doesn't matter if it's initialized), you can move this to `setup_test`.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,342 @@
...
+    .mppt_count = s_mppt_count,
+  };
+
+  // Ensure fan_control_init works
+  TEST_ASSERT_OK(fan_control_init(&settings));
+
+  // Initialize all mppt statuses to nominal
+  for (int i = 0; i < 6; i++) {
+    data_store_set(DATA_POINT_MPPT_STATUS(i), SPV1020_NOMINAL_MASK);
+  }
```
copperium, line 151: You shouldn't have to do this (here and in all the other tests) once you implement checking whether the mppt status is set in the data store before accessing it. `data_store_init` resets all the is-set flags to false, so you don't need to manually reset the statuses.

projects/solar/test/test_fan_control_solar.c
```diff
@@ -0,0 +1,342 @@
...
+  // too many MPPTs test
+  GpioAddress overtemp_addr = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress fan_fail_addr = { .port = GPIO_PORT_A, .pin = 1 };
+  GpioAddress full_speed_addr = { .port = GPIO_PORT_A, .pin = 2 };
+
+  FanControlSolarSettings invalid_settings = {
+    .overtemp_addr = overtemp_addr,
+    .fan_fail_addr = fan_fail_addr,
+    .full_speed_addr = full_speed_addr,
+    .full_speed_temp_threshold_dC = 4,
```
copperium, line 328: You can just use `s_overtemp_addr`, `s_fan_fail_addr`, `s_full_speed_addr`, `s_full_speed_temp_threshold` here eh?

copperium (changes requested):
Only thing that's blocking is the comment about checking whether the mppt status is set before using it, the rest are mostly nitpicks and typos. Nothing should take very long to address, then we can get this merged!

copperium (approved):

PR #59, opened 2020-04-18
Soft 147 centre console main file

ArshanKhanifar:

PR #328, opened 2021-04-01
[SOFT-233] Unflake CAN and some tests

jessm:
## Why am I doing this
This PR aims to shortcut socketcan when CAN is configured in loopback mode. We want to do this because there's an unfortunately long delay between when a message is TXed and RXed over socketcan, sometimes on the order of hundreds of milliseconds. This issue is exacerbated by the 3 thread model we use for CAN, in which the TX thread can take a while to wake up before TXing a message, leading to an even longer delay.

Some tests rely on CAN delay being deterministic, and occasionally fail when a message takes too long. This causes much grief and hurts developer velocity.

As much as I wanted this to be a clean change in just can_hw.c, some tests relied on CAN taking a non-zero amount of time to process a message. I believe this reliance is incorrect, since it causes non-determinism and doesn't account for what's really going on.

## The play by play
Probably best to be looking at the file I'm talking about at the same time as reading this section.
File by file:
### can_hw.c
If we init can with loopback mode, we don't bother with the threads at all, just copy directly from the TX fifo to the RX frame, and call the TX/RX callbacks sequentially. To account for filters, we apply them manually between TX and RX callbacks to mimic how it'd normally work. We also don't apply the filters to the socket, since we never open the socket and that errors out.

### test_can_hw.c
This test previously relied on CAN to have a delay between sending and receiving a message. Upon calling `can_hw_transmit()` (e.g. on line 59 in `test_can_hw_loop()`), the test would jump to `prv_wait_rx()`, check the number of messages received, then wait until it incremented. With the shortcut, the message would get received immediately after `can_hw_transmit()` was called, meaning the increment happened before we were waiting for it. Instead,  this PR sets a number of messages received to wait for explicitly, rather than an increment.

### test_spi_exchange.c
This test wasn't _that_ flaky, but it became really flaky when I shortened the timing, suggesting investigation was necessary.
This test relied on CAN being slower than a single callback (2 ms), but faster than two callbacks (4 ms). This meant the old flow was:
1. Send data over CAN, and TX/RX it via `MS_TEST_HELPER...`
2. TX, then spin while waiting for RX
3. Repeat for the second callback
You'd think that this process should be fine since now TX/RX should be faster so it should be fine to spin, but my theory is that with a really short delay, the callback would fire twice before a single event was RX/TXed, meaning there would be two TX events in the queue, and `MS_TEST_HELPER_CAN_TX_RX()` would fail. I changed this to delay appropriately for all callbacks to fire, then process all the TX events, then process all the RX events. This adds some determinism and lets us cut down the timings without worrying about <10ms delays messing us up.

### DIFF
--- a/libraries/ms-common/src/x86/can_hw.c
+++ b/libraries/ms-common/src/x86/can_hw.c
@@ -45,6 +45,7 @@ typedef struct CanHwSocketData {
   size_t num_filters;
   CanHwEventHandler handlers[NUM_CAN_HW_EVENTS];
   uint32_t delay_us;
+  int loopback;
 } CanHwSocketData;
 
 static pthread_t s_rx_pthread_id;
@@ -155,15 +156,21 @@ StatusCode can_hw_init(const CanHwSettings *settings) {
     pthread_join(s_tx_pthread_id, NULL);
   }
 
+  memset(&s_socket_data, 0, sizeof(s_socket_data));
+  s_socket_data.can_fd = -1;
+  s_socket_data.loopback = settings->loopback;
+  s_socket_data.delay_us = prv_get_delay(settings->bitrate);
+  fifo_init(&s_socket_data.tx_fifo, s_socket_data.tx_frames);
+
+  if (s_socket_data.loopback) {
+    return STATUS_CODE_OK;
+  }
+
   pthread_mutex_init(&s_keep_alive, NULL);
   // Init semaphore to thread-shared and locked
   sem_init(&s_tx_sem, 0, 0);
 
   pthread_mutex_lock(&s_keep_alive);
-
-  memset(&s_socket_data, 0, sizeof(s_socket_data));
-  s_socket_data.delay_us = prv_get_delay(settings->bitrate);
-  fifo_init(&s_socket_data.tx_fifo, s_socket_data.tx_frames);
 
   s_socket_data.can_fd = socket(PF_CAN, SOCK_RAW, CAN_RAW);
   if (s_socket_data.can_fd == -1) {
@@ -172,9 +179,8 @@ StatusCode can_hw_init(const CanHwSettings *settings) {
   }
 
   // Loopback - expects to receive its own messages
-  int loopback = settings->loopback;
-  if (setsockopt(s_socket_data.can_fd, SOL_CAN_RAW, CAN_RAW_RECV_OWN_MSGS, &loopback,
-                 sizeof(loopback)) < 0) {
+  if (setsockopt(s_socket_data.can_fd, SOL_CAN_RAW, CAN_RAW_RECV_OWN_MSGS, &s_socket_data.loopback,
+                 sizeof(s_socket_data.loopback)) < 0) {
     LOG_CRITICAL("CAN HW: Failed to set loopback mode on socket\n");
     return status_msg(STATUS_CODE_INTERNAL_ERROR, "CAN HW: Failed to set loopback mode on socket");
   }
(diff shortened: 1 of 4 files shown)


projects/baby_driver/test/test_spi_exchange.c
```diff
@@ -57,6 +61,22 @@ static SpiSettings s_spi_settings_test_2 = { .baudrate = 6000000,
...
+  // at once.
+
+  // The +1 is to account for the initial delay before the module starts TXing
+  delay_us((uint32_t)(TEST_SPI_EXCHANGE_DELAY_US * (num_messages + 1)));
+  for (uint16_t i = 0; i < num_messages; i++) {
+    MS_TEST_HELPER_CAN_TX(TEST_CAN_EVENT_TX);
+  }
+  for (uint16_t i = 0; i < num_messages; i++) {
+    MS_TEST_HELPER_CAN_RX(TEST_CAN_EVENT_RX);
+  }
```
copperium, line 77: This is still flaky on my machine - I think it might just be the very short length of the delay, causing the actual code execution to take a non trivial amount of time. On my machine lengthening `TEST_SPI_EXCHANGE_DELAY_US` by a couple orders of magnitude fixed it.

libraries/ms-common/src/x86/can_hw.c
```diff
@@ -260,8 +270,37 @@ StatusCode can_hw_transmit(uint32_t id, bool extended, const uint8_t *data, size
     // Fifo is full
     return status_msg(STATUS_CODE_RESOURCE_EXHAUSTED, "CAN HW TX failed");
   }
-  // Unblock TX thread
-  sem_post(&s_tx_sem);
```
copperium, line 264: I think we should still TX over vcan when loopback is enabled, just without enabling loopback on the socket itself, as long as it doesn't impact the flakiness of the tests. It's better hardware emulation that way since stm32 still txes on the CAN network in loopback mode. Do you think this is possible? If not we should put a warning by the `loopback` field in can.h saying that we don't tx over vcan while in loopback mode on x86.

copperium (changes requested):
🙏 

Just noticed a couple inconsistencies.

libraries/ms-common/src/x86/can_hw.c
```diff
@@ -260,8 +270,37 @@ StatusCode can_hw_transmit(uint32_t id, bool extended, const uint8_t *data, size
...
+    s_socket_data.rx_frame_valid = true;
+
+    // Apply filters that would normally be applied within socketcan
+    for (size_t i = 0; i < s_socket_data.num_filters; i++) {
+      struct can_filter filter = s_socket_data.filters[i];
+      uint32_t filt_id = extended ? id | CAN_EFF_FLAG : id;
+      if ((filt_id & filter.can_mask) != (filter.can_id & filter.can_mask)) {
+        return STATUS_CODE_OK;
+      }
+    }
```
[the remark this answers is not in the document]
jessm: Updated manual filter logic and updated `test_can_filter` to use two filters and hit on the second filter

projects/baby_driver/test/test_spi_exchange.c
jessm: Hmm, I changed it back to what it was originally. Since that's now a soft timer issue not a CAN issue I'll address it soon, but not here

libraries/ms-common/src/x86/can_hw.c
jessm: Makes sense. Writing to the socket itself is pretty fast, but I was doing some benchmarking and it seems like doing it over the thread can take up to 10 ms(!) so I just threw in a `write()` call within the short circuit logic. I also updated the RX thread to kill itself, since I think the constant `select` timeouts doesn't help determinism.

copperium: We should still be able to RX external messages while in loopback though, since that's a thing stm32 can do. E.g. for some reason loopback is enabled on centre console on master (?? thanks arshan), and we should still be able to send messages to it, as long as it doesn't cause the tests to be flaky of course.

jessm: That's not actually true. See https://www.st.com/resource/en/reference_manual/dm00031936-stm32f0x1stm32f0x2stm32f0x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf, page 819, section 29.5.2 Loop back mode:
> The actual value of the CANRX input pin is disregarded by the bxCAN. The transmitted messages can be monitored on the CANTX pin.

Loopback is enabled on centre_console on master because Arshan made a mistake a year ago and no one caught it: https://github.com/uw-midsun/firmware_xiv/pull/59
We were both tagged on that review so we're both guilty for that :) I fixed this on mockup, but I haven't merged mockup centre_console changes to master yet.

copperium: Oh interesting! Seems good then.

copperium (approved):

PR #329, opened 2021-04-02
[SOFT 453] Updated steering to rev 4.0

YiJie-Zhu:


projects/steering/src/steering_can.c
```diff
@@ -7,16 +7,18 @@
 #include "steering_digital_input.h"
 #include "steering_events.h"
 
+static bool enable_regen_braking = true;
```
copperium, line 10: Nitpick: the [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) say that static variables should be prefixed with `s_`, so this should be `s_enable_regen_braking`. (Sorry I know I gave you the variable name!)

copperium (changes requested):
Awesome job Jackie! There's just one thing you missed, it's the typo-fixing at the end of the ticket :)

projects/steering/src/steering_can.c
```diff
@@ -33,6 +35,11 @@ StatusCode steering_can_process_event(Event *e) {
     case STEERING_CONTROL_STALK_EVENT_RIGHT_SIGNAL:
       CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_SIGNAL_RIGHT, (EELightState)e->data);
       break;
+    case STEERING_REGEN_BRAKE:
```
copperium, line 38: Also, this should be `STEERING_REGEN_BRAKE_EVENT`, currently `STEERING_REGEN_BRAKE` is undefined which is causing the build to fail.

copperium (approved):

PR #331, opened 2021-04-02
[SOFT-233] Deterministic soft timers

jessm:
For the soft timers, this changes the implementation from array-based individual timers to list-based backed by a single timer, similar to how the stm32 version works.

Some tests still struggled. I found that busy-looping would occasionally miss timer events, but adding a `wait()` seemed to help. This seems worth having regardless of whether it helps consistency or not, so I threw a bunch of waits around.

### DIFF
--- a/libraries/ms-common/src/x86/soft_timer.c
+++ b/libraries/ms-common/src/x86/soft_timer.c
@@ -1,58 +1,111 @@
+// Timers are implemented as a doubly linked list, sorted in order of remaining
+// time until expiry. Thus, the head of the list is always defined as the next
+// timer to expire, so we back the list with a linux interval timer that's always
+// set to the head's remaining time.
 #include "soft_timer.h"
 
 #include <signal.h>
-#include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
-#include <stdio.h>
+#include <string.h>
 #include <time.h>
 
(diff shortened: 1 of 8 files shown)


libraries/ms-common/src/x86/soft_timer.c
```diff
@@ -1,58 +1,111 @@
...
+
+// Backing array for timer list objpool
+static volatile SoftTimer s_storage[SOFT_TIMER_MAX_TIMERS] = { 0 };
+
+// Set the linux timer to a new value. Checks for divide by zero.
+static void prv_set_timer(uint32_t val_us) {
+  // See timer_settime manual page for details
+  struct itimerspec spec = {
+    { 0, 0 },                                                                 //
+    { val_us ? val_us / 1000000 : 0, val_us ? val_us % 1000000 * 1000 : 0 },  //
```
copperium, line 46: I don't think there's any divide-by-zero danger here lol, the ternaries seem redundant.

copperium (approved):
LGTM, I think this is the first time I've ever seen `make test_all PLATFORM=x86` finish without errors on my system.

libraries/ms-common/src/x86/soft_timer.c
jessm: lol im dumb, thanks for that

PR #332, opened 2021-04-05
[SOFT-458] Add BUG macro for initialization debuggability

copperium:
In https://github.com/uw-midsun/firmware_xiv/pull/330 I added a `VERIFY_INIT` macro to make the PD main nicer, and Jess suggested it would be good as a libcore module. So here it is.

`BUG` wraps a call and `LOG_CRITICAL`s something like this when the call returns non-OK:
```
[2] projects/power_distribution/src/main.c:153:
****************** BUG TRIGGERED, ABORTING PROGRAM ******************
This indicates initialization failure or another unrecoverable fault.
Call returned status code 2:
  current_measurement_init(&current_measurement_settings)
in function main
at projects/power_distribution/src/main.c:153
*********************************************************************
```
Then it calls `__builtin_trap()` to exit immediately (via a hard fault on stm32, or something else like an illegal instruction on x86). The `***` is to draw attention to the debug output and away from the hard fault/illegal instruction output.

Also added a `wait()` in the PD event loop because it wasn't there before and that's bad for power consumption/my laptop fans.

### DIFF
--- a/libraries/libcore/inc/bug.h
+++ b/libraries/libcore/inc/bug.h
@@ -4,7 +4,7 @@
 // returns a non-OK status code.
 //
 // Usage:
-// BUG(my_critical_function_that_can_never_fail());
+// BUG(initialize_my_critical_module());
 //
 // BUG is meant to wrap operations where the only sensible thing to do on failure is to exit with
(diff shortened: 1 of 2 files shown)


jessm (approved):
LGTM! thanks for the good comment

PR #334, opened 2021-04-05
Soft 999 eddy kim fw 102

Eddy-M-K:
Looking for feedback on homework for FW102.


projects/soft_999_eddy_kim_fw_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdio.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_A_WAIT_TIME 500
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
copperium, line 11: Since this is a private function, it should be marked `static`, which makes it local to the file---otherwise, if another file tried to define a `prv_timer_callback`, you might get linker errors.

projects/soft_999_eddy_kim_fw_102/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  soft_timer_start_millis(COUNTER_A_WAIT_TIME, prv_timer_callback, counters, NULL);
+}
+
+int main(void) {
+  typedef struct Counters {
+    uint8_t counter_a;
+    uint8_t counter_b;
+  } Counters;
```
copperium, line 30: TIL you can typedef structs inside functions. In any case, you should do this on the top level so you don't have to do it twice (it's not considered a "global variable").

projects/soft_999_eddy_kim_fw_102/src/main.c
```diff
@@ -0,0 +1,44 @@
+#include <stdint.h>
+#include <stdio.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_A_WAIT_TIME 500
```
copperium, line 9: Awesome that you factored this out into a constant! Our convention is to suffix constants like this with their unit - in this case it would be `_MS`.

projects/debug_test/src/main.c
```diff
@@ -0,0 +1,69 @@
+#include <stdint.h>
```
copperium, line 1: These `debug_test` and `hello_world` projects aren't part of the homework, please remove.

copperium (changes requested):
Nice job Eddy! FYI, you don't have to make a new PR if you want to update it; just push to the branch and the PR will update automatically.

projects/soft_999_eddy_kim_fw_102/src/main.c
Eddy-M-K: Oh ok haha - I didn't realize that it wasn't considered a global variable.

copperium:
Looks good! Feel free to move on to FW 103. Closing this since homework is complete.

PR #336, opened 2021-04-08
Soft 450 baby driver integration test

farazkh80:
The tests did not raise any errors. I tested with both `make mpxe` and `make fastmpxe`. 

The only issue was that the `self.babydriver.sim.get_gpio()` threw a "KeyError (1,0)" which I could not figured out why. I also inlcuded this issue as a comment in the test for get_gpio.


projects/baby_driver/scripts/repl_setup.py
```diff
@@ -15,7 +15,10 @@
     init_notifier_gpio_it
 from can_send import can_send_raw, load_dbc, can_send
 
-
+def call_gpio_it():
+    """Initialize gpio intterupt"""
+    init_notifier_gpio_it()
```
farazkh80, line 20: I added this function so that  gpio_it can be initialized without going through parsing.

mpxe/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_gpio_get(self):
+        time.sleep(0.5)
+        gpio_get.gpio_get('A', 6)
+
+    def test_gpio_set(self):
+        time.sleep(0.5)
+        gpio_set.gpio_set('A', 5, True)
+        
+        #time.sleep(0.5)
+        #self.babydriver.sim.get_gpio(proj=self.babydriver, port=0, pin=5)
```
farazkh80, line 38: This was the error I mentioned in the stand up and the meeting.

I am not sure if the way I call it is correct.

mpxe/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,49 @@
+import sys
+sys.path.append('projects/baby_driver/scripts')
```
farazkh80, line 2: I am currently looking on the web if there is any other way of including this path but for now this works even though it s very hacky.

mpxe/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,49 @@
...
+        self.babydriver.sim.get_gpio(proj=self.babydriver, port='A', pin=5)
+        time.sleep(1)
+        gpio_set.gpio_set('A', 5, True)
+        time.sleep(1)
+        self.babydriver.sim.get_gpio(proj=self.babydriver, port='A', pin=5)
+    def test_gpio_interrupts(self):
+        time.sleep(1)
+        gpio_interrupts.register_gpio_interrupt(port='A', pin=3)
+        time.sleep(1)
+        gpio_interrupts.unregister_gpio_interrupt(port='A', pin=3)
```
copperium, line 40: To test gpio interrupts even better, you should simulate an interrupt:
* Pass a callback to `register_gpio_interrupt`, like `callback=interrupt_callback`. It can be defined inside `test_gpio_interrupts` and it has to take a single parameter `info`, but it doesn't have to use it. It can just set some property on `self` to True so you know that it's called (set that property to False out of the function so it's defined).
* After registering the interrupt, simulate the interrupt occurring. By default the interrupts are set on both rising and falling edges, so use `self.babydriver.sim.set_gpio` to set the pin to low and then high so you know the interrupt should occur.
* Then assert that the interrupt callback was called by asserting that the property set in the callback function is True.
This way you test all the use cases of gpio interrupts.

copperium (changes requested):
Nice job, this will be really valuable in ensuring that babydriver continues to function correctly! A few things to fix up.

If pylint is passing for you locally but failing in CI, do `pip3 install pylint` (or if that fails, `pip install pylint`) to upgrade your pylint version, then try again. And you can ignore pylint errors that you need to ignore (e.g. the one complaining about the `sys.path.append` call in the middle of the imports) by doing `# pylint-disable=<comma separated list of warnings to disable>` (without `<>`).

copperium:
FYI: MPXE got [renamed to MU](https://github.com/uw-midsun/firmware_xiv/pull/341) and there was a [bunch of refactoring done recently](https://github.com/uw-midsun/firmware_xiv/pull/344) so you may have some merge conflicts that need to get resolved, let me know if you need help.

copperium (changes requested):
I've resolved all the comments that you've implemented so the unresolved ones are ones you still need to address. In the meantime, a couple suggestions for fixing CI.

mu/sims/mci.py
```diff
@@ -9,7 +9,7 @@ def __init__(self, pm, proj_name):
 
     def handle_store(self, store, key):
         # Upon precharge enable (PA9) going high, set PB0 and PA10 to mock precharge
-        if not self.precharge_enable and self.get_gpio('a', 9):
-            self.precharge_enable = self.get_gpio('a', 9)
+        if not self.precharge_enable and self.'a', 9):
+            self.precharge_enable = self.'a', 9)
```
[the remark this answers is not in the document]
farazkh80: Yeah I have no idea what happened, will fix!
I ran pyformat and it apparently formatted a lot of files.

mpxe/integration_tests/test_babydriver.py
farazkh80: I have a few questions about this one:

1. Should I define an inner function inside `test_gpio_interrupts`  called `interrupt_callback`  that takes a parameter `info` which is set to True in the body of the function, and then pass this function to `register_gpio_interrupts` as `callback=interrupt_callback` ? 
2.  For the paramater `info`, Do I declare and set it to False in `test_gpio_interrupts` and then call it in the inner function `interrupt_callback` and set it to True in its body, and finally assert that `info is True`? so that I can test if the callback actually happened.

farazkh80: Please review the way I implemented it. 

The assert still fails but I am not quite sure if the way I coded is similar to the way you had in mind.

copperium: The way you've implemented it is generally correct just with a few errors, I'll review it in a proper review in a moment.

mu/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,88 @@
...
+        gpio_set.gpio_set('A', 5, True)
+        time.sleep(1)
+        assert self.babydriver.get_gpio(port='A', pin=5) is True
+
+    def test_gpio_interrupts(self): #IN PROGRESS
+        # Defining a Callback function
+        # pylint: disable=attribute-defined-outside-init
+        isCalled = False
+        def interrupt_callback(info):
+            self.isCalled = True
```
copperium, line 48: `isCalled` and `self.isCalled` are referring to two different variables: `isCalled` is a local variable in `test_gpio_interrupts`, whereas `self.isCalled` is an attribute on `self`. So this is one reason the assert in this test is failing.

To fix it, one solution is to use `self.isCalled` everywhere, but I think a nicer solution with fewer attributes defined on `self` (and which pylint won't complain about) would be to use `isCalled` in both places and put a [`nonlocal`](https://www.w3schools.com/python/ref_keyword_nonlocal.asp) declaration in `interrupt_callback`:
```py
def interrupt_callback(info):
    nonlocal isCalled
    isCalled = True
```
`nonlocal` is a Python 3 keyword which acts like `global`, but tells the interpreter that the variable referred to is in the next highest enclosing scope (`test_gpio_interrupts` in this case). Then you can remove the pylint comment here!

mu/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,88 @@
...
+        # Registring the interrupt
+        time.sleep(1)
+        gpio_interrupts.register_gpio_interrupt(port='A',
+                                                pin=5,
+                                                callback =interrupt_callback(info="info"))
```
copperium, line 53: The other reason this isn't working is because `interrupt_callback(info="info")` is actually *calling* `interrupt_callback` (with the argument `"info"`) rather than passing the function itself as a callback. Just pass the function name: `callback=interrupt_callback`.

(Also another `make pyformat` pass couldn't hurt to fix any spacing issues)

mu/integration_tests/test_babydriver.py
```diff
@@ -0,0 +1,88 @@
...
+        # Simulating
+        time.sleep(1)
+        self.babydriver.set_gpio(port='A', pin=5, state=True)
+        time.sleep(1)
+        self.babydriver.set_gpio(port='A', pin=5, state=False)
+        # Asserting
+        time.sleep(1)
+        assert isCalled is True
+        # Unregistring the interrupt
+        time.sleep(1)
```
copperium, line 63: Just to speed it up a notch, I don't think you need this particular `time.sleep` call since the assert shouldn't have done anything that needs to be waited for, so you can probably remove it. (But if things start failing, feel free to put it back.)

test.txt
```diff
@@ -0,0 +1 @@
+called
```
copperium, line 1: I think this got accidentally committed, please remove

copperium (changes requested):

test.txt
farazkh80: I think I deleted it and it does not appear on the branch repository. but still shows as one of the changed files in pr.

farazkh80:
I have implemented and fixed the final changes requested.

Just a few things to point out here:

1.  When starting vagrant and running `mu` or `fastmu` for `TEST=babydriver PLATFORM=x86` for the first time since vagrant is up, I get the error 
```bash
OS Error unknown device
```
However, it gets fixed after running `make babydriver` which I think makes the `.o` files needed. right?

2. Also, occasionally I get the Error
```bash
ERROR: test_adc_read (integration_tests.test_babydriver.TestBabyDriver)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vagrant/shared/firmware_xiv/mu/integration_tests/test_babydriver.py", line 26, in test_adc_read
    adc_read.adc_read(port='A', pin=6, raw=True)
  File "projects/baby_driver/scripts/adc_read.py", line 58, in adc_read
    status_msg = can_util.next_message(babydriver_id=BabydriverMessageId.STATUS)
  File "projects/baby_driver/scripts/can_util.py", line 222, in next_message
    msg.data[0] if msg.data else "empty message",
ValueError: next_message expected babydriver ID (0,) but got 5
```
for tests `test_adc_read` and `test_spi_exchange` sometimes when two consecutive `fastmu TEST=babydriver` are ran.
However, I was able to temporarily fix  the one for `test_adc_read` by changing the state passed from `state=False` to `state=True` and vice versa.

As I mentioned before, these errors don't usually occur but in some instances it shows up which makes me wonder if there is something that I implementing or understanding wrongly.

3- I have deleted the test.txt file which I initially created for debugging the callback function when I thought it is not working. However, despite being deleted from branch repo and code, It still shows as a changed file in pr request.

copperium:
1. That's probably because the virtual CAN network isn't up, run `make socketcan` if that happens and it should come up. It should be run automatically by `make mu` but I guess that's broken.
2. That's a strange error, probably nothing to do with your code. Integration tests are kind of flaky by nature, so it's probably just random, nothing you have to worry about.
3. Are you sure you `git add`'d the deleted file? I.e. try `git add --all` and then commit and push and see if that works.

farazkh80:
I tried `git add --all` but there was nothing to add. I was able to remove it by deleting the file directly in the pull request.

The file doesn't show up in the pull request nor the repository.

copperium (approved):

PR #337, opened 2021-04-09
[SOFT-460] Fix logging in MPXE

copperium:
We previously wouldn't initialize `store` in the MPXE `LOG` implementation. This caused segfaults when using babydriver with MPXE, because it calls `LOG_DEBUG` first thing before any initialization. Fix by calling `store_config()` in `LOG`.


jessm (approved):
for cleanliness of the macros sake, wanna put this in `log_export`? up to you, I'll approve either way

also while you're in here could also remove the fflush comment but not necessary, I do love me some one line PRs

copperium:
yeah I agree it fits better in `log_export`

PR #338, opened 2021-04-10
[SOFT-345] Write on main thread, and minor refactoring

jessm:
You thought regular spaghetti code was bad, try spaghetti concurrency

- MCI handler uses timers instead of blocking the poll thread
- Project gets a ref to PM instead of needing it passed as an arg sometimes
- All writes go through main thread via a queue and spaghetti concurrency


mitchellostler (commented):
LGTM

PR #339, opened 2021-04-11
FW101 and FW102 Onboarding Homework

michael1ding:


projects/firmware_homework102/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+void delay(unsigned milliseconds) {
+  clock_t pause;
+  clock_t start;
+
+  pause = milliseconds * (CLOCKS_PER_SEC / 1000);
+  start = clock();
+  while ((clock() - start) < pause) {
+    continue;
+  }
+}
```
copperium, line 31: This is quite creative, I'm glad you've got some C standard library knowledge. We do already have a few delay functions in our library (implemented using soft timers) though: `delay_us`, `delay_ms`, and `delay_s` from delay.h (in libraries/ms-common/inc).

A couple other points:
* we're using C11, you don't have to declare all variables at the start of the function (if that's what you were doing)
* please use types from <stdint.h> like `uint8_t`, `uint16_t` etc, normal types like `unsigned int` might have different sizes on x86 and stm32.

projects/firmware_homework102/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+
+void increment(SoftTimerId timer_id, void *context) {
```
copperium, line 33: This is a private function, so our [coding standards](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards) say that it should be `static` and prefixed with `prv_`. (The `static` makes the function file-local, so that if another file declares a `prv_increment` function, we don't have a linking issue.)

projects/firmware_homework102/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+  struct Counters *temp = context;
+  temp->counter_a += 1;
```
copperium, line 35: Nitpick: you can just use `++` instead of `+= 1`

projects/firmware_homework102/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+  LOG_DEBUG("Counter A: %d\n", temp->counter_a);
+  delay(0.5);
```
copperium, line 37: This is delaying for 0ms (`(unsigned)0.5`), not 500ms, eh? In any case, it's fine for this homework, but in real non-unit test firmware using `delay` isn't advised because it blocks execution. Usually we have several tasks we'd like each board to do, and using `delay` delays all of them.

projects/firmware_homework102/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+  soft_timer_start_millis(500, increment, temp, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  struct Counters *temp = counter_init();
+
+  soft_timer_start_millis(500, increment, temp, NULL);
```
copperium, line 53: It's good practice to factor out the 500 from here and the `increment` function into a nice `#define`'d macro so that the delay is only specified in one place. Make sure the macro has a descriptive name, and suffix the macro name with the unit (in this case `_MS`) so we know what it represents.

copperium (changes requested):
Great job Michael, I can see you've worked with C before! Just a few things to fix up before moving on to FW 103.

copperium:
Looks good! Closing since homework is complete, feel free to move on to FW 103.

PR #340, opened 2021-04-12
[SOFT-233] Adjust timings for MPXE

jessm:
This adjusts some smoketest timings and driver timings to better accomodate mpxe by making integration tests run faster.

Smoketest timings have super minimal effects because users can always just adjust them manually when using them.

The change in ads1015 is because it was previously a 625 us loop, which caused perf issues with the latest soft_timer library changes and mpxe locking strategy. This is only on the x86 version, where there shouldn't be any necessity to be reading values at that frequency.

### DIFF
--- a/libraries/ms-helper/src/x86/ads1015.c
+++ b/libraries/ms-helper/src/x86/ads1015.c
@@ -23,6 +23,9 @@
 #define ADS1015_CHANNEL_ARBITRARY_READING 0
 
 #ifdef MPXE
+#undef ADS1015_CHANNEL_UPDATE_PERIOD_US
+#define ADS1015_CHANNEL_UPDATE_PERIODS_US 10000
+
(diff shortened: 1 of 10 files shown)


libraries/ms-helper/src/x86/ads1015.c
```diff
@@ -23,6 +23,9 @@
 #define ADS1015_CHANNEL_ARBITRARY_READING 0
 
 #ifdef MPXE
+#undef ADS1015_CHANNEL_UPDATE_PERIOD_US
```
AveryChiu64, line 26: Is this macro defined elsewhere? Why do we need this undef?

AveryChiu64 (approved):

libraries/ms-helper/src/x86/ads1015.c
jessm: yes, it's defined 4 lines higher for normal stm32f0xx and x86 unit test use

PR #342, opened 2021-04-13
Soft 999 eddy kim fw 103

Eddy-M-K:
Homework for FW103:

I think I definitely got the code wrong this time because I didn't really understand which functions I should be using. Also, I do think the callback function I wrote is wrong so pointers on what I should be looking to change are appreciated.

Also, I couldn't find everything for the schematic scavenger hunt:

image
image


projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  GpioAddress *A6 = context;
+  uint16_t A6_data = 0;
+  adc_read_raw_pin(*A6, &A6_data);
+  LOG_DEBUG("%d\n", A6_data);
+  return;
+}
+
+int main(void) {
+  gpio_init();
+  interrupt_init();
```
copperium, line 21: You also need to initialize gpio interrupts with `gpio_it_init()`

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+  // Unsure if these are defined properly
+  GpioAddress B2 = { .port = GPIO_PORT_B, .pin = 2 };
+  GpioAddress A6 = { .port = GPIO_PORT_A, .pin = 6 };
```
copperium, line 25: These are defined properly :) Though "B2" and "A6" just refer to the port/pin combination, not the variable name - in fact you should name them something that reflects their purpose, like e.g. `button_pin`, instead of just what the pin is.

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+  // Unsure if these are defined properly as well
+  GpioSettings B2_settings = { GPIO_DIR_IN, GPIO_STATE_LOW, GPIO_RES_NONE, GPIO_ALTFN_ANALOG };
+
+  GpioSettings A6_settings = { GPIO_DIR_IN, GPIO_STATE_LOW, GPIO_RES_NONE, GPIO_ALTFN_ANALOG };
```
copperium, line 30: Two points about this:
* This is correct for A6, `GPIO_ALTFN_ANALOG` is the correct alt-function (something different for the MCU to do with the pin) for reading from ADC. For B2, you don't need to read from ADC for it, so use `GPIO_ALTFN_NONE` instead.
* On a stylistic note, we usually prefer GNU-style struct initialization syntax like this:
```c
GpioSettings settings = {
  .direction = GPIO_DIR_IN,
  .state = GPIO_STATE_LOW,
  .resistor = GPIO_RES_NONE,
  .alt_function = GPIO_ALTFN_ANALOG,
};
```
That way it's easy to verify that the fields match up with the values since it's not dependent on ordering. (Also, the names `B2_settings` and `A6_settings` don't *really* comply with the coding style since they start with capitals.)

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(A6, true);
+
+  // Unsure if it's necessary to set channel pin for B2 as well
+  adc_set_channel_pin(B2, true);
```
copperium, line 37: You don't need to do this for B2, all this does is enable the ADC for reading on the pin you pass in. You don't want to read from B2 with the ADC, so you don't need this.

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+  InterruptSettings settings = { INTERRUPT_TYPE_INTERRUPT, INTERRUPT_PRIORITY_NORMAL };
+  InterruptType edge = { INTERRUPT_EDGE_FALLING };
```
copperium, line 40: Does the `{}` syntax with non-struct values actually work? I'm surprised. In any case you can just use the value `INTERRUPT_EDGE_FALLING`, and the type should be `InterruptEdge`, not `InterruptType`.

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+static void callback(const GpioAddress *address, void *context) {
+  // Is it right to pass the GpioAddress A6 by *context? I'm unsure what *address is supposed to
+  // return
+  GpioAddress *A6 = context;
```
copperium, line 12: Yes, this is correct. `*address` is the address of the pin on which the interrupt was triggered (in this case B2, not A6).

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  uint16_t A6_data = 0;
+  adc_read_raw_pin(*A6, &A6_data);
```
copperium, line 14: The homework does say to use a _converted_ read, so use `adc_read_converted_pin`. The difference is that a converted read converts the value to millivolts, whereas a raw read is just whatever the ADC spits out.

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  LOG_DEBUG("%d\n", A6_data);
+  return;
```
copperium, line 16: Extraneous return statement, you don't need one at the end if the function returns void

projects/soft_999_eddy_kim_fw_103/src/main.c
```diff
@@ -0,0 +1,49 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+static void callback(const GpioAddress *address, void *context) {
```
copperium, line 9: This is a private function, so by our conventions it should be prefixed with `prv_`.

copperium (changes requested):
Nice, a few things to fix

projects/soft_999_eddy_kim_fw_103/src/main.c
Eddy-M-K: Whoops, I should pay more attention to the coding standards...

projects/soft_999_eddy_kim_fw_103/src/main.c
Eddy-M-K: The curly braces were my mistake haha; I can't believe I didn't catch this

copperium:
Looks good! Feel free to move on to FW 104 and/or ask for a task. Closing this since homework is complete.

PR #346, opened 2021-04-24
[SOFT-233] Set astroid version

jessm:


copperium (approved):
Note for future people wondering why astroid is pinned: astroid 2.5.4+ broke pylint 2.7.2 because somebody messed up semantic versioning and made a breaking change in a minor release

PR #349, opened 2021-04-25
Implemented Chip Id

YiJie-Zhu:


libraries/ms-common/inc/chip_id.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
```
copperium, line 2: Since users of the module will look at the header for how to use it, you should add a few comments here to document what the module is. In particular, mention that it's for accessing the hardware chip ID (point them to the datasheet like you did in the stm32 impl file) and mention that it's constant for every controller board. You could also mention how it's emulated in x86 if you want.

copperium (changes requested):
Nice job, I love the thorough commenting in the implementation files!

libraries/ms-common/test/test_chip_id.c
```diff
@@ -0,0 +1,32 @@
...
+#include "delay.h"
+#include "interrupt.h"
+#include "log.h"
+#include "misc.h"
+#include "soft_timer.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+ChipId chip_id_1 = { 0 };
+ChipId chip_id_2 = { 0 };
```
copperium, line 13: Could you change these to local variables? It's generally good practice to minimize global variables as much as possible.

libraries/ms-common/test/test_chip_id.c
```diff
@@ -0,0 +1,32 @@
...
+void teardown_test(void) {}
+
+// Verifies chip ID is consistant and prints chip ID for debugging
+void test_chip_id(void) {
+  const uint32_t delay_second_ms = 10;
+  chip_id_1 = chip_id_get();
+  delay_ms(delay_second_ms);
+  chip_id_2 = chip_id_get();
+  for (int i = 0; i < 2; i++) {
+    LOG_DEBUG("%" PRIu32 "\n", chip_id_1.id[i]);
```
copperium, line 29: A couple requests just to help with determining what a board's chip ID is:
* the upper bound on the loop should be 3 to loop through all three elements
* could you use hex instead of decimal when printing? This just requires you to use `PRIx32` instead of `PRIu32`. Also put an `0x` in front of the `%` so it prints out with an `0x` in front please.

Also, nitpick: you should use uint8_t or another fixed width type instead of `int`, even though it doesn't really matter.

libraries/ms-common/src/stm32f0xx/chip_id.c
```diff
@@ -0,0 +1,23 @@
+#include <stdint.h>
+
+#include "chip_id.h"
```
copperium, line 3: Nitpick: the import order should have the header file you're implementing first, so these includes should be swapped

libraries/ms-common/test/test_chip_id.c
```diff
@@ -0,0 +1,32 @@
+#include "chip_id.h"
+
+#include <inttypes.h>
+#include "delay.h"
```
copperium, line 4: Nitpick: there should be a blank line here

copperium (changes requested):
I ran the test on hardware and it worked great :) Just a couple of nitpicks to make it easier to use the test, then we can get this merged.

copperium (approved):

PR #350, opened 2021-04-26
[SOFT-466] Experimental stack analyzer

copperium:
Chromium OS's Embedded Controller (EC) team has a [super cool open source stack size analyzer tool](https://www.chromium.org/chromium-os/ec-development/stack-size-analyzer) which analyzes disassembly to find the maximum possible stack size. This copies and adapts the EC stack analyzer for our firmware. (It's BSD-licenced, so we're fine.) There's a good summary of the stack analyzer's capabilities in platform/stm32f0xx/scripts/stack_analyzer/README.md.

To make the stack analyzer work without spewing `??` for every standard library symbol, this also adds bundled [`newlib`](https://sourceware.org/newlib) binaries with debug symbols enabled. (They took like 6 hours to build on my machine!) Newlib is the implementation of the C standard library we use since it's designed for embedded stuff; usually we just use the one bundled with gcc-arm-embedded, but it has debug symbols stripped.

Using the custom newlib binaries could also be generally useful for debugging with gdb, so we add an `STDLIB_DEBUG` make option which you can set to `true` to build with debug symbols enabled.

Most of the actual stack_analyzer.py script is a) super long and complex and b) written by google engineers and not me, so it shouldn't have to be reviewed - if you want to search for the parts I wrote, I marked most of them with `MidSun extension` or something like that.


jessm (approved):
Super cool! Appreciate all the work that looks like went into getting this working

PR #358, opened 2021-05-01
[Soft 447] Bootloader can id handling

danielzyy:


danielzyy:
Test bootloader can gets stuck on [0] libraries/ms-common/src/x86/can_hw.c:75: CAN HW RX thread started.

libraries/ms-common/src/can_fsm.c
```diff
@@ -53,6 +55,12 @@ static void prv_handle_rx(Fsm *fsm, const Event *e, void *context) {
     return;
   }
 
+  // Process bootloader messages
+  if(rx_msg.source_id == SYSTEM_CAN_DEVICE_BOOTLOADER) {
+    result = bootloader_can_receive(&rx_msg);
+  }
+  return;
```
copperium, line 62: Not sure if this is the cause of the test error or not but the `return` should go inside the if statement, otherwise we're ignoring all normal CAN messages.

danielzyy:
Made the change, but still did not fix the issue of the test being stuck at: [0] libraries/ms-common/src/x86/can_hw.c:75: CAN HW RX thread started

copperium:
Two things that fixed the getting stuck error for me (and replaced it with fun new errors):
* initialize all the things that CAN needs to be initialized (the event queue, interrupts, soft timers, and gpio) in `setup_test` - you should also have a `// Requires X, Y, and Z to be initialized.` comment in your header file to remind you/others that those need initialized
* set `loopback = true` in the CanSettings so that we receive the CAN messages that we send

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
...
+#include "can_msg_defs.h"
+#include "can_transmit.h"
+#include "status.h"
+
+// Initialize the module using CanStorage and CanSettings, without needing to specify device_id
+// device_id will be overwritten to be SYSTEM_CAN_DEVICE_BOOTLOADER
+StatusCode bootloader_can_init(CanStorage *storage, CanSettings *settings);
+
+// Transmits a CAN message with data of length len, and sets the message type = CAN_MSG_TYPE_ACK
+// if is_start_message is true, otherwise, type = CAN_MSG_TYPE_DATA
```
copperium, line 21: Your doc comments in this header seem to be describing implementation details of the functions, rather than their purpose. The user of this module doesn't care exactly how the module formats a CAN ID - in fact, the entire purpose of the module is so that they don't *have* to care. If they wanted to know how the module indicates whether a message is a start message, they could read the implementation file or the design document.

So instead of talking about what `bootloader_can_transmit` does with the message type, say that it transmits a bootloader datagram message with the given data, from the given board id, and that it's a start message if `is_start_message` is true and a non-start message otherwise.

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+// This module allows the sending and receiving of CAN messages in the bootloader's CAN ID format.
```
copperium, line 3: You should link the bootloader design doc Confluence page here (short link: https://uwmidsun.atlassian.net/l/c/76h1GfbH).

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
...
+#include "status.h"
+
+// Initialize the module using CanStorage and CanSettings, without needing to specify device_id
+// device_id will be overwritten to be SYSTEM_CAN_DEVICE_BOOTLOADER
+StatusCode bootloader_can_init(CanStorage *storage, CanSettings *settings);
+
+// Transmits a CAN message with data of length len, and sets the message type = CAN_MSG_TYPE_ACK
+// if is_start_message is true, otherwise, type = CAN_MSG_TYPE_DATA
+StatusCode bootloader_can_transmit(uint8_t *data, size_t len, bool is_start_message,
+                                   uint16_t board_id);
```
copperium, line 23: IMO it's more intuitive if `board_id` is the first parameter.

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
...
+
+// Stores the callback and context for later use, if callback or context is NULL,
+// the function will return STATUS_CODE_INVALID_ARGS
```
copperium, line 26: This talks about implementation details instead of what the user of `bootloader_can_register_handler` needs to know - instead, you should say that the function registers a handler to be called when a bootloader CAN message is received, and that only one handler can be registered at a time.

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
...
+StatusCode bootloader_can_register_handler(CanRxHandlerCb callback, void *context);
+
+// Calls the registered handler function of the message if the msg_id is equal to the client ID, 0
+// Otherwise, the message is ignored
```
copperium, line 30: Make sure you mention here that this is mainly for internal use in the CAN library and doesn't have to be called by the user.

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,31 @@
...
+StatusCode bootloader_can_init(CanStorage *storage, CanSettings *settings);
+
+// Transmits a CAN message with data of length len, and sets the message type = CAN_MSG_TYPE_ACK
+// if is_start_message is true, otherwise, type = CAN_MSG_TYPE_DATA
+StatusCode bootloader_can_transmit(uint8_t *data, size_t len, bool is_start_message,
+                                   uint16_t board_id);
+
+// Stores the callback and context for later use, if callback or context is NULL,
+// the function will return STATUS_CODE_INVALID_ARGS
+StatusCode bootloader_can_register_handler(CanRxHandlerCb callback, void *context);
```
copperium, line 27: Hm, I can see that using a `CanRxHandlerCb` might be a bit confusing since the third parameter is always null. I think it actually makes more sense to typedef a new callback type like `CanRxHandlerCb` but without the `ack_reply` parameter and use that instead.

copperium (changes requested):
Nice job Daniel!

libraries/ms-common/src/bootloader_can.c
```diff
@@ -0,0 +1,51 @@
...
+}
+
+StatusCode bootloader_can_transmit(uint16_t board_id, uint8_t *data, size_t len,
+                                   bool is_start_message) {
+  CanMessage message = { .source_id = SYSTEM_CAN_DEVICE_BOOTLOADER,
+                         .msg_id = board_id,
+                         .type = CAN_MSG_TYPE_DATA,
+                         .dlc = len };
+  // copy the message data over to the message to be transmitted
+  memcpy(message.data_u8, data, 8);
```
copperium, line 29: We should only copy `len` bytes instead of 8 bytes - otherwise we'll segfault if you pass in a `data` array of <8 bytes. (Also, mention in the docs in the header that `len` has to be <=8, and maybe return `STATUS_CODE_INVALID_ARGS` early if `len > 8`.)

libraries/ms-common/inc/bootloader_can.h
```diff
@@ -0,0 +1,39 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "can.h"
+#include "can_msg_defs.h"
+#include "can_transmit.h"
+#include "status.h"
+
+// Called whenever a bootloader message is received with client controller board ID 0
+typedef StatusCode (*BootloaderCallback)(const CanMessage *msg, void *context);
```
copperium, line 20: Nitpick: could you call this `BootloaderCanCallback` instead? `BootloaderCallback` is a bit non-specific, we'll likely define more callbacks in the bootloader proper.

libraries/ms-common/test/test_bootloader_can.c
```diff
@@ -0,0 +1,112 @@
...
+  interrupt_init();
+  soft_timer_init();
+
+  TEST_ASSERT_OK(bootloader_can_register_handler(prv_bootloader_rx_callback, NULL));
+}
+
+void teardown_test(void) {}
+
+// Test that a bootloader CAN message with client board ID can be succesfully sent
+// and the appropreate callback called when received
```
copperium, line 72: Typo: `appropreate` -> `appropriate` (also repeated below)

copperium (changes requested):
I resolved all the comments you've completed, there are a couple doc changes from before still pending. The only really blocking comment from this round is the one about memcpying only `len` bytes, after that we should be good to merge!

copperium (approved):

PR #359, opened 2021-05-02
Firmware 103 initial commit

Frank-Yan:
Firmware 103 Part 2 Coding Challenge (Will send scavenger hunt later)

### DIFF
--- a/projects/firmware_homework/src/main.c
+++ b/projects/firmware_homework/src/main.c
@@ -9,20 +9,25 @@
 // Interrupt callback function
 static void prv_gpio_callback(const GpioAddress *address, void *context) {
   // Context should be the address of A6 as we want to read from this pin
-  GpioAddress *adc_channel = context;
+  GpioAddress *adc_reading = context;
   uint16_t adc_data = 0;
   // Read the values from the ADC channel and output the reading
-  adc_read_converted_pin(*adc_channel, &adc_data);
-  LOG_DEBUG("ADC Reading: %i\n", adc_data);
+  // Check the data to see if it is valid first
+  if (adc_read_converted_pin(*adc_reading, &adc_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("ADC Reading: %i\n", adc_data);
+  } else {
+    LOG_DEBUG("Invalid Reading: %i\n");
+    // I'll just put invalid reading for anything else than STATUS CODE OK
+  }
 }
 
 int main(void) {
-  /*Initializations*/
+  // Initializations
   interrupt_init();  // Enable interrupts
   gpio_init();       // Set up all GPIO pins
   gpio_it_init();    // Set GPIO pins for interrupts
 
-  /*Pin Defintions*/
+  // Pin Defintions
   // GPIO pin (B2) that is connected to the button and "triggers" the interrupt
   GpioAddress button_pin = {
     .port = GPIO_PORT_B,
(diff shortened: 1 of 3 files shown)


projects/firmware_homework/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+#include "log.h"
+#include "wait.h"
+
+// Interrupt callback function
+static void prv_gpio_callback(const GpioAddress *address, void *context) {
+  // Context should be the address of A6 as we want to read from this pin
+  GpioAddress *adc_channel = context;
+  uint16_t adc_data = 0;
+  // Read the values from the ADC channel and output the reading
+  adc_read_converted_pin(*adc_channel, &adc_data);
```
AveryChiu64, line 15: Though this is okay. What would be even better is if we add a check to see what the `StatusCode` is which is returned from this function to make sure that `adc_data` is valid.

AveryChiu64:
Great work! Just a few small changes to make.

projects/firmware_homework/src/main.c
Frank-Yan: Is there a way to confirm that adc_data is valid? I'm sure there is a module but I didn't see mentions of it in the tutorial.

copperium: Lots of our functions return a `StatusCode`, which is a return code that tells you whether everything went alright in the function. It's our alternative to exceptions as found in other languages. Generally if a function returns `STATUS_CODE_OK`, everything went alright, and if it returns another code then you should do any applicable cleanup and pass the status code up the stack by returning it. See status.h (in libraries/libcore) for the list of status codes, and check out the `status_ok_or_return` macro.

projects/firmware_homework/src/main.c
```diff
@@ -0,0 +1,79 @@
...
+static void prv_gpio_callback(const GpioAddress *address, void *context) {
+  // Context should be the address of A6 as we want to read from this pin
+  GpioAddress *adc_reading = context;
+  uint16_t adc_data = 0;
+  // Read the values from the ADC channel and output the reading
+  // Check the data to see if it is valid first
+  if (adc_read_converted_pin(*adc_reading, &adc_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("ADC Reading: %i\n", adc_data);
+  } else {
+    LOG_DEBUG("Invalid Reading: %i\n");
```
copperium, line 19: This is failing to compile in CI because you don't pass an argument to fill the `%i` in the format string. Pass an argument or remove the `%i` to fix that and get CI to pass.

copperium (changes requested):
Just one other thing before we can mark the homework as complete.

ryandancy:
Looks like you removed the `%i` from the first log as well, that's *also* causing the build to fail because you *do* pass an argument to fill the `%i` (`adc_data`).

In general, you can see that the build is failing (i.e. your code fails to compile or a unit test fails) because it says 'All checks have failed' at the bottom of the PR, and you can see why by clicking on 'Details'. (Sometimes it'll fail on an unrelated unit test; if that happens just click 'Rerun all jobs'.) You can also check that your code compiles with `make build PLATFORM=x86 PROJECT=firmware_homework`.

As soon as you get CI to pass, feel free to move onto FW 104 and/or ask me/Avery for a task.

ryandancy:
Looks good! Closing since homework is complete, feel free to move on to FW 104 and/or ask me/Avery for a task.

PR #361, opened 2021-05-03
FW103 Homework

michael1ding:
I was a little confused when doing some of this homework, and I don't really have a board to test if this code actually works, so I was hoping for some feedback!


projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+  soft_timer_init();
+  gpio_init();
+  gpio_it_init();
+
+  gpio_init_pin(&reading_addr, &reading_settings);
+  gpio_init_pin(&button_read_addr, &button_read_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_get_channel(reading_addr, &reading_channel);
```
AveryChiu64, line 52: "We also have new, easier-to-use functions in adc.h like adc_set_channel_pin, adc_read_raw_pin and adc_read_converted_pin which take GpioAddresses and do all the channel conversion for you, so you don’t have to do the messy adc_get_channel call."

Though this is okay, and does follow the example as in FW103. I would suggest using these new functions to simplify the code. It will also make your life easier so you don't have to make these calls to adc_set_channel and adc_get_channel.

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+
+static void prv_button_callback(const GpioAddress *address, void *context) {
+  uint16_t reading = 0;
+  adc_read_converted(reading_channel, &reading);
+  LOG_DEBUG("reading in pin A6: %d\n", reading);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
```
AveryChiu64, line 43: Do we need soft timers here?

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+static GpioAddress button_read_addr = { .port = GPIO_PORT_B, .pin = 2 };
+
+static GpioSettings button_read_settings = {
+  .direction = GPIO_DIR_IN,
+  .state = GPIO_STATE_HIGH,
+  .resistor = GPIO_RES_PULLDOWN,
+  .alt_function = GPIO_ALTFN_NONE,
+};
+
+static InterruptSettings button_press = {
```
AveryChiu64, line 30: `button_press` might not be the best name for the interrupt settings

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+  gpio_init_pin(&reading_addr, &reading_settings);
+  gpio_init_pin(&button_read_addr, &button_read_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_get_channel(reading_addr, &reading_channel);
+  adc_set_channel(reading_channel, true);
+  adc_set_channel(ADC_CHANNEL_TEMP, true);
+
+  gpio_it_register_interrupt(&button_read_addr, &button_press, INTERRUPT_EDGE_RISING,
```
AveryChiu64, line 56: I think it should be `INTERRUPT_EDGE_FALLING`.
"The adc reading should trigger whenever the button is released, not initially when the button is pressed."

Since the button is active-high, this means that it is high when the button is pressed, the signal sent to the MCU will be high. This means when we release the button, the signal will go from HIGH to LOW (falling)

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+                             prv_button_callback, &reading_addr);
```
AveryChiu64, line 57: &reading_addr is passed in as the context, but it looks like you don't actually use it in the callback function?

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+
+  while (1) {
```
AveryChiu64, line 59: You should add `wait();` in this function (see the wait.h library).
This will basically allow the stm32 to enter low power mode and stop executing code as we wait for an interrupt to occur.

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
+#include <stdint.h>
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
```
AveryChiu64, line 8: I think you can remove some of these libraries such as soft_timer.h or delay.h since they are not being used.

AveryChiu64 (commented):
This is a good start, I have added comments to help you make the code even better.

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+
+static AdcChannel reading_channel = NUM_ADC_CHANNELS;
+
+static GpioAddress reading_addr = { .port = GPIO_PORT_A, .pin = 6 };
+
+static GpioSettings reading_settings = {
```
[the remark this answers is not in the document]
michael1ding: Just making sure, the button_pin_settings should be for the button right? I'm assuming the A6 pin reading analog to digital still needs an alt_function of GPIO_ALTFN_ANALOG, or has that changed?

michael1ding (commented):
Let me know if these changes are satisfactory

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  .resistor = GPIO_RES_NONE,
+  .alt_function = GPIO_ALTFN_NONE,  // This pin doesn't need to do anything special
+};
+
+static InterruptSettings button_rising = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+static void prv_button_callback(const GpioAddress *address, void *context) {
```
AveryChiu64, line 33: reading_addr is not defined in this function

projects/fw_hw103/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+
+  gpio_init_pin(&reading_addr, &reading_settings);
+  gpio_init_pin(&button_pin_addr, &button_pin_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(reading_addr, true);
+
+  gpio_it_register_interrupt(&button_pin_addr, &button_rising, INTERRUPT_EDGE_FALLING,
+                             prv_button_callback, &reading_addr);
```
AveryChiu64, line 52: Do you need to pass in read_addr as context here? This variable is static no?

copperium:
Looks good, feel free to move on to FW 104 or ask a lead for a task. Closing since homework is complete.

PR #362, opened 2021-05-04
[SOFT-233] Fix typo in CI script

copperium:
I typo'd `DEFINES` instead of `DEFINE` when setting up the GitHub Actions CI script. That was causing STM32 builds to build with log-level debug instead of warn as it should be. Thanks @hewittmcg for finding this.


hewittmcg (approved):
LGTM!

jessm:
i love soft 233

PR #363, opened 2021-05-04
[SOFT-167] Fix a bunch of solar validation bugs

copperium:
These were actually discovered and fixed in like September, for some reason I never put up a PR for them. So here they are.

Changes in order of appearance:
* Increase the max MCP3427 conversion time which we wait before checking for a conversion, because I think it was throwing too many errors.
* Actually assign the sample rate from the settings to the storage in MCP3427, we weren't doing that before and it was defaulting to 16 bits (slowest and most precise).
* Fix an endianness issue in the SPV1020 driver, it turns out the SPV1020 is big-endian while the STM32 is little-endian. Plus, relying on the endianness of the host system is a bad idea anyways.
* Increase the SPI baudrate from 60000 to the 6000000 value we use everywhere else in the SPI and SPV1020 smoketests, plus a few tests and solar itself.
* A couple MCP3427 address configs were misconfigured in solar, fix those.
* The I2C1 and I2C2 pins were swapped in the MCP3427 smoketest and in solar. Fixed that by using the standard macros from controller_board_pins.h instead.

### DIFF
--- a/projects/smoke_mcp3427/rules.mk
+++ b/projects/smoke_mcp3427/rules.mk
@@ -6,4 +6,4 @@
 # $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
 
 # Specify the libraries you want to include
-$(T)_DEPS := ms-common ms-drivers
+$(T)_DEPS := ms-helpers ms-common ms-drivers
--- a/projects/smoke_mcp3427/rules.mk
+++ b/projects/smoke_mcp3427/rules.mk
@@ -6,4 +6,4 @@
 # $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
 
 # Specify the libraries you want to include
-$(T)_DEPS := ms-helpers ms-common ms-drivers
+$(T)_DEPS := ms-helper ms-common ms-drivers
--- a/projects/solar/rules.mk
+++ b/projects/solar/rules.mk
@@ -6,7 +6,7 @@
 # $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
 
 # Specify the libraries you want to include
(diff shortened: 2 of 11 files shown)


projects/solar/rules.mk
```diff
@@ -6,7 +6,7 @@
 # $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
 
 # Specify the libraries you want to include
-$(T)_DEPS := ms-common ms-drivers
+$(T)_DEPS := ms-helpers ms-common ms-drivers
```
AveryChiu64, line 9: I think this should be `ms-helper`
Looks like GitHub actions is failing because of this as well.

```
T: test_adc_read_runner.c -> test_adc_read_runner.o
Building test test_adc_read for stm32f0xx
T: test_gpio_interrupts.c -> test_gpio_interrupts.o
Generating test_gpio_interrupts_runner.c
T: test_gpio_interrupts_runner.c -> test_gpio_interrupts_runner.o
Building test test_gpio_interrupts for stm32f0xx
make: *** No rule to make target 'ms-helpers', needed by 'solar'.  Stop.
Error: Process completed with exit code 2.
```

projects/smoke_mcp3427/rules.mk
```diff
@@ -6,4 +6,4 @@
 # $(T)_SRC: $(T)_DIR/src{/$(PLATFORM)}/*.{c,s}
 
 # Specify the libraries you want to include
-$(T)_DEPS := ms-common ms-drivers
+$(T)_DEPS := ms-helpers ms-common ms-drivers
```
AveryChiu64, line 9: Should be `ms-helper` not  `ms-helpers`

projects/solar/rules.mk
copperium: oh shoot whoops

AveryChiu64 (approved):

PR #367, opened 2021-05-06
[SOFT-353] MCI updates from validation

hewittmcg:
This finally cleans up the changes I made to get MCI working with the MCs from ~October - March.

### MCP2515 fixes

The main issue I was running into was with the MCP2515 that communicates with the motor controllers.  With the old implementation (using a `GenericCan` interface), if it received multiple CAN messages from different IDs it would lock up for some reason and stop processing messages.  The following changes solve this:

- Use the mcp2515 driver directly in `mci_broadcast.c` to filter for one message ID at a time
- Add `mcp2515_set_filter()` to `mcp2515.c` to allow for changing the filters without having to reinitialize everything. This more or less just copies the parts of `mcp2515_init()` that configure the filters
- Generally, whenever a message is received from a MC, call its associated callback and change the filter to the next message we're looking for
- Sometimes messages that we're not filtering for get through (???), so while doing the above we also check to make sure that the message received is the one we were looking for and don't change the filter if not
- Changed `mci_output.c` to send messages to the MCs using the MCP2515 driver instead of generic CAN

### Other changes
- Added functionality to receive and store status info from the MCs
- Added callbacks for motor temp and DSP temp measurements.  Storing and broadcasting these values should be done in SOFT-421 
- Added a CAN message, `SYSTEM_CAN_MESSAGE_MOTOR_STATUS`, that just broadcasts the error/limit flags from both MCs' status info messages
- Changed `wavesculptor.h` velocity definitions from 100/-100 rpm to 20000/-20000 rpm so we can control the motor using current
- Updated tests to work with above changes while keeping as much of the original logic as possible

### Notes
- The MCP2515 still behaves weirdly sometimes, so I want to take a closer look at the driver when there's time and we're allowed to go onsite again
- `wavesculptor.h` needs to be updated for the WS22, see SOFT-481
- All these changes were tested with the actual motor at the end of March and seemed to work OK, could turn it on/off and run it at a constant speed by mocking pedal messages


libraries/ms-helper/src/mcp2515.c
```diff
@@ -387,3 +382,60 @@ StatusCode mcp2515_tx(Mcp2515Storage *storage, uint32_t id, bool extended, uint6
 #endif
   return STATUS_CODE_OK;
 }
+
+StatusCode mcp2515_set_filter(Mcp2515Storage *storage, uint32_t *filters) {
+  // Primarily just copy-pasted from mcp2515_init()
```
ryandancy, line 387: Is it possible to call this function from `mcp2515_init` instead of copy-pasting?

projects/mci/test/test_mci_broadcast.c
```diff
@@ -56,19 +63,32 @@ typedef struct TestWaveSculptorBusMeasurement {
...
   [RIGHT_MOTOR_CONTROLLER * NUM_TEST_MCI_MESSAGES + TEST_MCI_VELOCITY_MESSAGE] =
-      MOTOR_CAN_RIGHT_VELOCITY_MEASUREMENT_FRAME_ID,
+      RIGHT_MOTOR_CONTROLLER_BASE_ADDR + WAVESCULPTOR_MEASUREMENT_ID_VELOCITY,
   [LEFT_MOTOR_CONTROLLER * NUM_TEST_MCI_MESSAGES + TEST_MCI_BUS_MEASUREMENT_MESSAGE] =
-      MOTOR_CAN_LEFT_BUS_MEASUREMENT_FRAME_ID,
+      LEFT_MOTOR_CONTROLLER_BASE_ADDR + WAVESCULPTOR_MEASUREMENT_ID_BUS,
   [RIGHT_MOTOR_CONTROLLER * NUM_TEST_MCI_MESSAGES + TEST_MCI_BUS_MEASUREMENT_MESSAGE] =
-      MOTOR_CAN_RIGHT_BUS_MEASUREMENT_FRAME_ID,
+      RIGHT_MOTOR_CONTROLLER_BASE_ADDR + WAVESCULPTOR_MEASUREMENT_ID_BUS,
+  [LEFT_MOTOR_CONTROLLER * NUM_TEST_MCI_MESSAGES + TEST_MCI_STATUS_MESSAGE] =
```
ryandancy, line 80: Idea: stick the `X_MOTOR_CONTROLLER * NUM_TEST_MCI_MESSAGES + TEST_MCI_X_MESSAGE` expression in a macro to make this array slightly less wall-of-text like.

projects/mci/test/test_mci_broadcast.c
```diff
@@ -195,29 +212,33 @@ void setup_test(void) {
...
 
 // lv - left velocity
 // rv - right velocity
 // lb - left bus measurement
 // rb - right bus measurement
 
-// Test 1: lb rv lv rb (check 2 output)
-void test_all_measurements_lb_rv_lv_rb() {
+// Test 1: General broadcast flow.
+void test_left_all_right_all() {
```
ryandancy, line 238: Nitpick: parameter list should be `(void)`

ryandancy (commented):
Looks good! Maybe @jessm could review the heavy MCP2515 stuff but I didn't see anything blocking, only a few cleanup suggestions.

projects/mci/src/mci_broadcast.c
```diff
@@ -24,7 +49,105 @@ static void prv_broadcast_bus_measurement(MotorControllerBroadcastStorage *stora
...
+      break;
+    }
+  }
+  // check if received message ID doesn't have a CB index
+  if (cb_index == NUM_MOTOR_CONTROLLER_BROADCAST_MEASUREMENTS) {
+    LOG_WARN("WARNING - no cb for message ID 0x%x or received message from wrong MC\n", (int)id);
+    // TODO(SOFT-139): error handling here
+    // This should NEVER happen, so it may make sense to throw an error
+    // However, have had this happen a few times, always with offset 9 (voltage rail measurement)
+    // not sure what the cause is, doesn't happen often
```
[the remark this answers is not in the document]
hewittmcg: Yeah, it's really bizarre, there's no reason why it should be happening since we don't (intentionally) filter for it at any point.  I assume it's driver weirdness, but IIRC I've only ever experienced it when testing with the MCP2515 currently in MCI so I guess there's an extremely small chance that it's a very, very weird hardware issue.

projects/mci/test/test_mci_broadcast.c
hewittmcg: Good idea, thanks, should be a lot cleaner like that

projects/mci/src/mci_broadcast.c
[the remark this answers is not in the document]
ryandancy: Idea: could it be a critical section issue? I.e. while we're changing the filter, could there be a brief intermediate state where the voltage rail message is allowed? If so, unlucky soft timer / interrupt timing could result in rxing a message in that time. If this is the case, it might be able to be fixed by guarding any critical sections (which shouldn't be interrupted so as not to be in an invalid state during an interrupt), but idk how MCP2515 works. That's the only way I can think it could be a firmware issue anyways.

projects/mci/inc/mci_broadcast.h
```diff
@@ -5,29 +5,60 @@
...
 
 typedef struct MotorControllerBroadcastStorage {
+  Mcp2515Storage *motor_can;
   uint8_t bus_rx_bitset;
   uint8_t velocity_rx_bitset;
+  uint8_t status_rx_bitset;
   MotorControllerMeasurements measurements;
   MotorCanDeviceId ids[NUM_MOTOR_CONTROLLERS];
+  // Callbacks exposed for unit testing
+  MotorControllerCallbackStorage cb_storage;
```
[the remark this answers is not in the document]
hewittmcg: Whoops, I think I may have been accessing them earlier on but realized I don't need to directly call them when testing.  Will make static as mentioned below, thanks

libraries/ms-helper/src/mcp2515.c
hewittmcg: Sort of -- however, `mcp2515_init` does configure a few other things while in config mode that would require the CAN bitrate to be passed into `mcp2515_set_filter`  as well, which would complicate the modules implementing it a bit.  

The simplest way I can think of to get around this is to define a `prv_configure_filters` function that handles everything in `mcp2515_set_filter` from directly after entering config mode to right before we leave it.  Then we can call that in both `mcp2515_init` and `mcp2515_set_filter`, getting rid of most of the copy-pasted code.

projects/mci/src/mci_broadcast.c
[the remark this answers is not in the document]
hewittmcg: Whoops missed this reply until now, sorry 😐 

I believe that that isn't possible since, once we put the MCP2515 into config mode, I _think_ it doesn't listen to the CAN bus until after we put it back into normal mode.  These changes happen at the start/end of the `mcp2515_set_filter` function, and all the register-level filter changes happen with it in config mode. The datasheet isn't 100% clear on this, but it explicitly specifies on page 59 when other nodes do listen to the bus (and it'd be really weird to allow RXing messages while configuring the masks/filters).

That said, I'll keep this in mind for the next time I get access to the hardware as something to double-check to confirm.

ryandancy (approved):

PR #368, opened 2021-05-06
Soft 354 i2c read firmware

abdullahs26:


libraries/codegen-tooling/rules.mk
```diff
@@ -8,5 +8,4 @@
 # The code in this library is autogenerated and pulled from GitHub
 # https://github.com/uw-midsun/codegen-tooling
 
-$(T)_DEPS := ms-common
-
+$(T)_DEPS := ms-common
```
copperium, line 11: Please put the newline back at the end of this file!

copperium (changes requested):
Just as a preliminary review, CI isn't running because of a merge conflict in `make/new_target.py`, and it seems you've somehow added 12.6 million spaces to a line in that file. Please revert that file to how it is on master (you can do it with `git pull && git checkout origin/master -- make/new_target.py`) and commit and push so CI runs.

projects/baby_driver/src/i2c_read.c
```diff
@@ -0,0 +1,118 @@
...
+      .scl = CONTROLLER_BOARD_ADDR_I2C2_SCL,
+    };
+    i2c_init(I2C_PORT_2, &i2c2_settings);
+  }
+
+  if (s_storage.is_reg != 0) {
+    i2c_read_reg(s_storage.port, s_storage.address, s_storage.reg, response, s_storage.rx_len);
+  } else {
+    i2c_read(s_storage.port, s_storage.address, response, s_storage.rx_len);
+  }
```
copperium, line 98: You should wrap all the `i2c_init`, `i2c_read`, `i2c_read_reg` calls in `status_ok_or_return` to fail fast if something fails.

projects/baby_driver/src/i2c_read.c
```diff
@@ -0,0 +1,118 @@
...
+static void prv_timer_data_callback(SoftTimerId timer_id, void *context) {
+  // This function is used to send CAN messages, it is a soft timer
+  // callback since there needs to be a delay between messages sent.
+
+  CAN_TRANSMIT_BABYDRIVER(BABYDRIVER_MESSAGE_I2C_READ_DATA, response[bytes_sent],
+                          response[bytes_sent + 1], response[bytes_sent + 2],
+                          response[bytes_sent + 3], response[bytes_sent + 4],
+                          response[bytes_sent + 5], response[bytes_sent + 6]);
+
+  bytes_sent += 7;
```
copperium, line 53: To get rid of the `bytes_sent` static/global variable, you can instead pass the bytes-sent counter through `context` by casting to/from the `uintptr_t` type, which essentially stuffs an integer value into a pointer. So you'd say `uintptr_t bytes_sent = (uintptr_t)context;` at the beginning and pass `(void *)bytes_sent` in the `soft_timer_start_millis` call. Pass `(void *)0` for the context in the initial `soft_timer_start_millis` in `prv_i2c_read_command_callback` and then you don't need the global!

projects/baby_driver/src/i2c_read.c
```diff
@@ -0,0 +1,118 @@
...
+  s_storage.port = data[1];
+  s_storage.address = data[2];
+  s_storage.rx_len = data[3];
+  // Stores whether it is a register read, Nonezero for register read, 0 for normal read
+  s_storage.is_reg = data[4];
+
+  // Only if it is a register read, store the register
+  if (s_storage.is_reg != 0) {
+    s_storage.reg = data[5];
+  }
```
copperium, line 75: You don't use the port, address, reg, and is_reg outside this function, so they don't have to be static. I think you can get rid of the storage and get away with storing only the response and rx_len as static variables, and the rest of these can be locals.

projects/baby_driver/src/i2c_read.c
```diff
@@ -0,0 +1,118 @@
...
+
+static I2CReadCommand s_storage = { 0 };
+static bool s_expecting_message = true;
+static uint32_t s_soft_timer_delay;
+static size_t bytes_sent = 0;
+
+// The size of the array is rounded up to 266 to account for accessing an array ouy of bounds in the
+// loop where the data is transmitted
+static uint8_t response[266] = { 0 };
+static uint8_t response_index;
```
copperium, line 22: I don't think you use `response_index`, please remove.

projects/baby_driver/src/i2c_read.c
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  // Expects a message next
+  s_expecting_message = true;
+
+  return STATUS_CODE_OK;
+}
+
+StatusCode i2c_read_init(uint32_t timeout_soft_timer) {
+  s_soft_timer_delay = timeout_soft_timer;
```
copperium, line 114: `timeout_soft_timer` and `soft_timer_delay` aren't terribly good names for these variables ("soft timer delay" is super generic). Instead, I'd use something like `tx_delay_ms` since it's the delay between CAN message TXes and it's in milliseconds.

projects/baby_driver/inc/i2c_read.h
```diff
@@ -0,0 +1,21 @@
...
+
+typedef struct I2CReadCommand {
+  I2CPort port;
+  I2CAddress address;
+  uint8_t rx_len;
+  uint8_t is_reg;
+  uint8_t reg;
+} I2CReadCommand;
+
+#define I2C_READ_DEFAULT_TIMEOUT_MS 5
```
copperium, line 18: Similarly, this should be renamed something like `I2C_READ_DEFAULT_TX_DELAY_MS`.

projects/baby_driver/src/main.c
```diff
@@ -25,6 +25,7 @@
 #include "soft_timer.h"
 #include "spi_exchange.h"
 #include "wait.h"
+#include "i2c_read.h"
```
copperium, line 28: Please put this include in alphabetical order with the rest of them!

projects/baby_driver/test/test_i2c_read.c
```diff
@@ -0,0 +1,252 @@
+#include "i2c_read.h"
```
copperium, line 1: I don't think you committed the new changes in `test_i2c_read.c`, this seems to be an old version of the tests.

copperium (changes requested):
Proper review now: nice job, a bunch of things to fix.

projects/baby_driver/src/i2c_read.c
ryandancy: Please also rename the parameter!

projects/baby_driver/test/test_i2c_read.c
```diff
@@ -0,0 +1,213 @@
...
+
+  return STATUS_CODE_OK;
+}
+
+void setup_test(void) {
+  initialize_can_and_dependencies(&s_can_storage, SYSTEM_CAN_DEVICE_BABYDRIVER, TEST_CAN_EVENT_TX,
+                                  TEST_CAN_EVENT_RX, TEST_CAN_EVENT_FAULT);
+  TEST_ASSERT_OK(dispatcher_init());
+  dispatcher_register_callback(BABYDRIVER_MESSAGE_I2C_READ_DATA, prv_rx_i2c_read_callback, NULL);
+  i2c_read_init(I2C_READ_SOFT_TIMER_TIMEOUT_MS);
```
ryandancy, line 71: It looks like you didn't update this when you changed the name of the `I2C_READ_SOFT_TIMER_TIMEOUT_MS` macro, please make sure this test file compiles and passes with `make test PLATFORM=x86 PROJECT=baby_driver TEST=i2c_read` and then I'll look at it further.

ryandancy (changes requested):
This will probably be the last round of review - a few things to clean up, mostly in the tests. There's still one comment from earlier that needs to be addressed, about `status_ok_or_return`; it might be marked as "Outdated" but that's just because something else in the block of code I commented on got changed in one of the commits.

ryandancy (approved):

PR #370, opened 2021-05-08
[SOFT-484] Implement viewing and updating stores

jessm:
This PR implements viewing and updating stores.

### Misc changes
- small improvements:
    - make all 'sensor' driver (adc, ads1259, ads1015, adt7476a) actually export their stores, so that when viewing stores you always have a complete picture
    - set up CAN in server.py, so if CAN isn't set up yet it'll work out of the box. it'll still fail (could be more graceful but whatever) if you try to run it with not vcan0 and you don't have the network set up.
    - make store.c not segfault if the update isn't expected. This could happen if a store isn't initialized and the store is updated. it's kind of an unnecessary check but I still want it because of the new ability to set stores from python.
    - req.py returns the request object instead of just the text, so that callers can check the status code. this is why there's a bunch of random changes in sims.py.
    - in decoder.py, factor out `store_from_name` and `store_from_enum` as other helper functions, since it's useful for api access to stores
    - factor out `sim_from_name` in pm.py since it's a useful helper function for api access to sims
### Viewing stores
- in args.py, set it so only the sim argument is required. if store type isn't specified, print all stores.
- set up stores.py as the container for store related functions
- just pretty print the json
- in board_sim.py, implement `stores_str_lookup` for looking up the store by the type name using a bit of protobuf reflectino magic
### Updating stores
- store type is a required arg, but key is still optional and is assumed 0 if not present
- Flow is as follows for muctl:
1. Grab the raw json from the `view` command
2. write it to a temp file
3. open vim on the temp file to allow for editing
4. read back the temp file into a string
5. use the `apply` method on the server with the updated store
- On the handler side, the apply method works by:
1. reading json from request body using the `Content-Length` header
2. getting a new appropriate store object from the `decoder` function
3. loading the json string into the store object using google's protobuf library. There's a weird intermediate step that seems necessary but I don't know why (loading it as a dict first, then parsing it into the actual message object)
4. get a full mask from a decoder helper function, because we want to apply all values. yes this can have unintended side effects if the stores update while the manual update is in progress, but I don't know how to deal with this
5. write the store and mask into the project
# Ok but that's kind of annoying, what about an easy way?
### SimIo / getters and setters
- a SimIo is a simple class that holds a getter and a setter, and has a name
- Any sim can register a new SimIo in `pm`
- This makes it really easy to set up new controls. This is meant for really simple things like toggling buttons or sensor values.
- idk it's pretty self explanatory, the commit is nicely separated from the rest of the stuff if you want to look at it individually

### DIFF
--- a/mu/harness/pm.py
+++ b/mu/harness/pm.py
@@ -89,6 +89,13 @@ class ProjectManager:
 
     def stop(self, sim):
         del self.fd_to_sim[sim.fd]
+        # avoid deleting from self.simios during iteration
+        simios = []
+        for name, simio in self.simios.items():
+            if simio.sim == sim:
+                simios.append(name)
+        for simio in simios:
+            del self.simios[simio]
         sim.stop()
 
     def stop_name(self, sim_name):
@@ -186,7 +193,7 @@ class ProjectManager:
 
     def get_all_io(self):
         ret = {}
-        for name, simio in self.simios:
+        for name, simio in self.simios.items():
             ret[name] = simio.get()
         return ret
 
--- a/mu/harness/pm.py
+++ b/mu/harness/pm.py
@@ -30,6 +30,7 @@ class ProjectManager:
         print('Configuration:', config)
         self.config = config
         self.fd_to_sim = {}
+        self.simios = {}
         self.proj_name_list = os.listdir(os.path.join(REPO_DIR, 'projects'))
         self.sim_catalog = self.sim_cat()
         self.killed = False
@@ -174,6 +175,23 @@ class ProjectManager:
                     continue
                 catalog[sim_name] = sim_class
         return catalog
+
+    def new_io(self, simio):
+        if simio.name in self.simios:
+            raise ValueError('SimIo name taken')
+        self.simios[simio.name] = simio
+
+    def get_io(self, name):
+        return self.simios[name].get()
+
+    def get_all_io(self):
+        ret = {}
+        for name, simio in self.simios:
+            ret[name] = simio.get()
+        return ret
+
+    def set_io(self, name, val):
+        self.simios[name].set(val)
 
     def end(self):
         self.killed = True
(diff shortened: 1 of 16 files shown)


mu/harness/pm.py
```diff
@@ -172,6 +176,23 @@ def sim_cat(self):
...
+        if simio.name in self.simios:
+            raise ValueError('SimIo name taken')
+        self.simios[simio.name] = simio
+
+    def get_io(self, name):
+        return self.simios[name].get()
+
+    def get_all_io(self):
+        ret = {}
+        for name, simio in self.simios:
```
ryandancy, line 189: If I do:
```
musrv -bg
muctl start PedalBoard
muctl get
```
then I get `too many values to unpack (expected 2)` and I suspect it might be because you need to iterate over `self.simios.items()` rather than just `self.simios` here.

mu/sims/pedal_board.py
```diff
@@ -1,7 +1,22 @@
...
-from mu.sims.sub_sims.ads1015 import Ads1015
+from mu.harness.sim_io import SimIo
+from mu.sims.sub_sims.ads1015 import Ads1015, ADS1015_KEY
 
 
 class PedalBoard(BoardSim):
     def __init__(self, pm, proj_name='pedal_board'):
         super().__init__(pm, proj_name, sub_sim_classes=[Ads1015])
+        self.pm.new_io(SimIo('throttle', self.get_throttle, self.set_throttle))
+        self.pm.new_io(SimIo('brake', self.get_brake, self.set_brake))
```
ryandancy, line 10: These need to get cleaned up somehow when a sim gets stopped, otherwise this will error out:
```
musrv -bg
muctl start PedalBoard
muctl stop PedalBoard
muctl start PedalBoard # gives 'SimIo name taken'
```
and after that `muctl get` refers to the wrong PedalBoard instance (and `muctl set` is timing out for some reason).

PR #372, opened 2021-05-12
[SOFT_482] Regen Brake Toggling

michael1ding:


projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+  gpio_init();
+  interrupt_init();
+  event_queue_init();
+  gpio_it_init();
+  soft_timer_init();
+  TEST_ASSERT_OK(steering_digital_input_init());
+  TEST_ASSERT_OK(can_init(&s_can_storage, &can_settings));
+}
+
+void test_steering_digital_input_cc_toggle() {
```
ryandancy, line 49: Nitpick: this parameter list should be `(void)` rather than `()`. Also I think you forgot to rename the test? This isn't testing steering_digital_input nor is it testing cruise control toggling...

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+                                         prv_test_regen_toggle_rx_cb_handler, NULL));
+  GpioAddress regen_toggle_address = REGEN_BRAKE_TOGGLE_GPIO_ADDR;
+  TEST_ASSERT_OK(gpio_it_trigger_interrupt(&regen_toggle_address));
+  Event e = { 0 };
+  MS_TEST_HELPER_ASSERT_NEXT_EVENT(e, (EventId)STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE,
+                                   (uint16_t)GPIO_STATE_LOW);
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+  TEST_ASSERT_OK(steering_can_process_event(&e));
+  MS_TEST_HELPER_CAN_TX_RX(STEERING_CAN_EVENT_TX, STEERING_CAN_EVENT_RX);
+  TEST_ASSERT_EQUAL(1, count);
```
ryandancy, line 60: This test seems to mostly be testing the gpio interrupt logic in steering_digital_input.c rather than exclusively the logic in steering_can.c. Since this is the test file for steering_can.c and test_steering_digital_input.c already tests the logic in steering_digital_input.c, I think you can just test steering_can.c by constructing the events directly (like `Event e = { STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE, 0 }` or similar) and passing those events directly into `steering_can_process_event`.

(It's also a little weird to have a test for the regen braking CAN message but not the rest of them, but I think that's ok.)

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+  return STATUS_CODE_OK;
+}
+
+void setup_test(void) {
+  gpio_init();
+  interrupt_init();
+  event_queue_init();
+  gpio_it_init();
+  soft_timer_init();
+  TEST_ASSERT_OK(steering_digital_input_init());
```
ryandancy, line 45: After making the change to only test steering_can, you won't have to initialize steering_digital_input or gpio interrupts.

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+  TEST_ASSERT_OK(can_init(&s_can_storage, &can_settings));
```
ryandancy, line 46: You can actually use `initialize_can_and_dependencies` from ms_test_helpers_can.h to initialize gpio, interrupts, the event queue, soft timers, and CAN all at once and not have to worry about a bunch of the fields of the can settings.

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+#include "ms_test_helpers.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
+#include "steering_digital_input.h"
+#include "steering_events.h"
+#include "test_helpers.h"
+
+#define REGEN_BRAKING_TOGGLE_CAN \
+  { .id = 12, .data = 0 }
```
ryandancy, line 17: This macro doesn't seem to be used? Please remove

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+  Event e = { 0 };
+  MS_TEST_HELPER_ASSERT_NEXT_EVENT(e, (EventId)STEERING_DIGITAL_INPUT_REGEN_BRAKE_TOGGLE,
+                                   (uint16_t)GPIO_STATE_LOW);
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+  TEST_ASSERT_OK(steering_can_process_event(&e));
+  MS_TEST_HELPER_CAN_TX_RX(STEERING_CAN_EVENT_TX, STEERING_CAN_EVENT_RX);
+  TEST_ASSERT_EQUAL(1, count);
+}
+
+void teardown_test(void) {}
```
ryandancy, line 63: Nitpick: we usually write teardown_test immediately after setup_test.

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+static CanSettings can_settings = { .device_id = SYSTEM_CAN_DEVICE_STEERING,
+                                    .bitrate = CAN_HW_BITRATE_500KBPS,
+                                    .rx_event = STEERING_CAN_EVENT_RX,
+                                    .tx_event = STEERING_CAN_EVENT_TX,
+                                    .fault_event = STEERING_CAN_FAULT,
+                                    .tx = { GPIO_PORT_A, 12 },
+                                    .rx = { GPIO_PORT_A, 11 },
+                                    .loopback = true };
+
+static int count = 0;
```
ryandancy, line 30: I like that you're using a static count to make sure the regen braking toggle request was sent! A couple points about it:
* you should reset it to 0 in setup_test in case other tests use it in the future
* please use a fixed-width int type like `uint8_t` instead of `int` so it's consistent on x86 and stm32
* since it's a static variable, it should be prefixed with `s_`. Also, `count` is kind of a generic name - how about something that gives an indication of its use a bit more, like `s_msg_rx_count`?

projects/steering/test/test_steering_can.c
```diff
@@ -0,0 +1,63 @@
...
+#include "can_msg_defs.h"
+#include "event_queue.h"
+#include "exported_enums.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt_def.h"
+#include "ms_test_helpers.h"
+#include "soft_timer.h"
+#include "status.h"
+#include "steering_can.h"
```
ryandancy, line 11: Small nitpick: the coding standards specify that the #include for the module you're testing (in this case `steering_can.h`) should go at the very top of the includes, separated from the others by a blank line.

ryandancy (changes requested):
Nice job Michael! Just a bunch of comments about the tests.

ryandancy (approved):

PR #375, opened 2021-05-15
[SOFT-415] Can Datagram 

mitchellostler:
Brief description of how Can Datagram works (see wiki [here](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2030895112/Bootloader+Flash+over+CAN) for more info):

Effectively, there is one state machine which is used for both rx and tx. 

For tx, message are sent sequentially through the state machine as fast as they are processed. `can_datagram_start_tx()` starts the process, calling the tx callback with start=true, and raising the first state machine event.

For rx, an init message must be recieved  for processing to start. messages are received into a circular buffer, and then processed with sequential calls to can_datagram_rx.

To read/write datagram data, you must pass a buffer either with the relevant information, or with enough room to capture it.


AveryChiu64 (commented):
Great work! This is clean. Though, it looks like the build is failing due to some formatting issues. Could you address this?

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,471 @@
...
+
+// RX Circular buffer functions
+static void prv_init_rx_buffer(void) {
+  s_rx_msg_buf.head = s_rx_msg_buf.buf;
+  s_rx_msg_buf.tail = s_rx_msg_buf.head;
+  s_data_write_count = 0;
+}
+
+static void prv_write_rx_buffer(uint8_t *data, size_t len) {
+  if (s_rx_msg_buf.head - s_rx_msg_buf.buf > DATAGRAM_RX_BUFFER_LEN) {
```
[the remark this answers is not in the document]
mitchellostler: Yeah good point. I added a catch to exit with an error. 
We can't really add any other failsafe to wait for read to happen I thinkl, as the point of the buffer itself was to do this, so failing out should be enough

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,471 @@
...
+  FSM_ADD_TRANSITION(DATAGRAM_EVENT_ERROR, state_done);
+}
+
+FSM_STATE_TRANSITION(state_dt_type) {
+  FSM_ADD_TRANSITION(DATAGRAM_EVENT_DST_LEN, state_dst_len);
+  FSM_ADD_TRANSITION(DATAGRAM_EVENT_DT_TYPE, state_dt_type);
+  FSM_ADD_TRANSITION(DATAGRAM_EVENT_ERROR, state_done);
+}
+
+FSM_STATE_TRANSITION(state_dst_len) {
```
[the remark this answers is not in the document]
mitchellostler: DST is the notation they use for "Destination nodes", which is the ids of all the nodes the message needs to reach. `dst_len` is essentially the number of uint8 node ids the datagram has. It is definitely not the most clear, I could change 'dst' to 'dest' or 'nodes' if that's clearer, or I could just add some comments?

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,476 @@
...
+  fsm_state_init(state_data, prv_process_data);
+  fsm_state_init(state_done, prv_process_done);
+}
+
+static void prv_rx_watchdog_expiry_cb(void *context) {
+  event_raise_no_data(DATAGRAM_EVENT_ERROR);
+  LOG_DEBUG("WATCHDOG EXPIRED. EXITING WITH ERROR...\n");
+}
+
+StatusCode can_datagram_init(CanDatagramSettings *settings) {
```
ryandancy, line 385: The API for this module is a little weird - to send/receive multiple messages you'd have to reinitialize it! Here is my idea for how to make the API more intuitive:
- make `can_datagram_init` a thin init function that only initializes the fsm and status and stuff like that
- give `state_done` a way to return to `state_idle` once we're done rxing/txing a datagram
- have `can_datagram_start_listening` and `can_datagram_start_tx` functions which can only be called in `state_idle`, and which start an RX or TX respectively
  - `can_datagram_start_listening` initializes everything for RX mode, and takes the data buffer where we put the datagram data
  - `can_datagram_start_tx` initializes everything for TX mode, takes all the datagram fields, and starts the TX
- also give users a way to know if we're currently in the middle of a tx/rx so they know if it's ok to start listening/txing
- set the event data in all the events (or at least the completion event) to the state we're in (tx/rx) so that downstream clients know if we've just txed or rxed
- document all this in the header!

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,476 @@
...
+      }
+    }
+    event_raise_no_data(DATAGRAM_EVENT_DST);
+  } else {
+    uint8_t *rx_data;
+    size_t len;
+    if (!prv_read_rx_buffer(&rx_data, &len)) {
+      event_raise_no_data(DATAGRAM_EVENT_DST);
+    } else {
+      memcpy(dt->destination_nodes + s_store.rx_bytes_read, rx_data, len);
```
ryandancy, line 280: I think we should take in a node id and then return early from rxing any datagrams that don't have a destination node matching our node id since they aren't meant for us. That also might mean we don't have to store the destination nodes!

libraries/ms-common/test/test_can_datagram.c
```diff
@@ -0,0 +1,598 @@
...
+    .crc = crc,
+    .dt_type = 3,
+    .destination_nodes_len = dst_size,
+    .destination_nodes = dst_data,
+    .data_len = data_size,
+    .data = data_data,
+  };
+  switch (s_num_tx) {
+    case 0: {
+    }
```
ryandancy, line 112: Why the `{}` on all the case labels?

libraries/ms-common/inc/can_datagram.h
```diff
@@ -0,0 +1,114 @@
...
+  CAN_DATAGRAM_TYPE_D,
+  NUM_CAN_DATAGRAM_TYPES,
+} CanDatagramType;
+
+typedef struct RxBufStore {
+  uint8_t data[8];
+  size_t len;
+} RxBufStore;
+
+typedef struct DatagramRxBuffer {
```
ryandancy, line 60: Some of these structs are only used in the implementation I believe, see if you can move any of them to the C file so they aren't polluting client code namespaces.

ryandancy (changes requested):
Didn't look at the tests, but wow this is way more involved that I thought it would be. Looks awesome.

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,476 @@
...
+  write += DEST_LEN_SIZE_BYTES;
+  memcpy(write, &dst_crc, sizeof(uint32_t));
+  write += sizeof(uint32_t);
+  memcpy(write, &dt->data_len, DATA_LEN_SIZE_BYTES);
+  write += DATA_LEN_SIZE_BYTES;
+  memcpy(write, &data_crc, sizeof(uint32_t));
+  write += sizeof(uint32_t);
+
+  // process final crc
+  return crc32_arr(stream, (size_t)(write - stream));
```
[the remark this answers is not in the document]
mitchellostler: Could also just condense the metadata section to `memcpy(stream, &dt, sizeof(dt))` and then append everything in the same way? This would probably be cleaner but will be more opaque and use more byte space
```
static uint32_t prv_can_datagram_compute_crc(void) {
  uint8_t stream[sizeof(CanDatagram) + 2*sizeof(uint32_t)];
  uint8_t *write = stream;
  CanDatagram *dgram = &s_store.dgram;

  // CRC for data and dst_nodes
  uint32_t dst_crc = crc32_arr(dgram->destination_nodes, dgram->destination_nodes_len);
  uint32_t data_crc = crc32_arr(dgram->data, dgram->data_len);

  // Copy dgram with metadata to stream
  memcpy(stream, &dgram, sizeof(CanDatagram));
  write += sizeof(CanDatagram);
  memcpy(write, &data_crc, sizeof(uint32_t));
  write += sizeof(uint32_t);
  memcpy(write, &data_crc, sizeof(uint32_t));
  write += sizeof(uint32_t);
  return crc32_arr(stream, (size_t)(write - stream));
  }
  ```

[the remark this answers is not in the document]
Comment (ryandancy) [reply libraries/ms-common/src/can_datagram.c:157]: Hm, with that I'm concerned that with the datagram not being packed the compiler might add padding in places that would be hard to replicate in Python - I like the original way because it's more explicit.

**Comment by mitchellostler** [comment]
Still working on it, just wanted to have my work in remote lol

File: libraries/ms-common/src/can_datagram.c [hunk 8]
```diff
@@ -0,0 +1,616 @@
...
+#define CAN_BUFFER_SIZE 8
+
+static Fsm s_dgram_fsm;
+static CanDatagramStorage s_store;
+static WatchdogStorage s_watchdog;
+
+static uint8_t s_can_buffer[CAN_BUFFER_SIZE];
+
+static Fifo s_fifo;
+static uint8_t s_tx_buffer[DATAGRAM_RX_BUFFER_LEN];
```
mitchellostler, line 39: @ryandancy With the new structure, the fifo where data is and process is 2048 bytes long. It is one instance which is used for both tx/rx (I am renaming it to s_buffer), and all received/transmitted data is processed through here. 

If I make it any smaller than 2048, then it overflows when the `data` segment received is large. This also has the added bonus of allowing me to read data and destination_node buffers in one call to fifo_pop_arr()  *See below*

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,616 @@
...
+//    LOG_DEBUG("RX LEN: %zu, %d\n", store->rx_bytes_read, dgram->destination_nodes_len);
+//    if (store->rx_bytes_read == dgram->destination_nodes_len) {
+//      LOG_DEBUG("data_len\n");
+//      event_raise_no_data(store->rx_event);
+//      store->rx_bytes_read = 0;
+//    } else {
+//      event_raise_no_data(store->repeat_event);
+//    }
+//  }
+  if (!prv_read_rx_buffer(dgram->destination_nodes, dgram->destination_nodes_len)) {
```
mitchellostler, line 423: @ryandancy (part 2) Because buffer is large enough to hold both the largest possible destination_nodes buffer (255) and  the largest possible data buffer (2048), we can read directly from the fifo in bulk, and eliminate a bunch of code.

I admit that this does seem a bit sus for `data`, as if it is the largest possible length it will fit the buffer exactly, but we can assume that for any given message, all other messages will be already read.

Tested it with max conditions and it is working, just wanted to clarify the idea here

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,616 @@
...
+      event_raise_no_data(store->rx_event);
+    }
+  }
+}
+ 
+static void prv_process_done(Fsm *fsm, const Event *e, void *context) {
+  CanDatagramStorage *store = context;
+  // Handle tx specific cleanup
+  if (e->id == store->tx_event) {
+    //tx any final data
```
mitchellostler, line 485: This is valid, as the max length pushed to the fifo for tx is 8 bytes, so we won't lose any data

mitchellostler:
Alrighty, I've updated the API, fixed the PR comments, and added a few more tests. Everything seems to be working.
It's worth noting though, that our watchdog library seems to be having some issues giving some weird behaviour, I'm not to sure what's up with that.

I will improve the explanation in the header and rewrite the SDD once the changes are approved 🥇

libraries/ms-common/src/can_datagram.c
mitchellostler: - [ ] make can_datagram_init a thin init function that only initializes the fsm and status and stuff like that

- [ ] give state_done a way to return to state_idle once we're done rxing/txing a datagram

- [ ] have can_datagram_start_listening and can_datagram_start_tx functions which can only be called in state_idle, and which start an RX or TX respectively

- [ ] can_datagram_start_listening initializes everything for RX mode, and takes the data buffer where we put the datagram data
    - Now using tx and rx config structs  

- [ ] can_datagram_start_tx initializes everything for TX mode, takes all the datagram fields, and starts the TX

also give users a way to know if we're currently in the middle of a tx/rx so they know if it's ok to start listening/txing
set the event data in all the events (or at least the completion event) to the state we're in (tx/rx) so that downstream clients know if we've just txed or rxed
   - there are now two different statuses used for tx/rx completion

libraries/ms-common/src/watchdog.c
```diff
@@ -18,3 +18,7 @@ void watchdog_kick(WatchdogStorage *storage) {
   soft_timer_cancel(storage->timer_id);
   soft_timer_start_millis(storage->timeout_ms, prv_expiry_callback, storage, &storage->timer_id);
 }
+
+void watchdog_cancel(WatchdogStorage *storage) {
+  soft_timer_cancel(storage->timer_id);
```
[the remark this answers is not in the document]
mitchellostler: There may be some issues in the Watchdog changes or library in general, I was getting some weird behaviour. Not too sure what it is though

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,634 @@
...
+  // Set up rx fifo for rxed datagram messages, zero data buffers
+  if (s_store.status == DATAGRAM_STATUS_ACTIVE) {
+    return STATUS_CODE_UNREACHABLE;
+  }
+  // Set up data buffers
+  CanDatagram *dgram = &s_store.dgram;
+  dgram->data = config->data;
+  dgram->destination_nodes = config->destination_nodes;
+
+  // Store config * to be filled out after rx complete
```
ryandancy, line 588: Might want to specify in the header that the `CanDatagramRxConfig` pointer has to remain valid until the datagram is fully sent, but the `CanDatagramTxConfig` pointer for `can_datagram_start_tx` doesn't (although its `data` field does).

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,634 @@
...
+}
+
+// Functions used to write and read to fifo
+static void prv_flush_tx_buffer(void) {
+  size_t tx_len = fifo_size(&s_fifo) < CAN_BUFFER_SIZE ? fifo_size(&s_fifo) : CAN_BUFFER_SIZE;
+  for (size_t i = 0; i < tx_len; i++) {
+    fifo_pop(&s_fifo, &s_can_buffer[i]);
+  }
+  fifo_pop_arr(&s_fifo, s_can_buffer, tx_len);
+  s_store.tx_cb(s_can_buffer, tx_len, false);
```
ryandancy, line 232: Even though you've already null checked elsewhere, maybe surround this with a null check just in case :)

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,634 @@
...
+    return STATUS_CODE_UNINITIALIZED;
+  }
+
+  // Rx normal data
+  watchdog_kick(&s_watchdog);
+  prv_write_rx_buffer(data, len);
+  return STATUS_CODE_OK;
+}
+
+StatusCode can_datagram_start_listener(CanDatagramRxConfig *config) {
```
ryandancy, line 578: What's the strategy for notifying client code when we've successfully received a datagram? Currently we seem to just notify by setting the status - maybe we should take an optional "rx success" callback? (And on that note - perhaps an optional "tx success" callback in `can_datagram_start_tx` and an optional "error" callback in `can_datagram_init` (or maybe both tx and start listener functions) for hard errors?)

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,634 @@
...
+  dgram->protocol_version = CAN_DATAGRAM_VERSION;
+  dgram->dgram_type = config->dgram_type;
+  dgram->destination_nodes_len = config->destination_nodes_len;
+  dgram->data_len = config->data_len;
+  dgram->destination_nodes = config->destination_nodes;
+  dgram->data = config->data;
+  dgram->crc = prv_can_datagram_compute_crc();
+
+  s_store.tx_cb = config->tx_cb;
+  s_store.tx_cb(NULL, 0, true);
```
ryandancy, line 622: Hm, I suppose this does conform to the standard - my original thought was that the start message would also contain data, so we'd just set the start flag to true on the first tx_cb call. That way is more efficient especially for short messages, but this is simpler - your call on whether to change it or not.

ryandancy (changes requested):
I haven't looked through the tests but it looks great! Pretty much all of the comments are minor stuff that's quick to fix, thanks for getting this through.

libraries/ms-common/src/can_datagram.c
```diff
@@ -0,0 +1,634 @@
...
+      // tx any final data
+      if (fifo_size(&s_fifo) > 0) {
+        prv_flush_tx_buffer();
+      }
+      event_raise_no_data(store->tx_event);
+      s_store.tx_bytes_sent = 0;
+      return;
+    }
+  }
+  event_raise_no_data(store->repeat_event);
```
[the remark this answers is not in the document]
mitchellostler: I realized I am handling the final messages twice, just do it at the end here now

ryandancy (commented):
Looked it all over and it looks good, just need the endianness-independence change + whatever CRC change you're making

ryandancy (approved):

PR #376, opened 2021-05-16
soft 999 shicheng lu fw 102

ShiCheng-Lu:
completed fw 102 counter homework

### DIFF
--- a/projects/soft_999_shicheng_lu_fw_102/src/main.c
+++ b/projects/soft_999_shicheng_lu_fw_102/src/main.c
@@ -15,11 +15,11 @@ typedef struct Counters {
 static void prv_increment_counters(SoftTimerId id, void *context) {
   Counters *counters = context;
 
-  counter->counter_a++;
-  LOG_DEBUG("Counter A: %i\n", counter->counter_a);
-  if (counter->counter_a % 2 == 0) {
-    counter->counter_b++;
-    LOG_DEBUG("Counter B: %i\n", counter->counter_b);
+  counters->counter_a++;
+  LOG_DEBUG("Counter A: %i\n", counters->counter_a);
+  if (counters->counter_a % 2 == 0) {
+    counters->counter_b++;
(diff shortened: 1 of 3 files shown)


projects/soft_999_shicheng_lu_fw_102/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include "wait.h"
+
+#define DELAY_MS 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void increment_counters(SoftTimerId id, void *context) {
```
AveryChiu64, line 15: This function should be static and you should append a prv_ prefix in the name of the function to indicate it is private.

Eg
`prv_increment_counters_callback`

projects/soft_999_shicheng_lu_fw_102/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+  Counters *counter = context;
```
AveryChiu64, line 16: I'm not a big fan of naming it `counter` since it technically holds counter a and counter b, maybe we could give this a better name?

`counter` sort of implies its just a single counter

projects/soft_999_shicheng_lu_fw_102/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  soft_timer_start_millis(DELAY_MS, increment_counters, context, NULL);
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counter = { 0 };
```
AveryChiu64, line 32: See comment about naming

AveryChiu64 (commented):
image

Please address the build issues

AveryChiu64 (commented):
Looks good. Please close this PR

PR #377, opened 2021-05-16
Soft 999 shicheng lu fw 103

ShiCheng-Lu:
completed homework

### DIFF
--- a/projects/soft_999_shicheng_lu_fw_103/src/main.c
+++ b/projects/soft_999_shicheng_lu_fw_103/src/main.c
@@ -7,11 +7,11 @@
 #include "wait.h"
 
 // LOG_DEBUG the adc converted value of context pin when called (context is A6 pin)
-void log_adc_a6(const GpioAddress *address, void *context) {
-  GpioAddress *pin_a6_addr = context;
-  uint16_t pin_a6_data = 0;
-  adc_read_converted_pin(*pin_a6_addr, &pin_a6_data);
-  LOG_DEBUG("adc converted pin A6 data: %d\n", pin_a6_data);
+static void priv_log_adc_value(const GpioAddress *address, void *context) {
+  GpioAddress *adc_addr = context;
+  uint16_t adc_data = 0;
+  adc_read_converted_pin(*adc_addr, &adc_data);
+  LOG_DEBUG("adc converted pin A6 data: %d\n", adc_data);
 }
 
 int main(void) {
@@ -20,17 +20,24 @@ int main(void) {
   adc_init(ADC_MODE_SINGLE);
 
   // output gpio pin A6
-  GpioAddress analog_pin_a6_addr = { .port = GPIO_PORT_A, .pin = 6 };
-  GpioSettings analog_pin_a6_settings = {
+  GpioAddress adc_addr = {
+    .port = GPIO_PORT_A,
+    .pin = 6,
+  };
+  GpioSettings adc_settings = {
     GPIO_DIR_IN,
     GPIO_STATE_LOW,
     GPIO_RES_NONE,
     GPIO_ALTFN_ANALOG,
   };
-  gpio_init_pin(&analog_pin_a6_addr, &analog_pin_a6_settings);
+  gpio_init_pin(&adc_addr, &adc_settings);
+  adc_set_channel_pin(adc_addr);
 
   // button gpio pin
-  GpioAddress button_addr = { .port = GPIO_PORT_B, .pin = 2 };
+  GpioAddress button_addr = {
+    .port = GPIO_PORT_B,
+    .pin = 2,
+  };
   GpioSettings button_settings = {
     GPIO_DIR_IN,
     GPIO_STATE_LOW,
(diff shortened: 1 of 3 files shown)


projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+  GpioAddress *pin_a6_addr = context;
+  uint16_t pin_a6_data = 0;
+  adc_read_converted_pin(*pin_a6_addr, &pin_a6_data);
+  LOG_DEBUG("adc converted pin A6 data: %d\n", pin_a6_data);
+}
+
+int main(void) {
+  interrupt_init();
+  gpio_init();
+  adc_init(ADC_MODE_SINGLE);
```
AveryChiu64, line 20: You will need to call `adc_set_channel_pin` after you've initialized your gpio pin for the adc.
With the new functions in the ADC library, what happens is that you pass in a pin and a channel is handled for you so you don't have to do it manually.

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+
+  // output gpio pin A6
+  GpioAddress analog_pin_a6_addr = { .port = GPIO_PORT_A, .pin = 6 };
```
AveryChiu64, line 23: This is a bit nitpicky, but you should place a comma after the last value (ie place a comma after the '6'.
This will allow the formater to fix the formatting of this and change it to look something like:
```
GpioAddress analog_pin_a6_addr = {
    .port = GPIO_PORT_A,
    .pin = 6,
  };
```

Also the naming is sort of inconsistent since you name this pin `analog_pin_a6_addr`, but then later for your button pin, you simply name it `button_addr`. Please try to keep it consistent, I would probably just rename 
`analog_pin_a6_addr` to something like 'adc_addr'

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+  GpioSettings analog_pin_a6_settings = {
+    GPIO_DIR_IN,
+    GPIO_STATE_LOW,
+    GPIO_RES_NONE,
+    GPIO_ALTFN_ANALOG,
+  };
+  gpio_init_pin(&analog_pin_a6_addr, &analog_pin_a6_settings);
+
+  // button gpio pin
+  GpioAddress button_addr = { .port = GPIO_PORT_B, .pin = 2 };
```
AveryChiu64, line 33: Same comment as above. Please place a comma after the 2

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,54 @@
...
+    GPIO_DIR_IN,
+    GPIO_STATE_LOW,
+    GPIO_RES_NONE,
+    GPIO_ALTFN_NONE,
+  };
+  gpio_init_pin(&button_addr, &button_settings);
+
+  InterruptSettings interrupt_settings = {
+    INTERRUPT_TYPE_INTERRUPT,
+    INTERRUPT_PRIORITY_HIGH,
```
AveryChiu64, line 44: Any reason why the interrupt priority is HIGH? Normally we would use this for interrupts that are urgent.

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,54 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "interrupt_def.h"
+#include "log.h"
+#include "wait.h"
+
+// LOG_DEBUG the adc converted value of context pin when called (context is A6 pin)
+void log_adc_a6(const GpioAddress *address, void *context) {
```
AveryChiu64, line 10: This should be static and prefix with prv_
eg `prv_log_adc_value`

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "interrupt_def.h"
+#include "log.h"
+#include "wait.h"
+
+// LOG_DEBUG the adc converted value of context pin when called (context is A6 pin)
+static void priv_log_adc_value(const GpioAddress *address, void *context) {
```
AveryChiu64, line 10: Prefix should be `prv`

AveryChiu64 (commented):
image

Looks like the build is failing. Could you resolve this?

https://github.com/uw-midsun/firmware_xiv/runs/2614433140?check_suite_focus=true

projects/soft_999_shicheng_lu_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+#include "interrupt.h"
+#include "interrupt_def.h"
+#include "log.h"
+#include "wait.h"
+
+// LOG_DEBUG the adc converted value of context pin when called (context is A6 pin)
+static void prv_log_adc_value(const GpioAddress *address, void *context) {
+  GpioAddress *adc_addr = context;
+  uint16_t adc_data = 0;
+  adc_read_converted_pin(*adc_addr, &adc_data);
```
AveryChiu64, line 13: This is good. However, notice that this function returns a value of type `StatusCode` which tells us if the function works properly. What happens if this function fails and returns a failure rather than OK. 

Right now we aren't checking for it, so if we got bad data from the function we would not know and we would be printing it out still. It would be best to add a check for this.

PR #378, opened 2021-05-18
[SOFT-446] bootloader protos makefile changes and added .proto files

Kyle02:


Makefile
```diff
@@ -23,12 +23,13 @@
 #   make remake [PL] [PR] [SD] [DF] - Cleans and rebuilds the target project (does not force-rebuild dependencies)
 #   make test [PL] [PR|LI] [TE] [SD] [DF] - Builds and runs the specified unit test, assuming all tests if TE is not defined
 #   make pytest [PR] [TE] - Runs the specified python unit test, assuming all tests in scripts directory of a project if TE is not defined
-#   make pytest_all - Runs all python tests in the scripts directory of every project 
+#   make pytest_all - Runs all python tests in the scripts directory of every project
```
AveryChiu64, line 26: There seems to be a lot of these strange formatting changes in this file. Were these changes made by our formatter?

AveryChiu64 (approved):

AveryChiu64 (commented):
This looks good
Maybe we want to check with @ryandancy before merging, but from what I see this looks fine and matches with what is here:
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2030895112/Bootloader%2BFlash%2Bover%2BCAN

Makefile
Kyle02: yea this was made by running the make clean function i think. I was using that because i changed something in that function.

ryandancy: probably your IDE removing trailing spaces, I think it's fine

projects/bootloader/protos/querying_bootloader.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+message QueryingMCU {
```
ryandancy, line 3: What's the `MCU` suffix?

Also I don't think you need the `_bootloader` suffix in this file and querying_response_bootloader.proto.

Kyle02: oh that was originally to distinguish between the response and client, but i figured out a better way to do it, some im just removing that

PR #379, opened 2021-05-18
[SOFT-381] Add mcp3427 support and Solar sim

mitchellostler:
Added Mcp3427 driver sim with tests, allowing us to write the values that the adc will read (on both channels) as well as trigger the fault callback.

As it is an adc, there is no need to assert values, only write them, and the test verifies it is working by the smoketest output matching what is expected. 

Also added solar sim in same format as bms carrier, as from what I can see this is where the driver is used

There is some slight weirdness that I am not sure about, added it as comments to double check


mu/integration_tests/test_mcp3427.py
```diff
@@ -0,0 +1,25 @@
...
+from mu.integration_tests import int_test
+from mu.sims.bms_carrier import BmsCarrier
+
+class TestMcp3427(int_test.IntTest):
+    def setUp(self):
+        super().setUp()
+        self.board = self.manager.start(Solar, proj_name='smoke_mcp3427')
+
+    def test_mcp3427(self):
+        time.sleep(0.1)
```
mitchellostler, line 15: It seems like there might be some racing going on here. If I don't add this small delay, then the update store doesn't happen until after the next delay.

mu/sims/sub_sims/mcp3427.py
```diff
@@ -0,0 +1,25 @@
...
+from mu.protogen import mcp3427_pb2
+
+MCP3427_KEY = (stores_pb2.MuStoreType.MCP3427, 0)
+NUM_MCP3427_CHANNELS = 2
+
+class Mcp3427(SubSim):
+    # val1 writes channel1 of mcp3427, val2 channel2
+    def update_mcp3427(self, val1, val2, fault_flag):
+        mcp3427_msg = mcp3427_pb2.MuMcp3427Store()
+        mcp3427_msg.readings.extend([0] * NUM_MCP3427_CHANNELS)
```
mitchellostler, line 13: I feel like there's maybe another way to do this as mcp23008 also has a repeated value, but doesn't use it?
I get a keyerror without it

ryandancy: Just tested it, turns out it's actually a bug in the mcp23008 sim since the test didn't run the update_pin function. This way is correct.

ryandancy (approved):
LGTM!

PR #381, opened 2021-05-22
[SOFT-481] Update wavesculptor.h for WaveSculptor 22

hewittmcg:
There were some outdated message definitions in wavesculptor.h (I assume from when we used a WS20 motor controller).  This updates them to work with the WS22:

- Remove `WaveSculptorFanSpeedMeasurement`, `WaveSculptorAirInCpuTempMeasurement`, `WaveSculptorAirOutCapTempMeasurement`
- Add `WaveSculptorDspTempMeasurement` (I believe this is similar to `WaveSculptorAirInCpuTempMeasurement`), `WaveSculptorSlipSpeedMeasurement`, `WaveSculptorActiveMotorChange`
- Update other message internals where needed
- Define message offsets in hex to match format in WS user manual
- Update references

Note: just noticed that for cruise control, we currently send velocity assuming that it's in m/s, but the WS22 expects it in RPM.  This should be updated -- will probably need to get the wheel diameter from mech if they know it atm.


AveryChiu64 (approved):

projects/mci/inc/wavesculptor.h
```diff
@@ -16,61 +17,75 @@ typedef union WaveSculptorCanId {
...
+} WaveSculptorConfigId;
+
+// Motor Controller Base Addr + 0x12
+typedef struct WaveSculptorActiveMotorChange {
+  // Must spell "ACTMOT" in ASCII (0x54 4F 4D 54 43 41)
+  uint16_t access_key[3];
+
+  // Desired active motor (0 to 9)
+  uint16_t active_motor;
+} WaveSculptorActiveMotorChange;
```
ryandancy, line 56: Consider adding static_asserts for the size of these structs like you have below, unless these structs aren't sent verbatim to the wavesculptor?

ryandancy (approved):
LGTM

projects/mci/inc/wavesculptor.h
hewittmcg: Thanks, agreed, will add/push asserts in a sec.

Re: sending to the WS -- we don't send those structs verbatim and instead use `motor_can.c` to pack drive commands, which are the only ones we send to the MC atm.  This is in `prv_handle_drive()` in `mci_output.c`.  We also use a `MotorCanDriveCommand` struct defined in `motor_can.h` instead of the `WaveSculptorDriveCmd` struct from `wavesculptor.h`, although I believe they're functionally the same.

I tested briefly and it seems like using a `wavesculptor.h` `WaveSculptorCanData` union, filling the `WaveSculptorDriveCmd` field with the velocity/current values and directly sending the union's raw data to the WS should work the same.  This is on x86 so it's little endian like the STM32F0 (I think), but there's probably hardware weirdness I'm not accounting for.

Should probably be a separate ticket, but I think it makes sense to deprecate the `motor_can.c` implementation so we're not carrying around two different drive command definitions, if you agree.

ryandancy (approved):

PR #383, opened 2021-05-24
Firmware 102 HW

meganpark01:


AveryChiu64 (commented):
Code looks good. However, it looks like you mixed up the naming of the branch versus the project a bit.

`soft_999_Megan_Park_fw_102` should be your branch name instead of `102_hw_megan`
We name our branches this way to easily find them based off of their ticket number on the JIRA board. Eg if I see a ticket on JIRA numbered SOFT-137, I can easily find the branch that corresponds to the ticket, which should be something like `soft_137_steering`

The project name is up to you in this case, eg you can do something like `102_hw_megan` if you want. Normally we don't put the JIRA ticket number in the project names since a project may be composed of multiple JIRA tickets (eg BabyDriver)

ryandancy:
Yep, looks great - feel free to move onto FW 103 and fix the branch naming for that homework. Closing since homework is complete.

PR #385, opened 2021-05-28
firmware 102 hw

prajval-malhotra:
My firmware 102 hw for review

### DIFF
--- a/projects/soft_999_prajval_malhotra_fw_102/src/main.c
+++ b/projects/soft_999_prajval_malhotra_fw_102/src/main.c
@@ -1,0 +1,49 @@
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#include "delay.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// Counters structure with two counters
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+// function to increment and print counter a 
+void a_plus_one(SoftTimerId timer_id, void *context) {
+  Counters *obj = context;
+  obj->counter_a++;
+  LOG_DEBUG("A = %d\n", obj->counter_a);
+  soft_timer_start_millis(500, a_plus_one, obj, NULL);
+}
+
+// function to increment and print counter b
+void b_plus_one(SoftTimerId timer_id, void *context) {
+  Counters *obj = context;
+  obj->counter_b++;
+  LOG_DEBUG("B = %d\n", obj->counter_b);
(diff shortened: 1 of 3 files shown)


projects/soft_999_prajval_malhotra_fw_102/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+// function to increment and print counter a 
+void a_plus_one(SoftTimerId timer_id, void *context) {
+  Counters *obj = context;
+  obj->counter_a++;
+  LOG_DEBUG("A = %d\n", obj->counter_a);
+  soft_timer_start_millis(500, a_plus_one, obj, NULL);
```
ryandancy, line 22: Generally it's a good idea to factor out these constants like 500 into `#define`'d constants with good names so that we know what they represent, and if we have to change them we only have to do so in one place. Make sure you name them something descriptive and suffix them with the unit (in this case `_MS`).

projects/soft_999_prajval_malhotra_fw_102/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+  
+  Counters obj = { 0 };
+  
+  interrupt_init();
+  soft_timer_init();
+
+  soft_timer_start_millis(500, a_plus_one, &obj, NULL);
+  soft_timer_start_millis(1000, b_plus_one, &obj, NULL);
+
+  while (1) {
```
ryandancy, line 44: Nitpick: we usually use the constants `true` and `false` (from <stdbool.h>, which I notice you imported!) instead of raw 1 and 0 for things like infinite while loops.

projects/soft_999_prajval_malhotra_fw_102/src/main.c
```diff
@@ -0,0 +1,49 @@
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
```
ryandancy, line 3: Nitpick: you don't need <stdio.h> since you don't use any symbols from it.

ryandancy (changes requested):
Nice job Prajval, just a couple things to fix up! It looks like you may have gotten a bit mixed up between the project name and the branch name - for future onboarding homework the *branch* name should be `soft_999_prajval_malhotra_fw_[102/103/etc]`.

ryandancy:
Also make sure that you take a look at CI whenever it fails (see the "All checks have failed" at the bottom) - in this case it's failing because of formatting, so do `make format` to make sure your code is autoformatted properly. Also useful is `make lint` to make sure you don't have any linting errors.

prajval-malhotra:
Thanks Ryan, for the comprehensive feedback! I'll be sure to keep all these points in mind from next time.

PR #386, opened 2021-05-29
Soft 999 vaaranan yogalingam fw 102

vaaranan-y:
Hello,

This is my timer_challenge submission for the firmware 102 homework. Please let me know if I am missing anything. Thanks!

Vaaranan Yogalingam

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,7 +242,7 @@ lint_quick:
 
 # Globally disable the following pylint messages:
 PYLINT_DISABLE := \
-	import-error redefined-outer-name unused-argument \
+	consider-using-with import-error redefined-outer-name unused-argument \
 	too-few-public-methods duplicate-code no-self-use
 
 # Disable these additional pylint messages for MU:
--- a/projects/timer_challenge/src/main.c
+++ b/projects/timer_challenge/src/main.c
@@ -1,37 +1,33 @@
-// Firmware 102 Homework
-// Vaaranan Yogalingam
-// 2021-05-29
-
 #include <stdint.h>
 #include "interrupt.h"
 #include "log.h"
 #include "soft_timer.h"
 #include "wait.h"
-#define PERIOD 500
+#define DELAY_TIME_MS 500
 
 typedef struct counters {
   uint8_t counter_a;
   uint8_t counter_b;
 } Counters;
 
-void increment_callback(SoftTimerId timer_id, void *context) {
+static void prv_increment_callback(SoftTimerId timer_id, void *context) {
   Counters *counter_data = context;
-  counter_data->counter_a += 1;
+  counter_data->counter_a++;
   LOG_DEBUG("Counter A: %d\n", counter_data->counter_a);
 
-  if (counter_data->counter_a % 2 == 0 && counter_data->counter_a != 0) {
-    counter_data->counter_b += 1;
+  if (counter_data->counter_a % 2 == 0) {
+    counter_data->counter_b++;
     LOG_DEBUG("Counter B: %d\n", counter_data->counter_b);
   }
-  soft_timer_start_millis(PERIOD, increment_callback, context, NULL);
+  soft_timer_start_millis(DELAY_TIME_MS, prv_increment_callback, counter_data, NULL);
 }
 
 int main() {
   interrupt_init();
   soft_timer_init();
 
-  Counters counters = { 0, 0 };
-  soft_timer_start_millis(PERIOD, increment_callback, &counters, NULL);
+  Counters counters = { 0 };
+  soft_timer_start_millis(DELAY_TIME_MS, prv_increment_callback, &counters, NULL);
 
   while (true) {
     wait();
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -242,7 +242,7 @@ lint_quick:
 
 # Globally disable the following pylint messages:
 PYLINT_DISABLE := \
-	import-error redefined-outer-name unused-argument \
+	consider-using-with import-error redefined-outer-name unused-argument \
```
AveryChiu64, line 245: hmm, this change shouldn't be part of the hwl

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
+// Firmware 102 Homework
```
AveryChiu64, line 1: I think putting your name and labeling it is nice. But for actual projects, we normally don't sign and date them.

I know companies like Google and Qualcomm will keep a log of changes for all their code in the source files for certain projects, but at Midnight Sun we don't do this.
Maybe we could in the future, but for now this isn't part of our coding standards

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+// Vaaranan Yogalingam
+// 2021-05-29
+
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+#define PERIOD 500
```
AveryChiu64, line 10: I'm not sure I like this naming here. Though it makes sense to me (since I know what the tasks of this hwk assignment is) it would be unclear to someone else.

What is the units for PERIOD? What is this PERIOD for? It would be best to change the naming so that it is more descriptive.

Normally in our code, we will add a suffix to the variable to describe the units eg:
DELAY_TIME_MS

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+
+typedef struct counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void increment_callback(SoftTimerId timer_id, void *context) {
```
AveryChiu64, line 17: We should add a static keyword and add a 'prv_' prefix to the name of the function to indicate that this is a private function which should only be used by this source file

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+  Counters *counter_data = context;
+  counter_data->counter_a += 1;
```
AveryChiu64, line 19: We can use `counter_data->counter_a++; ` here

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+  LOG_DEBUG("Counter A: %d\n", counter_data->counter_a);
+
+  if (counter_data->counter_a % 2 == 0 && counter_data->counter_a != 0) {
```
AveryChiu64, line 22: `counter_data->counter_a != 0`
is not needed since counter_a will never be 0. You increment counter_a before this line so this check is not needed.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+    counter_data->counter_b += 1;
```
AveryChiu64, line 23: We can just use `counter_b++`  here

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+    LOG_DEBUG("Counter B: %d\n", counter_data->counter_b);
+  }
+  soft_timer_start_millis(PERIOD, increment_callback, context, NULL);
```
AveryChiu64, line 26: I would personally pass in `counter_data` to make it clear what I'm passing in, but passing in `context` is fine as well.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counters = { 0, 0 };
```
AveryChiu64, line 33: `Counters storage = { 0 };` will make all elements 0 so you don't need to put 0 twice

AveryChiu64 (commented):
This is a good start. I've added some comments to help you improve the code

Makefile
vaaranan-y: Hi Avery, I put this in because I was having trouble with pylint in pushing my code. It is a temporary solution suggested by Mitchell, but I am looking into a more permanent solution. Other than that, I have applied all your suggestions. Thank you for the feedback! If there is anything else I should know or do, please let me know.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+#define DELAY_TIME_MS 500
+
+typedef struct counters {
```
AveryChiu64, line 8: In `typedef struct counters`, `counters` should be capitalized

vaaranan-y: Thanks Avery, I have fixed this as well.

AveryChiu64:
Looks good, please close this PR. You can move onto firmware 103 when you're ready.

PR #387, opened 2021-05-29
fw 102 hm

ducquy2200:


projects/hello_world/README.md
```diff
@@ -0,0 +1,16 @@
...
+        - What problem does it solve?
+        - How do I use it? (with usage examples / example commands, etc)
+        - How does it work? (architectural overview)
+    A README for a board project (powering a hardware board, e.g. power distribution, centre console, charger, BMS carrier) should answer the following questions:
+        - What is the purpose of the board?
+        - What are all the things that the firmware needs to do?
+        - How does it fit into the overall system?
+        - How does it work? (architectural overview, e.g. what each module's purpose is or how data flows through the firmware)
+-->
+# hello_world
```
AveryChiu64, line 15: hello_world should not be included with the firmware 102 homework

We normally want to keep each project on separate branches (or have multiple branches for a single project if its a big project) so that we don't have gigantic pull requests.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+#define HALF_SEC 500
```
AveryChiu64, line 7: I'm not sure I like this naming. Though it is clear what this is, someone reading this might get confused since they see a 500 beside HALF_SEC

Normally what we would do is name it something like DELAY_TIME_MS to indicate the units of the value and also to describe what this is used for

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_counter_increment(SoftTimerId id, void *context) {
```
AveryChiu64, line 14: This should be static so that it is localized to this source file

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
+#include <stdint.h>
+#include <stdlib.h>
```
AveryChiu64, line 2: I don't think we need stdlib here? Though I may be wrong
Try removing it and see if it still runs properly

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_counter_increment(SoftTimerId id, void *context) {
+  Counters *counters = context;
+
+  counters->counter_a += 1;
```
AveryChiu64, line 17: I sort of prefer counter_a++ but functionally it does the same thing in this code

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+  LOG_DEBUG("Counter A: %i\n", counters->counter_a);
+
+  if (counters->counter_a % 2 == 0) {
+    counters->counter_b += 1;
```
AveryChiu64, line 21: See comment about counter_a++

AveryChiu64 (commented):
Looks good, I've added some comments

PR #388, opened 2021-05-31
Soft 999 ben lin fw 102

benlin1221:
This is my homework for the firmware 102 session


projects/getting_started/src/main.c
```diff
@@ -0,0 +1,28 @@
+// Turns the LED on and sits forever
```
AveryChiu64, line 1: This this project should not be part of this pull request. We normally keep our work on seperate branches so when we make a PR we know exactly what is on that branch and what ticket it relates to.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,47 @@
...
+#include "wait.h"        // for wait function
+
+#define A_INC_PERIOD_MS 500  // milliseconds between a-increment
+
+typedef struct Counters {
+  uint16_t counter_a;
+  uint16_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
AveryChiu64, line 14: This should be static to localize it to this source file

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,47 @@
...
+
+  Counters storage = { 0 };  // we use this to initialize a struct to be all 0
+
+  soft_timer_start_millis(A_INC_PERIOD_MS,     // timer duration
+                          prv_timer_callback,  // function to call after timer
+                          &storage,            // automatically gets cast to void*
+                          NULL);               // timer id - not needed here
+
+  while (true) {
+    wait();  // waits until an interrupt is triggered rather than endlessly spinning
```
AveryChiu64, line 43: I really like how you have comments in many of these lines.
However, for actual projects, we generally don't add these comments unless it is needed to clarify things.
A lot of it is self-explanatory

eg I think the below is sort of obvious
```
  // start the timer again
  soft_timer_start_millis(A_INC_PERIOD_MS, prv_timer_callback, storage, NULL);
```

Again, I like how you've added a lot of comments for your learning. However, when it comes to actual projects, remeber that the more comments you add, the more the code must be maintained in case there are changes.
It can also clutter the code if there are too many comments that are self-explanatory.

projects/timer_challenge/src/main.c
```diff
@@ -0,0 +1,47 @@
+#include <stdint.h>      // for integer types
+#include "interrupt.h"   // interrupts are required for soft timers
+#include "log.h"         // for printing
+#include "soft_timer.h"  // for soft timers
+#include "wait.h"        // for wait function
+
+#define A_INC_PERIOD_MS 500  // milliseconds between a-increment
+
+typedef struct Counters {
+  uint16_t counter_a;
```
AveryChiu64, line 10: Any reason for using uint16_t instead of uint8_t?
Though this isn't a huge deal, we normally try to take up just enough space so that we don't cause something like a stack overflow.
Since this is on x86 it should be fine, however when you start writing code for the stm32s this may be more of an issue.

benlin1221: No real reason, I was just using the supplementary material as a template and didn't change the uint16_t counters. I will keep this in mind next time

PR #389, opened 2021-06-03
[SOFT-362] Removing ebrake from centre_console

PurpleRhythms:
Current issue:
drive_fsm not transitioning from neutral discharged state to set precharge despite this identical transition being made multiple times beforehand successfully. In fact, the drive_fsm does not transition at all after receiving a drive event. Event queue is not overflowing and neutral_discharge_state can only transition to set_precharge (in all four events drive, reverse, parking, and neutral). Neutral discharged is also the initial state and has no callback function which could complicate the state transition. Furthermore, the transition of neutral discharge to set precharge through a drive event was not modified during this ticket and should not have failed.

Description:
ebrake_tx.c alongside its header file & tests are deleted.

"Cascade deleted" instances of ebrake including ebrake states, enums, and events.

Restructured drive_fsm to remove set_ebrake state. Will be posted on Confluence once ticket has been resolved.

I noticed a possible enum numbering issue where multiple enums in centre_console_events.h would have the same numberation.


PurpleRhythms:
Here's a diagram of the new state machine: https://drive.google.com/file/d/1Gq2aHcBecG0SmXuGGIU8v_E8JwOMDRpg/view?usp=sharing

PurpleRhythms:
Drive_fsm information page!
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/2687074370/Drive+Finite+State+Machine+drive+fsm

ryandancy (approved):
Sorry for the super long wait for a review, it looks great!

PR #390, opened 2021-06-03
[Soft 483] Regen braking toggling

danielzyy:


projects/centre_console/src/regen_braking_toggle.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+static StatusCode prv_toggle_callback(const CanMessage *msg, void *context,
+                                      CanAckStatus *ack_reply) {
+  regen_braking_state = !regen_braking_state;
+  return prv_process_request(regen_braking_state);
+}
+
+StatusCode regen_braking_init() {
+  regen_braking_state = false;
```
ryandancy, line 39: IMO, just to enforce that this is the default state on power-up across the system, I think that we should try and send the set-regen-braking CAN message on init (i.e. call `prv_process_request(false)` here as well). That way everything is consistent even if one component in the system has a different default.

projects/centre_console/inc/regen_braking_toggle.h
```diff
@@ -0,0 +1,23 @@
...
+#include "can_ack.h"
+#include "can_msg_defs.h"
+#include "can_pack.h"
+#include "can_transmit.h"
+#include "event_queue.h"
+#include "status.h"
+
+StatusCode regen_braking_init();
+
+bool get_regen_braking_state();
```
ryandancy, line 19: Nitpicks: the parameter lists here should be `(void)` rather than `()`, same in the .c file too.

projects/centre_console/src/regen_braking_toggle.c
```diff
@@ -0,0 +1,63 @@
...
+StatusCode set_regen_braking(bool state) {
+  if (state != regen_braking_state) {
+    regen_braking_state = state;
+    return prv_process_request(regen_braking_state);
+  }
+  return STATUS_CODE_OK;
+}
+
+bool regen_braking_process_event(Event *e) {
+  if (e != NULL && e->id == POWER_MAIN_SEQUENCE_EVENT_COMPLETE) {
```
ryandancy, line 57: Maybe document this behaviour (that we set the regen braking state to true upon power main sequence complete) in the header file somewhere

ryandancy (changes requested):
Awesome job Daniel, I love the test cases! Sorry for taking so long to review - it looks great, just one thing to add that shouldn't take too long.

ryandancy (approved):

ryandancy:
Ah looks like CI caught some format stuff, could you address?

ryandancy:
Fixed a few merge conflicts that cropped up, also had to get rid of the `TELEMETRY` target on the regen braking CAN message because build_dbc.py was trying to generate an ACK message from telemetry in the DBC file. @AveryChiu64 should removing the `TELEMETRY` target cause any issues? (or is putting `TELEMETRY` as a target just documentation?)

AveryChiu64:
Removing the TELEMETRY target is fine. Telemetry will pick up all CAN messages regardless of the target

PR #391, opened 2021-06-03
Soft 999 megan park fw 103

meganpark01:


projects/103_hw_megan/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  uint16_t adc_data = 0;
+  adc_read_converted_pin(*adc_addr, &adc_data);
+
+  LOG_DEBUG("The ADC converted reading is: %d\n", adc_data);
+}
+
+int main() {
+  interrupt_init();
+  gpio_init();
+
```
AveryChiu64, line 20: `gpio_it` needs to be initialized as well since we are using interrupts with our GPIO pins

projects/103_hw_megan/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  // gpio pin 6
+  GpioAddress adc_addr = {
+    .port = GPIO_PORT_A,
+    .pin = 6,
+  };
+
+  GpioSettings adc_settings = {
```
AveryChiu64, line 27: We should keep the way we initialize these structs consistently.
It is best to explicitly state which setting are charging (similar to what you have in adc_addr)

```
eg
.direction = GPIO_DIR_IN
.state = ...
.resistor = ...
```

projects/103_hw_megan/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  adc_init(ADC_MODE_SINGLE);
+  adc_set_channel_pin(adc_addr, true);
+
+  // gpio button pin 2
+  GpioAddress button_addr = {
+    .port = GPIO_PORT_B,
+    .pin = 2,
+  };
+
+  GpioSettings button_settings = {
```
AveryChiu64, line 44: Same comment as above

projects/103_hw_megan/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+    GPIO_DIR_IN,
+    GPIO_STATE_LOW,
+    GPIO_RES_NONE,
+    GPIO_ALTFN_NONE,
+  };
+
+  gpio_init_pin(&button_addr, &button_settings);
+
+  InterruptSettings interrupt_settings = {
```
AveryChiu64, line 53: Same comment as above. Let's explicitly state what fields we are setting.

projects/103_hw_megan/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+    INTERRUPT_TYPE_INTERRUPT,
+    INTERRUPT_PRIORITY_NORMAL,
+  };
+
+  gpio_it_register_interrupt(&button_addr, &interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             prv_log_adc_reading, &adc_addr);
+
+  while (true) {
+    wait();
+  }
```
AveryChiu64, line 63: Hmm this is strange. Did the code compile properly?
I thought GCC would give you an error since you don't seem to have a return statement in main.
Even though technically this won't end since its stuck in a while true loop, we still need to add a 
`return 0` below it

AveryChiu64 (commented):
Great work! I've added a few comments.

AveryChiu64 (commented):
Looks good. You can close this PR

PR #392, opened 2021-06-04
[SOFT-496] Fixed tutorial board adc

AveryChiu64:
I noticed a lot of new members making the same mistakes over and over for fw 103 hwk and realized the tutorial_board_adc does not follow our coding standards and is outdated


mitchellostler (approved):
Yeah looks good, definitely will reduce confusion

ryandancy (approved):

PR #393, opened 2021-06-05
added interrupt to can_hw_receive

JarvisWeng:


Max-MZ (commented):
LGTM, as long as the main function of preventing polling is still there. If Ryan is okay with it I'm good.

JarvisWeng:
> LGTM, as long as the main function of preventing polling is still there. If Ryan is okay with it I'm good.

The `can_hw_receive` is called continually in the `prv_rx_handler` in the `ms-common/src/can.c`. It will be called until it RXes all the messages in the queue, in a given cycle. So it'll "wake up" the main thread for every message.
The `prv_rx_handler` in the `ms-common/src/can.c` is registered to be the call back of `CAN_HW_EVENT_MSG_RX`
```c
can_hw_register_callback(CAN_HW_EVENT_MSG_RX, prv_rx_handler, s_can_storage);
```
And in `can_hw.c`, the `CAN_HW_EVENT_MSG_RX` callback is called in `prv_rx_thread` in `can_hw.c` for non-loopback mode.
```c
if (filter_match && s_socket_data.handlers[CAN_HW_EVENT_MSG_RX].callback != NULL) {
      s_socket_data.handlers[CAN_HW_EVENT_MSG_RX].callback(
          s_socket_data.handlers[CAN_HW_EVENT_TX_READY].context);
    }
```
it's also just called at the end of the `can_hw_transmit` function as part of Jess's fix to unflake CAN tests.

ryandancy (approved):

PR #396, opened 2021-06-06
Soft 999 ben lin fw 103

benlin1221:


projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
```
AveryChiu64, line 6: I'm not sure we need soft timers here?

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+// reads from the specified address and logs the value
+static void prv_read_from_adc_addr(const GpioAddress *address, void *context) {
+  GpioAddress addr = {
+    .port = address->port,  //
+    .pin = address->pin     //
+  };
+  uint16_t adc_data = 0;
+  adc_read_converted_pin(addr, &adc_data);
+
+  LOG_DEBUG("ADC pin has value %d", adc_data);
```
AveryChiu64, line 18: Please add an \n at the end to make sure the logs are printed on new lines

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
```
AveryChiu64, line 23: No need for soft timer

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+  gpio_init();
+  gpio_it_init();
+
+  GpioAddress input_addr = { .port = GPIO_PORT_A, .pin = 6 };
```
AveryChiu64, line 27: Please add a comma beside 6
`GpioAddress input_addr = { .port = GPIO_PORT_A, .pin = 6, };`
Then once you run make format, you will get a similar format to your input_settings struct

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+    .state = GPIO_STATE_LOW,            //
+    .resistor = GPIO_RES_NONE,          //
+    .alt_function = GPIO_ALTFN_ANALOG,  //
+  };
+
+  gpio_init_pin(&input_addr, &input_settings);
+  adc_init(ADC_MODE_SINGLE);
+  adc_set_channel_pin(input_addr, true);
+
+  GpioAddress button_addr = { .port = GPIO_PORT_B, .pin = 2 };
```
AveryChiu64, line 40: Same comment for input_addr

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,63 @@
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// reads from the specified address and logs the value
+static void prv_read_from_adc_addr(const GpioAddress *address, void *context) {
+  GpioAddress addr = {
```
AveryChiu64, line 11: The GPIO address you passed in when you initialized the interrupt actually is passed in `void *context`
You will need to cast this to a `GpioAddress`

`const GpioAddress *address` is actually the address of the pin that triggers the interrupt

AveryChiu64 (commented):
Good work, ive added some comments.

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  gpio_init();
+  gpio_it_init();
+
+  GpioAddress input_addr = {
+    .port = GPIO_PORT_A,
+    .pin = 6,
+  };
+
+  GpioSettings input_settings = {
+    .direction = GPIO_DIR_IN,           //
```
AveryChiu64, line 28: Do we need the `//` at the ends? I think i've seen this before but it seems unneseccary

projects/adc_challenge/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+// reads from the specified address and logs the value
+static void prv_read_from_adc_addr(const GpioAddress *address, void *context) {
+  GpioAddress *addr = context;
+  uint16_t adc_data = 0;
+  adc_read_converted_pin(*addr, &adc_data);
```
AveryChiu64, line 12: This function will actually return a value of type `StatusCode`
If we don't get `STATUS_CODE_OK` we know we are getting corrupted data.
At the moment, we don't check for this. We could be printing out garbage without us knowing it.

Though what you have here is okay for the homework, for production code, we should almost always check the StatusCodes to make sure thing's dont fail (unless we are certain they won't)

AveryChiu64 (commented):
Looks good, just some minor comments for the future when you start working on projects.
You can close this PR.
Great work!

PR #399, opened 2021-06-08
Soft 461 assert solar relay

farazkh80:
Added the `prv_assert_relay` function, to check if the solar relay is open and the current is set and valid.

Also, added several test cases to check:

- If the assertion throws two `EE_SOLAR_RELAY_OPEN_ERROR` with `fault_data=0`  when the relay has not opened properly
- If the assertion throws one `EE_SOLAR_RELAY_OPEN_ERROR`  with `fault_data=current_value` when the current is more than the threshold
- If the assertion throws one `EE_SOLAR_RELAY_OPEN_ERROR` with `fault_data=0xFF` when the current is never set

Also, an extra check and two `rx_handlers` has been added to handle success `SYSTEM_CAN_MESSAGE_SOLAR_FAULT_6_MPPTS` and `SYSTEM_CAN_MESSAGE_SOLAR_FAULT_5_MPPTS` messages after each `relay_open`.


projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+
+static void prv_assert_relay(SoftTimerId timer_id, void *context) {
+  RelayFsmStorage *storage = (RelayFsmStorage *)context;
+  if (storage->isErrCalled == true) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0);
+    return;
+  }
+
+  int32_t data_value;
+  bool isSet = false;
```
ryandancy, line 28: This should be snake_case

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+
+  data_store_get_is_set(DATA_POINT_CURRENT, &isSet);
+
+  if (!isSet) {
+    if (storage->isSetCounter >= 30) {
+      LOG_DEBUG("Aborting, The current is not set\n");
+      fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0xFF);
+      return;
+    }
+    storage->isSetCounter += 1;
```
ryandancy, line 38: Nit: can just do `++` instead of `+= 1`.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+  data_store_get(DATA_POINT_CURRENT, (uint32_t *)&data_value);
+
+  if (data_value > CURRENT_ASSERT_THRESHOLD_uA) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, data_value);
+    return;
+  } else {
+    // Success message
+    if (solar_mppt_count == 6) {
+      CAN_TRANSMIT_RELAY_CURRENT_6_MPPTS();
+    } else if (solar_mppt_count == 5) {
```
ryandancy, line 52: Use `SOLAR_BOARD_5/6_MPPTS` from solar_boards.h instead of plain 5 and 6, that way e.g. if we add another one it's easier to figure out where to change.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+    storage->isSetCounter += 1;
+    soft_timer_start_millis(DATA_STORE_ASSERTION_DELAY_MS, prv_assert_relay, context, NULL);
+    return;
+  }
+
+  data_store_get(DATA_POINT_CURRENT, (uint32_t *)&data_value);
+
+  if (data_value > CURRENT_ASSERT_THRESHOLD_uA) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, data_value);
+    return;
```
ryandancy, line 47: Nit: this return is unnecessary

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
 #include "fault_handler.h"
 #include "fsm.h"
+#include "gpio.h"
 #include "log.h"
+#include "solar_boards.h"
 #include "solar_config.h"
 #include "solar_events.h"
 #include "status.h"
 
+static SolarMpptCount solar_mppt_count;
```
ryandancy, line 18: This should be prefixed with `s_` since it's static.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+
+static void prv_assert_relay(SoftTimerId timer_id, void *context) {
+  RelayFsmStorage *storage = (RelayFsmStorage *)context;
```
ryandancy, line 21: Nit: since `context` is a void pointer, you don't actually need the cast here.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+  if (storage->isErrCalled == true) {
```
ryandancy, line 22: [Boolean zen](https://courses.cs.washington.edu/courses/cse143/20sp/style/zen): you don't need the `== true`.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+    return;
+  }
+
+  int32_t data_value;
+  bool isSet = false;
+
+  data_store_get_is_set(DATA_POINT_CURRENT, &isSet);
+
+  if (!isSet) {
+    if (storage->isSetCounter >= 30) {
```
ryandancy, line 33: Factor the 30 out into a #define constant with a good name - usually we try not to leave these 'magic numbers' in code without an explanation.

projects/solar/inc/relay_fsm.h
```diff
@@ -8,14 +8,23 @@
...
 #include "event_queue.h"
 #include "fsm.h"
+#include "soft_timer.h"
+#include "solar_boards.h"
 #include "status.h"
 
+#define CURRENT_ASSERT_THRESHOLD_uA 10
+
+#define RELAY_ASSERTION_DELAY_MS 10
+#define DATA_STORE_ASSERTION_DELAY_MS 1000
```
ryandancy, line 18: You should add comments saying what each of these is (e.g. the current assert threshold one is the current threshold below which we consider the relay disconnected).

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+    }
+    storage->isSetCounter += 1;
+    soft_timer_start_millis(DATA_STORE_ASSERTION_DELAY_MS, prv_assert_relay, context, NULL);
+    return;
+  }
+
+  data_store_get(DATA_POINT_CURRENT, (uint32_t *)&data_value);
+
+  if (data_value > CURRENT_ASSERT_THRESHOLD_uA) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, data_value);
```
ryandancy, line 46: I'm not sure that your scheme of putting the current value in the fault data will work, because the fault data is a uint8 while the current is an int16, so it won't fit. I guess we could edit the solar fault CAN message to make the data field wider, but that could be a pain - I think it might be better to just signal that there was an overcurrent fault (and not specify what the current is).

To do this, add an enum in exported_enums.h beside the other solar ones and give it a good name with three values - one for "drv120 error", one for "current exceeded / relay not opened", and one for "current not set after timeout" (i.e. the >= 30 case above).

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+      return;
+    }
+    storage->isSetCounter += 1;
+    soft_timer_start_millis(DATA_STORE_ASSERTION_DELAY_MS, prv_assert_relay, context, NULL);
+    return;
+  }
+
+  data_store_get(DATA_POINT_CURRENT, (uint32_t *)&data_value);
+
+  if (data_value > CURRENT_ASSERT_THRESHOLD_uA) {
```
ryandancy, line 45: I think we should specify that the absolute value of the data_value has to be greater than the threshold (i.e. pass it through `abs`), just to account for the odd scenario where the current is going the wrong way.

projects/solar/src/relay_fsm.c
```diff
@@ -2,17 +2,64 @@
...
+    }
+  }
+}
+
 static void prv_relay_err_cb(void *context) {
   LOG_DEBUG("RELAY_ERROR CALLBACK\n");
-  fault_handler_raise_fault(EE_SOLAR_FAULT_DRV120, 0);
+  RelayFsmStorage *storage = (RelayFsmStorage *)context;
+  storage->isErrCalled = true;
+  fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0);
```
ryandancy, line 62: This one should still be `EE_SOLAR_FAULT_DRV120` right? Since it's not an error in opening the relay, it's an error with the drv120 normally.

projects/solar/test/test_relay_fsm.c
```diff
@@ -23,35 +24,62 @@
...
 static RelayFsmStorage s_storage;
 
-static bool relay_err_rx_cb_called;
-static StatusCode prv_err_rx_cb(const CanMessage *msg, void *context, CanAckStatus *ack_reply) {
-  relay_err_rx_cb_called = true;
+static bool relay_error_rx_cb_called;
+static bool relay_5_mppts_rx_cb_called;
+static bool relay_6_mppts_rx_cb_called;
+static uint8_t error_array[5];
+static int current_error_num;
```
ryandancy, line 48: Nit: these should all be prefixed with `s_` (and `current_error_num` shouldn't be `int` but a fixed-width type).

projects/solar/test/test_relay_fsm.c
```diff
@@ -23,35 +24,62 @@
...
+  error_array[current_error_num] = solar_fault;
+  current_error_num++;
+  return STATUS_CODE_OK;
+}
+static StatusCode prv_current_5_mppts_rx_cb(const CanMessage *msg, void *context,
+                                            CanAckStatus *ack_reply) {
+  relay_5_mppts_rx_cb_called = true;
+  return STATUS_CODE_OK;
+}
+static StatusCode prv_current_6_mppts_rx_cb(const CanMessage *msg, void *context,
```
ryandancy, line 64: Nit: there should be blank lines between the functions here.

projects/solar/test/test_relay_fsm.c
```diff
@@ -185,9 +217,67 @@ void test_invalid_input(void) {
...
+  TEST_ASSERT_TRUE(relay_fsm_process_event(&s_storage, &e));
+  MS_TEST_HELPER_CAN_TX_RX(SOLAR_CAN_EVENT_TX, SOLAR_CAN_EVENT_RX);
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+
+  TEST_ASSERT_TRUE(relay_error_rx_cb_called);
+  TEST_ASSERT_EQUAL(current_error_num, 1);
+  TEST_ASSERT_EQUAL(error_array[0], EE_SOLAR_RELAY_OPEN_ERROR);
+  MS_TEST_HELPER_ASSERT_NO_EVENT_RAISED();
+}
+// Test that the module handles EE_SOLAR_RELAY_OPEN_ERROR gracefully
```
ryandancy, line 265: Nit: there should be a blank line here

projects/solar/test/test_relay_fsm.c
```diff
@@ -23,35 +24,62 @@
...
+  return STATUS_CODE_OK;
+}
+static StatusCode prv_current_6_mppts_rx_cb(const CanMessage *msg, void *context,
+                                            CanAckStatus *ack_reply) {
+  relay_6_mppts_rx_cb_called = true;
   return STATUS_CODE_OK;
 }
 
 void setup_test(void) {
+  data_store_set(DATA_POINT_CURRENT, 5);
```
ryandancy, line 71: Using just '5' has a couple issues: it's hard to understand why it's being used just glancing at it, and if we ever change the current threshold (which is likely) then this might break. To fix this I'd set the current to `CURRENT_ASSERT_THRESHOLD_uA - 1`, which is dynamic and more obvious. (Though a comment doesn't hurt too!)

codegen/can_messages.asciipb
```diff
@@ -104,6 +104,28 @@ msg {
 
 # IDs: 11-17 Reserved
 
+msg {
+  id: 14
+  source: SOLAR_5_MPPTS
+  target: "TELEMETRY"
+  msg_name: "relay current 5 mppts"
```
ryandancy, line 111: A better name might be possible for these, since the message doesn't really have anything to do with the current - maybe "relay open ok 5 mppts" (and "6 mppts" for the other one)?

ryandancy (changes requested):
Nice job! Sorry for the super long wait for the review - it looks great! There's a lot of comments but they're mostly style nitpicks, nothing too major.

projects/solar/inc/relay_fsm.h
```diff
@@ -8,14 +8,23 @@
...
+
 typedef struct RelayFsmStorage {
   Fsm fsm;
+  bool isErrCalled;
+  uint32_t isSetCounter;
```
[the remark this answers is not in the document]
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: do I have to do like `(RelayFsmStorage *)context)->is_error_called` since just doing `context->is_error_called` throws me 
expression must have pointer-to-struct-or-union type but it has type "void *" error

projects/solar/src/relay_fsm.c
farazkh80: could you elaborate on this? the `drv120_error` and `relay not opened` errors already exist. so would I just add two more which would be `current exceeded and `current not set after timeout` errors but should I make new errors for all with some common word like:
`RELAY_ERROR_DRV120`
`RELAY_ERROR_RELAY_NOT_OPEN`
`RELAY_ERROR_CURRENT_EXCEDDED`
`RELAY_ERROR_CURRENT_NEVER_SET`

projects/solar/src/relay_fsm.c
ryandancy: No what I meant is you can just do `RelayFsmStorage *storage = context` without the cast.

projects/solar/src/relay_fsm.c
ryandancy: I meant another enum in exported_enums.h in addition to the `EESolarError` enum just for data values for `EE_SOLAR_FAULT_RELAY_OPEN_ERROR` faults. So you'd have an enum maybe called `EESolarRelayOpenErrorReason` with members like `EE_SOLAR_RELAY_ERROR_DRV120`, `EE_SOLAR_RELAY_ERROR_CURRENT_EXCEEDED_NOT_OPEN`, `EE_RELAY_ERROR_CURRENT_NEVER_SET` (and "relay not opened/current exceeded" should be one error since it's sent from the same place).

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/inc/relay_fsm.h
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/test/test_relay_fsm.c
farazkh80: done

projects/solar/test/test_relay_fsm.c
farazkh80: done

projects/solar/test/test_relay_fsm.c
farazkh80: done

projects/solar/test/test_relay_fsm.c
farazkh80: done

codegen/can_messages.asciipb
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

codegen/can_messages.asciipb
```diff
@@ -104,6 +104,28 @@ msg {
 
 # IDs: 11-17 Reserved
```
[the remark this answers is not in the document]
farazkh80: done

projects/solar/src/relay_fsm.c
```diff
@@ -28,35 +31,38 @@ static void prv_assert_relay(SoftTimerId timer_id, void *context) {
 
   data_store_get_is_set(DATA_POINT_CURRENT, &is_set);
 
+  // Checking if the current will ever be set
   if (!is_set) {
-    if (((RelayFsmStorage *)context)->is_set_counter >= MAX_NUMBER_OF_CURRENT_CHECKS) {
+    if (storage->is_set_counter >= MAX_NUMBER_OF_CURRENT_CHECKS) {
       LOG_DEBUG("Aborting, The current is not set\n");
-      fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0xFF);
+      fault_handler_raise_fault(EE_RELAY_ERROR_CURRENT_NEVER_SET, 0);
```
ryandancy, line 38: What I meant is that we should still have the `EE_SOLAR_RELAY_OPEN_ERROR` as the first argument here (the fault ID), but as the fault data (the second argunent) we should have `EE_RELAY_ERROR_CURRENT_NEVER_SET`.

projects/solar/src/relay_fsm.c
```diff
@@ -18,8 +18,11 @@
 static SolarMpptCount s_solar_mppt_count;
 
 static void prv_assert_relay(SoftTimerId timer_id, void *context) {
-  if (((RelayFsmStorage *)context)->is_error_called) {
-    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0);
+  RelayFsmStorage *storage = context;
+
+  // Checking if a relay open error has been raised
+  if (storage->is_error_called) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_ERROR_CURRENT_EXCEEDED_NOT_OPEN, 0);
```
ryandancy, line 25: Similarly here the first argument should be `EE_SOLAR_RELAY_OPEN_ERROR` and the second should be `EE_SOLAR_RELAY_ERROR_CURRENT_EXCEEDED_NOT_OPEN`.

projects/solar/src/relay_fsm.c
```diff
@@ -28,35 +31,38 @@ static void prv_assert_relay(SoftTimerId timer_id, void *context) {
...
     return;
   }
 
   data_store_get(DATA_POINT_CURRENT, (uint32_t *)&data_value);
 
-  if (data_value > CURRENT_ASSERT_THRESHOLD_uA) {
-    fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, data_value);
+  // Checking if the current is valid
+  if (abs(data_value) > CURRENT_ASSERT_THRESHOLD_uA) {
+    fault_handler_raise_fault(EE_SOLAR_RELAY_ERROR_CURRENT_EXCEEDED_NOT_OPEN, 0);
```
ryandancy, line 50: Also here, the first argument should be `EE_SOLAR_RELAY_OPEN_ERROR` and the second should be `EE_SOLAR_RELAY_ERROR_CURRENT_EXCEEDED_NOT_OPEN`.

projects/solar/src/relay_fsm.c
```diff
@@ -28,35 +31,38 @@ static void prv_assert_relay(SoftTimerId timer_id, void *context) {
...
   }
 }
 
 static void prv_relay_err_cb(void *context) {
   LOG_DEBUG("RELAY_ERROR CALLBACK\n");
-  ((RelayFsmStorage *)context)->is_error_called = true;
-  fault_handler_raise_fault(EE_SOLAR_RELAY_OPEN_ERROR, 0);
+  RelayFsmStorage *storage = context;
+  storage->is_error_called = true;
+  fault_handler_raise_fault(EE_SOLAR_RELAY_ERROR_DRV120, 0);
```
ryandancy, line 65: And here, the first argument should be `EE_SOLAR_RELAY_OPEN_ERROR` and the second argument should be `EE_SOLAR_RELAY_ERROR_DRV120`.

libraries/ms-helper/inc/exported_enums.h
```diff
@@ -200,11 +200,15 @@ typedef enum EESolarFault {
...
 
-  // The drv120 relay has signaled that overtemp/undervolt lockout conditions have been triggered
-  EE_SOLAR_FAULT_DRV120,
-
-  // The drv120 relay has not opened
-  EE_SOLAR_RELAY_OPEN_ERROR,
-
   NUM_EE_SOLAR_FAULTS,
 } EESolarFault;
+typedef enum EESolarRelayOpenErrorReason {
```
ryandancy, line 205: Nit: blank line between the enums please.

projects/solar/inc/relay_fsm.h
```diff
@@ -12,12 +12,21 @@
 #include "solar_boards.h"
 #include "status.h"
 
+// The relay is disconnected  if current is
+// more than current thershold,
```
ryandancy, line 16: It's actually disconnected if current is **less** than the current threshold right? Also typo: `thershold` should be `threshold`, and there's two spaces in the first comment line when there should be only one, and there's an extraneous comma at the end.

ryandancy (changes requested):
Looks good, just a couple tiny things to change (should take like 5 minutes)

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

projects/solar/src/relay_fsm.c
farazkh80: done

libraries/ms-helper/inc/exported_enums.h
farazkh80: done

projects/solar/inc/relay_fsm.h
farazkh80: done

ryandancy (approved):

ryandancy:
@farazkh80 Could you tell strategy/the UXD team about the various messages? Probably the easiest thing is to edit https://uwmidsun.atlassian.net/wiki/spaces/S/pages/2347598112/Errors+and+severities to mention the new errors + the relay-ok messages and how to parse them, then tell UXD about it in #strategy-uxd on Slack.

PR #400, opened 2021-06-10
Soft 999 vaaranan yogalingam fw 103

vaaranan-y:
Hello,

This is my submission for the second part of the firmware 103 homework. Please let me know if there is anything else I should do, but I will submit the first part of the homework as soon as possible. Thanks!

Vaaranan


projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,59 @@
+#include <stdint.h>
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+
+// Addresses
+static const GpioAddress button_address = { .port = GPIO_PORT_B, .pin = 2 };
```
AveryChiu64, line 9: I don't think you need all these global variables. You can move them into main

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  gpio_init();
+  gpio_it_init();
+
+  gpio_init_pin(&button_address, &button_settings);
+  gpio_init_pin(&adc_address, &adc_settings);
+  adc_init(ADC_MODE_SINGLE);
+
+  AdcChannel adc_channel = NUM_ADC_CHANNELS;
+
+  adc_get_channel(adc_address, &adc_channel);
```
AveryChiu64, line 48: Using channels is outdated, though this looks okay.
I would recommend looking at `adc.h` again. There is a better way to do this so you don't have to manually set the channels.

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  adc_set_channel(adc_channel, true);
+  adc_set_channel(ADC_CHANNEL_TEMP, true);
+
+  gpio_it_register_interrupt(&button_address, &interrupt_settings, INTERRUPT_EDGE_RISING,
+                             prv_adc_callback, &adc_address);
```
AveryChiu64, line 53: You're passing in `&adc_address` as context, but in the callback function, it looks like when you dereference the void ptr you're receiving the channel.

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+
+  while (true) {
```
AveryChiu64, line 55: It would be good to add a `wait();` inside the loop to put the processor to sleep. It will wake up when there is an interrupt

vaaranan-y: Hi Avery, I have made all the changes you suggested. Please let me know if there is anything else I should do. Thanks!

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,64 @@
+#include <stdint.h>
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+// Callback function
```
AveryChiu64, line 9: Though these comments are okay for the firmware homework. We normally wouldn't put such comments since the name of the function or variable itself is usually self-explanatory (eg we know main() is Main). 

You can leave these though since this is just the homework.

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Main
+int main() {
+  interrupt_init();
+  gpio_init();
+  gpio_it_init();
+
+  // Addresses
+  const GpioAddress button_address = { .port = GPIO_PORT_B, .pin = 2 };
```
AveryChiu64, line 25: For formatting, we normally put a comma at the end
(eg .pin = 2,) instead of just (.pin = 2)
Please fix this for all the structs

projects/button_challenge/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+                                           .priority = INTERRUPT_PRIORITY_NORMAL };
+
+  // Initializations
+  gpio_init_pin(&button_address, &button_settings);
+  gpio_init_pin(&adc_address, &adc_settings);
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(adc_address, true);
+
+  // Outdated ADC reading
```
AveryChiu64, line 50: Normally, for production code we would probably remove this.
Since this is for learning its fine to keep here

vaaranan-y: Hi Avery, I applied the latest changes, please let me know if there is anything else I should do, thanks!

AveryChiu64 (commented):
This looks good, you can close this PR

PR #402, opened 2021-06-15
soft_999_logan_yu_fw_102

123Logan321:
homework check for 102, thank you very much : )


projects/blinking_led/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+#include "wait.h"        // for wait function
+
+#define COUNTER_PERIOD_MS 500  // the time between each count in ms
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void SoftTimerCallCounter(SoftTimerId timer_id, void *context) {
```
AveryChiu64, line 13: The function should be static and also needs a prv in front of it as such
image

Our coding standards are here:
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards

projects/blinking_led/src/main.c
```diff
@@ -0,0 +1,37 @@
...
+  Counters *storage = context;
+  if ((storage->counter_a + 1) % 2 != 0 || storage->counter_a == 0) {
```
AveryChiu64, line 15: I think this would be better if you just checked and incremented counter b
It seems counter_a will increment regardless, so it would probably be better just to put it outside the if-else statement

projects/blinking_led/src/main.c
```diff
@@ -0,0 +1,34 @@
+#include "interrupt.h"   // apparently soft timers needs this
```
AveryChiu64, line 1: These comments are okay for your learning and for this homework.
Normally we wouldn't put these comments since they are sort of self explanatory, so for actual projects I would probably not put these sorts of comments in.

AveryChiu64 (commented):
Looks good, You can close this PR

projects/blinking_led/src/main.c
123Logan321: alright, thank you very much for the help

PR #405, opened 2021-06-30
SOFT 431 Python CAN Datagram Library

Cookieth:
Essentially the python library/script version of: https://github.com/uw-midsun/firmware_xiv/pull/375
Usage can be seen in test_can_datagram.

Brief overview of usage:

0. Import all classes:
```
import can
from can_datagram import Datagram
from can_datagram import DatagramSender
from can_datagram import DatagramListener
```

1. Create a message:
```
message = Datagram(
            protocol_version=TEST_PROTOCOL_VERSION,
            datagram_type_id=TEST_DATAGRAM_TYPE_ID,
            node_ids=TEST_NODES,
            data=bytearray(TEST_DATA))
```

2. Modify the message (optional)
```
message.set_protocol_version(protocol_version)
message.set_datagram_type_id(datagram_type_id)
message.set_node_ids(test_nodes)
message.set_data(bytearray(test_data))
```

3. Serialize the message into a bytearray (optional)
```
serialized = message.serialize()
```

4. Create a message sender
```
sender = DatagramSender(channel=TEST_CHANNEL, receive_own_messages=True)
```

5. Create a notifier
a) Create a callback
```
def triggerCallback(msg):
        # do stuff
        # note, you can interpret a bytearray of input using message.deserialize(BYTEARRAY).
```
b) Create a listener with the callback
```
listener = DatagramListener(triggerCallback)
```
c) Create the notifier and supply the sender's bus
```
notifier = can.Notifier(sender.get_bus(), [listener])
```

6. Send the message
```
sender.send(message)
```

Note: I believe the code to receive messages and call the callback is definitely suboptimal, but I couldn't think of any other way to do it. Pylint needs to be disabled for this section specifically because of how convoluted and nested it is. Let me know your thoughts. I took the approach of a finite state machine, since each call to the function only has information really about itself, and the call structure is very odd.

Also noting: I have no idea how that mu stuff got there, I'll try to figure out what's going on with that.


projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,343 @@
...
+            self.receiving_datagram = False
+            self.num_node_id = -1
+            self.incomplete_data_bytes = False
+            self.num_data_bytes = -1
+
+            # Here we want to handle what happens when we get the first message in a datagram
+            first_bytearray = msg.data
+
+            # Get the number of node ids
+            self.num_node_id = first_bytearray[HEADER_SIZE]
```
ryandancy, line 246: Mitchell's implementation sends an empty start message at the beginning to simplify things... I don't think you can assume that every message has 8 bytes. I haven't studied it in detail, but I think your logic in this method might be a bit overcomplicated by trying to deal with messages individually. Really your goal in this method is to answer the question "is there another message in this datagram?", which I think you might be able to do like this:
* every time you get a message (whether it's a start message or a non-start message), add its bytes to a list (I think it's `datagram_messages` here)
* to check if you've received all the messages, check if you have enough bytes to parse the number of nodes, parse the number of nodes if you do, then check if you have enough bytes to parse the data size, parse it if you do, and then check if you've got all the data based on that

(so really the goal is to get enough bytes to extract all the fields that determine the size of the datagram, and leave the rest of the parsing to `deseralize`)

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,343 @@
...
+                    self.num_data_bytes = self.num_data_bytes - bytes_remaining
+                else:
+                    # There are no more data bytes left, message is complete!
+                    self.receiving_datagram = False
+            for byte in msg.data:
+                self.datagram_messages.append(byte)
+
+            # Once we're on the last message, call the callback
+            if not self.receiving_datagram:
+                self.callback(self.datagram_messages)
```
ryandancy, line 330: IMO you should deserialize the bytearray before passing it into the callback so clients don't have to do that - not sure how you'd want to deal with notifying the user of deserializing errors then, maybe have an optional error callback?

ryandancy (changes requested):
I haven't looked at the tests, but great job! Looks mostly correct, only one or two larger-scale changes and the rest are mostly Python things or API critiques.

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,343 @@
...
+            for msg in can_messages:
+                self.bus.send(msg)
+            print("{} messages were sent on {}".format(len(can_messages), self.bus.channel_info))
+
+        except can.CanError:
+            print("Message could not be sent")
+
+    def get_bus(self):
+        """Accessor for the bus"""
+        return self.bus
```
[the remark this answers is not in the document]
Cookieth: Should I remove all `get_x` functions then?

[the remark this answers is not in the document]
ryandancy: Sure yeah

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,363 @@
...
+        message_extended_arbitration = False
+
+        # Populate an array with the can message from the library
+        for chunk_message in chunk_messages:
+            can_messages.append(can.Message(arbitration_id=message_arbitration_id,
+                                            data=chunk_message,
+                                            is_extended_id=message_extended_arbitration))
+
+            # After the first message, set the arbitration ID to 0
+            message_arbitration_id = CAN_ARBITRATION_ID
```
ryandancy, line 214: One more thing - I think that Mitchell's C implementation assumes that the start message has no data, so it would probably be good to send an empty message with `CAN_START_ARBITRATION_ID` first then send all the messages with data with `CAN_ARBITRATION_ID`.

Cookieth: What is meant by no data? Does that mean no node id's as well? Is it just a CAN message where the data is all 0's?

ryandancy: It's a CAN message where there is no data, the DLC (data-length-code) is 0. (Nothing to do with the node IDs or anything with the structure of the datagram protocol, just a CAN message with no data. Like send an empty array instead of a list of 8 bytes of data.)

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,343 @@
...
+CAN_BITRATE = 500000
+
+MESSAGE_SIZE = 8
+HEADER_SIZE = 6
+
+
+class Datagram:
+    """This class acts as an easy modular interface for a datagram."""
+
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
Cookieth: I mainly wrote this in because of the list of errors I was working through to get pylint to work. However, now that I've added: `# pylint: skip-file`, I no longer need to worry about the errors. Would you rather I list out those parameters, ignoring pylint?

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,343 @@
...
+        self.incomplete_data_bytes = False
+        self.num_data_bytes = -1
+
+        super().__init__()
+
+    def on_message_received(self, msg):
+        """This (SHOULD) wait for the first message in the datagram 0x010."""
+        super().on_message_received(msg)
+
+        if(msg.arbitration_id == 0x0010 and not self.receiving_datagram):
```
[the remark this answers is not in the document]
Cookieth: I thought of perhaps an easier way of implementing this, utilizing the dictionary idea. I'm not sure if this is what you had in mind. Let me know if this is what you had in mind.

projects/bootloader/scripts/can_datagram.py
Cookieth: My newer implementation handles this in a different way now. I'll move the conversation to the other comment, and look for your feedback there.

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+
+        crc32_array = bytearray([self._datagram_type_id,
+                                 len(self._node_ids),
+                                 *node_crc32,
+                                 len(self._data) & 0xff,
+                                 (len(self._data) >> 8) & 0xff,
+                                 *data_crc32])
+        # Update the crc32
+        crc32 = zlib.crc32(crc32_array)
+        crc32 = self._convert_to_bytearray(crc32, 4)
```
Cookieth, line 100: Note to self: I forgot to move this crc32 logic to the function I made for it... I will do that.

ryandancy: Do this!

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+        return out_bytearray
+
+    def _convert_from_bytearray(self, in_bytearray, bytes):
+        """This is a helper function that converts a bytearray into a value"""
+        value = 0
+        for i in range(0, bytes):
+            value = value | ((in_bytearray[i] & 0xff) << (i * 8))
+        return value
+
+    def _calculate_crc32(self, datagram_type_id, node_ids, data):
```
ryandancy, line 194: Python point: you don't use `self` in any of these, and you call them like `cls._convert_to_bytearray(cls, ...)`. Instead you could decorate them with `@staticmethod`, so that you can omit the `self` argument and call them like `cls._convert_to_bytearray(...)`. That's more Pythonic.

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+        """This function describes the node ids property."""
+        return self._node_ids
+
+    @node_ids.setter
+    def node_ids(self, value):
+        """This function sets the node ids."""
+        assert isinstance(value, list)
+        for val in value:
+            assert val & 0xff == val
+            assert val < 64
```
ryandancy, line 147: Cool Python thing you can do: `assert all(val & 0xff == val and val < 64 for val in value)`. Also I think `0 <= val < 0xff` would be a nicer way to check those conditions (and Python allows you to chain conditionals like that).

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+        assert isinstance(kwargs["node_ids"], list)
+        assert isinstance(kwargs["data"], bytearray)
+
+        # Verify all inputs
+        assert kwargs["datagram_type_id"] & 0xff == kwargs["datagram_type_id"]
+
+    def _convert_to_bytearray(self, in_value, bytes):
+        """This is a helper function that creates a little-endian bytearray"""
+        out_bytearray = bytearray()
+        for i in range(0, bytes):
```
ryandancy, line 183: Just `range(bytes)` is equivalent to `range(0, bytes)`

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+                is_extended_id=message_extended_arbitration)]
+
+        # Populate an array with the can message from the library
+        for chunk_message in chunk_messages:
+            can_messages.append(can.Message(arbitration_id=message_arbitration_id,
+                                            data=chunk_message,
+                                            is_extended_id=message_extended_arbitration))
+
+            # After the first message, set the arbitration ID to 0
+            message_arbitration_id = CAN_ARBITRATION_ID
```
ryandancy, line 249: I don't think you need these lines anymore since `message_arbitration_id` never changes

projects/bootloader/scripts/can_datagram.py
```diff
@@ -0,0 +1,294 @@
...
+
+        board_id = (msg.arbitration_id & 0b11111100000) >> 5
+        start_message = (msg.arbitration_id & 0x10) >> 4
+
+        if start_message == 1:
+            # Reset the datagram message when receiving a start message
+            self.datagram_messages[board_id] = msg.data
+
+        if start_message == 0:
+            self.datagram_messages[board_id] += msg.data
```
ryandancy, line 285: One robustness point: this will error out if a non-start message is sent before any start messages are sent. This is possible if e.g. we start listening in the middle of a datagram. So you want to check that `board_id in self.datagram_messages` before doing this (or attempting to receive the datagram below) in order to just ignore "partial" datagrams.

projects/bootloader/scripts/test_can_datagram.py
```diff
@@ -0,0 +1,145 @@
...
+            node_ids=TEST_NODES,
+            data=bytearray(TEST_DATA))
+
+        self.assertEqual(message.serialize(), bytearray(
+            b'\x01m\xc2\x18\xfe\x01\n\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x1a\x00\x03\x01\x04\x01\x05\t\x02\x06\x05\x03\x05\x08\t\x07\t\x03\x02\x03\x08\x04\x06\x02\x06\x04\x03\x03'))
+
+    def test_deserialize(self):
+        """Test retrieving Datagram information from the bytearray"""
+        message = Datagram.deserialize(bytearray(
+            b'\x01m\xc2\x18\xfe\x01\n\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x1a\x00\x03\x01\x04\x01\x05\t\x02\x06\x05\x03\x05\x08\t\x07\t\x03\x02\x03\x08\x04\x06\x02\x06\x04\x03\x03'))
```
ryandancy, line 78: Did you generate these using this script? If so, it's probably better to generate them manually, to ensure that they're in adherence with the spec. You can define a bytearray (shared between the two tests) from an array of bytes like `bytearray([0x01, 0x20, ...])` and then add comments so someone can easily verify that it's in accordance with the spec. Then we can make sure that serialize/deserialize is strictly in accordance with the spec.

projects/bootloader/scripts/test_can_datagram.py
```diff
@@ -0,0 +1,145 @@
...
+            datagram_type_id=TEST_DATAGRAM_TYPE_ID,
+            node_ids=TEST_NODES,
+            data=bytearray(TEST_DATA))
+        sender.send(message)
+
+        timeout = time.time() + 10
+        while not self.callback_triggered:
+            if time.time() > timeout:
+                break
+            pass
```
ryandancy, line 134: You don't need this `pass`

ryandancy (changes requested):
Nice! Comments are mostly just python tips, shouldn't take overly long.

projects/bootloader/scripts/can_datagram.py
Cookieth: I will make this change, but this makes me very curious-- do you think you could explain to me on the side about pythonic class methods? e.g. when it is appropriate to use classmethod, staticmethod, and no decorator? Thanks!

Cookieth:
Hi @ryandancy I have made the changes requested in your latest comments. However, I am still waiting your input on kwargs (see: https://github.com/uw-midsun/firmware_xiv/pull/405#discussion_r661912204), and your suggestion on the approach there.

Further, should I assume that since there is no further comment on this comment: https://github.com/uw-midsun/firmware_xiv/pull/405#discussion_r661918171 , that I can consider it resolved?

projects/bootloader/scripts/can_datagram.py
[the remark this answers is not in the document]
ryandancy: Eh, it's probably fine for now - listing them out would probably be the best practice but don't sweat about it

projects/bootloader/scripts/can_datagram.py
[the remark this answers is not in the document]
ryandancy: Yep, looks good!

projects/bootloader/scripts/can_datagram.py
ryandancy: There's sort of two distinctions here: one between no decorator and using classmethod or staticmethod, and then one between classmethod and staticmethod.

Methods that belong to a particular instance of a class shouldn't be decorated with classmethod or staticmethod, and methods that belong to the class as a whole (not to any particular instance, just grouped with the class so you can call it with `MyClass.my_method`) should be decorated with classmethod / staticmethod. It's the same as the distinction between static and instance methods in C++/Java/C# if you're familiar with those.

Within 'static' methods, a method decorated with classmethod will receive a `cls` parameter as the first argument, like an instance method gets `self` as the first argument. `cls` is an object representing the class (everything is an object in Python, even classes), and in particular you can call `cls()` to construct an object of that class. It's like you did `cls = MyClass`. The main use for this is so that if you call a class method on a subclass, it'll automatically use the subclass in `cls` and you don't have to rewrite the class method. Classmethods are pretty much a Python-only thing, I haven't seen them elsewhere.

A method decorated with staticmethod, on the other hand, is very much like a C++/Java/C# `static` method. It doesn't receive any special parameters, and it's static so it can't access instance fields. It's basically a regular function except grouped with a class so you can do `MyClass.my_static_method()` to call it.

ryandancy (approved):
Looks great! Let me know if you want to change the kwargs thing, else I'll just merge this.

PR #408, opened 2021-07-02
[SOFT 421] Transmit motor temp

YiJie-Zhu:
Half-finished code for transmitting motor temperature through CAN.


hewittmcg (changes requested):
Thanks for taking the time to work on finishing this, Frank!  It looks good overall, and I like the idea of moving the temperature checking logic into its own module.  Some of my comments are minor issues -- the general important things are:

- Make sure to write unit tests so that we know the module is working as expected. These will likely be similar to the ones in `test_mci_broadcast.c`, so you can use that as a reference. 
- Change how the messages get received from the motor controllers so that the handling functions get called from `mci_broadcast.c`. 
- Figure out a way to differentiate between the different temperature readings when broadcasting over CAN so that we can easily process them when we receive a `SYSTEM_CAN_MESSAGE_MOTOR_TEMPS` message

I mentioned the possibility of moving things into `mci_broadcast.c`, but that's only if it makes life easier for you 😃

ryandancy (approved):
Looks good to me! Will merge once @hewittmcg approves

PR #410, opened 2021-07-02
Soft 999 darren choi fw 103

choiii-darren:


projects/firmware_103/src/main.c
```diff
@@ -0,0 +1,74 @@
+#include "../../../libraries/ms-common/inc/adc.h"
```
AveryChiu64, line 1: This is strange, you shouldn't need to do this for it to work?
`#include "adc.h"`
Why doesen't this work?

projects/firmware_103/src/main.c
```diff
@@ -0,0 +1,74 @@
+#include "../../../libraries/ms-common/inc/adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+typedef enum { B2_BUTTON = 0, NUM_BUTTONS } Button;
```
AveryChiu64, line 9: Add a comma after NUM_BUTTONS, so when you run `make format`, it will format itself correctly

AveryChiu64: Also this may be a bit overkill for an enum, we can just use `#define` in this case

projects/firmware_103/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+typedef enum { B2_BUTTON = 0, NUM_BUTTONS } Button;
+
+typedef enum { A6_SENSOR = 0, NUM_SENSORS } Sensor;
```
AveryChiu64, line 11: Same comments as above

projects/firmware_103/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+
+static GpioAddress button_addresses[] = {
+  [B2_BUTTON] = { .port = GPIO_PORT_B, .pin = 2 },
```
AveryChiu64, line 14: Add comma after .pin=2

projects/firmware_103/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+};
+
+static GpioAddress sensor_addresses[] = { [A6_SENSOR] = { .port = GPIO_PORT_A, .pin = 6 } };
```
AveryChiu64, line 17: Same as above

projects/firmware_103/src/main.c
AveryChiu64: You also probably don't need to make this an array. I like that it makes it scaleable, but in this case we don't really need it.

AveryChiu64:
Closing since this has been open for more than a month with no update. Please reopen this once/if changes are made

PR #411, opened 2021-07-03
[SOFT-503] Look into event queue overloading on PD

michael1ding:


ryandancy:
This is great - one other thing you might want to try is lengthening `CURRENT_MEASUREMENT_INTERVAL_US` in main.c. How PD measures stuff is every `CURRENT_MEASUREMENT_INTERVAL_US` microseconds, the current_measurement module will measure current from a bunch of places, then invoke `publish_data_publish` to TX all of it over CAN. So if `CURRENT_MEASUREMENT_INTERVAL_US` is too short, we could end up TXing more stuff than we want at a time.

Also if you double `ITERATION_WAIT_MS`, you should also halve `NUM_MSG_PER_ITERATION`, since that's the number of messages TXed at once - longer wait times between them will help, but we also want to avoid putting a lot of strain on the event queue at one time.

(Also feel free to adjust the parameters more - we want to be really sure that the event queue isn't getting overloaded, and having slower current measurement is way better than dropping events.)

ryandancy (approved):
Nice job, this should definitely help

PR #412, opened 2021-07-07
[SOFT-501] Uncommitted Protobuf Generated Code Check in CI

PurpleRhythms:


.github/workflows/main.yml
```diff
@@ -16,19 +16,38 @@ env:
...
+        id: cache-packages
+        with:
+          path: |
+            /usr/local/bin/protoc
+            /usr/local/bin/protoc-c
+            /usr/local/bin/protoc-gen-c
+            /usr/local/include/google/protobuf
+            /usr/local/include/google/protobuf-c
+            /usr/local/include/protobuf-c
+            /usr/local/lib
```
ryandancy, line 35: Try and rerun the CI jobs to see if the caching works; if it does, great, but if it doesn't, I worry that caching all of /usr/local/lib will cause issues. I assume there's a particular protobuf library in there that you want to cache, so to cache only that library I'd recommend copying it from /usr/local/lib to some path you can restore (like make ~/.cache/protobuf or something and copy it there) in the install-protobuf step, then copying it back in a step that runs if the cache successfully restored.

PurpleRhythms:
Note to self: If check cancels, try again in a couple hours.

ryandancy:
Note: in order to get the checks to run again you'll need to merge master (and deal with merge conflicts) because the github actions ubuntu-16.04 environment was removed and we've updated it to 18.04 on master.

PR #415, opened 2021-07-11
Soft 500 adding test coverage

raiyansayeed:


ryandancy:
Looks great! Could you just add `sudo apt-get install lcov` to `make install_requirements`? I had to install lcov separately to get it working.

ryandancy (approved):

PR #416, opened 2021-07-14
Soft 452 regen braking mci

farazkh80:
Implemented the regen braking on MCI side by 

- [ ] - Adding a regen braking state in `regen_braking.c`
- [ ] - Making test cases in `test_regen_braking.c` to ensure the functionality of `regen_braking.c`
- [ ] - Adding the internal regen_brake state to `mci_output.c` to set the current to zero when regen is diabled and the target velocity is less than the current velocity.
- [ ] - Added additional test cases to 'test_mci_output.c` to test all driving and pedal modes both when regen barking is enabled and disabled.


projects/mci/src/mci_output.c
```diff
@@ -59,6 +60,8 @@ static void prv_handle_drive(SoftTimerId timer_id, void *context) {
   MotorCanDriveCommand drive_command = { 0 };
   EEDriveOutput drive_state = drive_fsm_get_drive_state();
   bool is_cruise = drive_fsm_is_cruise();
+  uint8_t is_regen_brake = get_regen_braking_state();
```
AveryChiu64, line 63: Can we make this a `bool`? SImilar to `is_cruise`? Or is there something preventing us from doing this?

projects/mci/src/regen_braking.c
```diff
@@ -0,0 +1,27 @@
+#include "regen_braking.h"
+
+#include "can.h"
+#include "can_msg_defs.h"
+#include "can_unpack.h"
+#include "status.h"
+
+// regen braking state 1:ON - 0:OFF
```
AveryChiu64, line 8: Probably would be good just to define the states somewhere with an enum

projects/mci/src/regen_braking.c
```diff
@@ -0,0 +1,27 @@
...
+// Callback function to set the regen braking state
+static StatusCode prv_regen_braking_callback(const CanMessage *msg, void *context,
+                                             CanAckStatus *ack_reply) {
+  CAN_UNPACK_REGEN_BRAKING(msg, &s_regen_braking_state);
+  return STATUS_CODE_OK;
+}
+
+StatusCode regen_braking_init(void) {
+  // Default to enabled
+  s_regen_braking_state = 1;
```
AveryChiu64, line 20: Similar to above, it would be nice if we define the states somewhere so we don't have to add comments to make it clear

projects/mci/test/test_mci_output.c
```diff
@@ -114,6 +116,22 @@ static void prv_setup_motor_can(void) {
...
 
+static StatusCode prv_empty_ack_callback(CanMessageId msg_id, uint16_t device, CanAckStatus status,
+                                         uint16_t num_remaining, void *context) {
+  return STATUS_CODE_OK;
+}
+
+static StatusCode prv_set_regen_braking_state(uint8_t state) {
+  CanAckRequest req = { .callback = prv_empty_ack_callback,
+                        .context = NULL,
+                        .expected_bitset =
```
AveryChiu64, line 127: It looks like something is wrong with the formatting here.
My guess is to try and add a comma after `CAN_ACK_EXPECTED_DEVICES(SYSTEM_CAN_DEVICE_MOTOR_CONTROLLER)`

projects/mci/test/test_regen_braking.c
```diff
@@ -0,0 +1,46 @@
...
+  TEST_ASSERT_EQUAL(1, get_regen_braking_state());
+}
+
+void teardown_test(void) {}
+
+void test_regen_state_off(void) {
+  CanAckRequest req = { .callback = prv_empty_ack_callback,
+                        .context = NULL,
+                        .expected_bitset =
+                            CAN_ACK_EXPECTED_DEVICES(SYSTEM_CAN_DEVICE_MOTOR_CONTROLLER) };
```
AveryChiu64, line 32: Same formatting issue here it looks like

projects/mci/test/test_regen_braking.c
```diff
@@ -0,0 +1,46 @@
...
+  CAN_TRANSMIT_REGEN_BRAKING(&req, 0);
+  MS_TEST_HELPER_CAN_TX_RX(MCI_CAN_EVENT_TX, MCI_CAN_EVENT_RX);
+  TEST_ASSERT_EQUAL(0, get_regen_braking_state());
+}
+
+void test_regen_state_on(void) {
+  CanAckRequest req = { .callback = prv_empty_ack_callback,
+                        .context = NULL,
+                        .expected_bitset =
```
AveryChiu64, line 41: Formatting issue here as well, try adding a comma before the `};` which ends the struct

projects/mci/test/test_regen_braking.c
```diff
@@ -0,0 +1,46 @@
...
+                            CAN_ACK_EXPECTED_DEVICES(SYSTEM_CAN_DEVICE_MOTOR_CONTROLLER) };
+  CAN_TRANSMIT_REGEN_BRAKING(&req, 1);
```
AveryChiu64, line 43: What is 1 defined as here?

farazkh80: 1= REGEN_ON. I fixed it by defining an enum

projects/mci/test/test_regen_braking.c
farazkh80: done

projects/mci/test/test_regen_braking.c
farazkh80: done

projects/mci/test/test_mci_output.c
farazkh80: done

projects/mci/src/mci_output.c
farazkh80: I used the enum typedef for it which `RegenBrakingState`

projects/mci/src/regen_braking.c
farazkh80: done

projects/mci/src/regen_braking.c
farazkh80: done

projects/mci/inc/regen_braking.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
```
ryandancy, line 2: Could you include a comment here with a brief description of what the module does, and then what libraries it needs to be initialized? Like `// Requires x, y, and z to be initialized.`. All our head usually have that kind of documentation. (This one especially I think you should mention that this module is for managing whether or not we're currently regen braking, not for actually setting the current/velocity related to regen braking.)

projects/mci/src/mci_output.c
```diff
@@ -78,6 +81,13 @@ static void prv_handle_drive(SoftTimerId timer_id, void *context) {
     drive_command.motor_velocity =
         is_cruise ? cruise_rx_get_target_velocity() : s_velocity_lookup[drive_state];
   }
+
+  // Set current to zero if regen braking is disabled
+  // target velocity is less than actual velocity
+  if (is_regen_brake == 0 && drive_command.motor_velocity < s_velocity_lookup[drive_state]) {
```
ryandancy, line 87: 1) You should check `is_regen_brake == REGEN_ENABLED` (or `!is_regen_brake` if you change to `bool`), not 0
2) I don't think `s_velocity_lookup` is the way to find the actual velocity - it just tells us what target velocity we should set when pressing the throttle in the current drive state. @hewittmcg what would be the best way to find the actual current velocity? (Also setting current to 0 if target velocity < actual velocity is the correct way to disable regen braking right?)

projects/mci/src/mci_output.c
```diff
@@ -91,6 +101,7 @@ StatusCode mci_output_init(MotorControllerOutputStorage *storage, Mcp2515Storage
   };
   storage->motor_can = motor_can;
   status_ok_or_return(pedal_rx_init(&storage->pedal_storage, &pedal_settings));
+  status_ok_or_return(regen_braking_init());
```
ryandancy, line 104: IMO this should be initialized in main rather than here since the regen_braking module isn't really "owned" by the mci_output module - other modules could conceivably care about the regen braking state.

projects/mci/inc/regen_braking.h
```diff
@@ -0,0 +1,14 @@
...
+#include <stdbool.h>
+
+#include "can.h"
+#include "can_msg_defs.h"
+#include "can_unpack.h"
+#include "status.h"
+
+typedef enum { REGEN_DISABLED = 0, REGEN_ENABLED = 1 } RegenBrakingState;
```
ryandancy, line 10: Ah, centre console just uses a `bool` to represent "is regen braking enabled" - probably best to just use `bool` for consistency.

ryandancy (changes requested):
Code looks mostly good, the one important comment is the one about finding the actual velocity.

projects/mci/src/mci_output.c
hewittmcg: For current velocity, copy-pasting from my message to Faraz:
> We already periodically send the velocity to `cruise_rx.c `around line 95 of `mci_broadcast.c` using `cruise_rx_update_velocity`.  The easiest way to get the velocity in `mci_output.c` is probably to just track current velocity with a global and implement `mci_output_update_velocity` to periodically update it.  Then you can just call `mci_output_update_velocity `at the same point in `mci_broadcast.c` .

Also, yeah, `s_velocity_lookup` just maps to unobtainable values for forward/reverse and 0 for off, which lets us control the car based on current when not using cruise control.  Note that this means that when the car is going backwards, the target velocity should always be **less** than the actual velocity, which is an edge case that might need to be considered.

I read over the datasheet again and yeah, any instance where target < actual will cause regen braking if current is > 0. Because of this, I think we may want to avoid using cruise control with regen braking off, since the actual velocity could reach above the target velocity before we get a message and turn off current.  This would cause a bit of regen braking to still occur.

projects/mci/src/mci_output.c
farazkh80: Done. moved to main

projects/mci/inc/regen_braking.h
farazkh80: Done, changed to bool

projects/mci/inc/regen_braking.h
farazkh80: Done

projects/mci/src/mci_output.c
farazkh80: I mostly implemented these.

Two points that I was unsure about were:

1- How to update actual_velocity for test cases in `test_mci_output.c`. I directly called `mci_output_actual_velocity()` and set velocity for OFF: 0.0f, DRIVE: 10.0f and REVERSE: -10.0f

2- Copying from hewitt's comment:
>I read over the datasheet again and yeah, any instance where target < actual will cause regen braking if current is > 0. Because of this, I think we may want to avoid using cruise control with regen braking off, since the actual velocity could reach above the target velocity before we get a message and turn off current. This would cause a bit of regen braking to still occur.

Does that mean that the regen braking should be enabled(no change in current) when in cruise control? I don't think the tests in `test_mci_output.c` cover the cruise control on and off mode for testing, so in case I add this feature, should the test be made in  `test_cruise_rx.c` or `test_mci_output.c` ?

projects/mci/src/mci_output.c
```diff
@@ -91,6 +109,7 @@ StatusCode mci_output_init(MotorControllerOutputStorage *storage, Mcp2515Storage
   };
   storage->motor_can = motor_can;
   status_ok_or_return(pedal_rx_init(&storage->pedal_storage, &pedal_settings));
+  // status_ok_or_return(regen_braking_init());
```
ryandancy, line 112: We shouldn't leave commented-out code lying around, it's not gonna be needed in the future.

projects/mci/src/mci_output.c
```diff
@@ -23,6 +25,12 @@ static float s_velocity_lookup[] = {
 
 static float s_max_regen_in_throttle = 0.0f;
 static float s_regen_threshold = 0.0f;
+static float s_actual_velocity_ms = 0.0f;
+
+// Updateds currents_current_velocity_ms
```
ryandancy, line 30: Typo: `Updateds -> Updates`, and I think the variable is s_actual_velocity_ms lol

projects/mci/inc/regen_braking.h
```diff
@@ -1,14 +1,22 @@
 #pragma once
 
+// A module to set the regen braking state in MCI
+// upon receiving a CAN_TRANSMIT_REGEN_BRAKING can
+// message. The regen braking state is used in
+// mci_output.c which determines wether we are
```
ryandancy, line 6: Typo: `wether -> whether`

projects/mci/src/regen_braking.c
```diff
@@ -16,11 +16,11 @@ static StatusCode prv_regen_braking_callback(const CanMessage *msg, void *contex
 
 StatusCode regen_braking_init(void) {
   // Default to enabled
-  s_regen_braking_state = REGEN_ENABLED;
+  s_regen_braking_state = 1;
```
ryandancy, line 19: FYI you can use `true` and `false` instead of 0 and 1 if you have `<stdbool.h>` included

ryandancy (changes requested):
Looks good! Just a couple trivial typos, only blocking thing is the commented-out code which we shouldn't leave in. After that this can get merged!

projects/mci/src/regen_braking.c
farazkh80: done

projects/mci/inc/regen_braking.h
farazkh80: done

projects/mci/src/mci_output.c
farazkh80: done

projects/mci/src/mci_output.c
farazkh80: [done](url)

ryandancy (approved):

PR #417, opened 2021-07-15
[Soft-507] bootloader ping

ShiCheng-Lu:
There is only the response datagram right now. listening for the commands should be done by main.c along with all the other commands.


ShiCheng-Lu:
This PR also contains the dispatcher and datagram enums.

projects/bootloader/src/ping.c
```diff
@@ -0,0 +1,44 @@
...
+static uint8_t board_id;
+static uint8_t destination_node = 0;
+static CanDatagramTxConfig response_config = {
+  .dgram_type = BOOTLOADER_DATAGRAM_PING_RESPONSE,
+  .destination_nodes_len = 1,
+  .destination_nodes = &destination_node,
+  .data_len = 1,
+  .data = &board_id,
+  .tx_cb = prv_tx_callback,
+};
```
ryandancy, line 21: There's a little bit of a gotcha here. As I understand it (@mitchellostler correct me if I'm wrong), after `can_datagram` has TXed everything from the ping response message (which could take several iterations of the event loop), we need to call `can_datagram_start_listener` again so that we start listening for datagrams again.

A good place to do that would be in the TX completion callback. Since every bootloader operation will need to do this, I think the best thing to do would be to add a function to `dispatcher` which you can pass to `tx_cmpl_cb` here which just calls `can_datagram_start_listener`. Then when a module wants to send a datagram, it just has to pass that function as the `tx_cmpl_cb` and doesn't have to worry about calling `can_datagram_start_listener` itself.

ryandancy (changes requested):
Awesome job ShiCheng, great job getting this done so quickly! A few suggestions, none super major.

projects/bootloader/src/dispatcher.c
```diff
@@ -0,0 +1,61 @@
...
+  .rx_cmpl_cb = prv_dispatch,
+};
+
+// when a datagram rx is finished, the corresponding registered callback for that datagram id
+// is called with the datagram data and registered context
+static void prv_dispatch(void) {
+  BootloaderDatagramId id = datagram_rx.dgram_type;
+  if (id < NUM_BOOTLOADER_DATAGRAMS && s_callback_map[id] != NULL) {
+    s_callback_map[id](s_data, datagram_rx.data_len, s_context_map[id]);
+  }
```
mitchellostler, line 34: Since this doesn't return a statuscode, might be good to add a LOG_ERROR letting us know that an invalid ID or function has been called

projects/bootloader/src/dispatcher.c
```diff
@@ -0,0 +1,61 @@
...
+
+// callbacks and context
+static DispatcherCallback s_callback_map[NUM_BOOTLOADER_DATAGRAMS] = { NULL };
+static void *s_context_map[NUM_BOOTLOADER_DATAGRAMS] = { NULL };
+
+// setup datagram rx config
+static void prv_dispatch(void);
+static uint8_t s_destination_nodes[MAX_DEST_NODES_SIZE_BYTES] = { 0 };
+static uint8_t s_data[MAX_DATA_SIZE_BYTES] = { 0 };
+static CanDatagramRxConfig datagram_rx = {
```
mitchellostler, line 21: Nit: should be prefixed with s_ to be inline with our coding standards

projects/bootloader/src/ping.c
mitchellostler: Yeah, both can_datagram_start_listener disables after completion of one message, then calls the callback. The implementation in the new patch set looks good. The only thing to consider is that if a TX is started then nothing can be rxed, but this looks like it shouldn't be an issue with the current changes

projects/bootloader/src/ping.c
```diff
@@ -0,0 +1,31 @@
...
+  .dgram_type = BOOTLOADER_DATAGRAM_PING_RESPONSE,
+  .destination_nodes_len = 0,  // client listens to all datagrams
+  .destination_nodes = NULL,
+  .data_len = 1,
+  .data = &s_board_id,
+  .tx_cb = bootloader_can_transmit,
+  .tx_cmpl_cb = tx_cmpl_cb,
+};
+
+static StatusCode prv_ping_response(uint8_t data[2048], uint16_t data_len, void *context) {
```
mitchellostler, line 23: Unless there is a benefit to using a formal array parameter here that I am missing, I would make this a uint8_t*

projects/bootloader/src/main.c
```diff
@@ -1,21 +1,74 @@
...
+  uint8_t board_id = blconfig.controller_board_id;
+
+  // can and datagram
+  bootloader_can_init(&s_can_storage, &s_can_settings, board_id);
+  can_datagram_init(&s_datagram_settings);
+
+  dispatcher_init(board_id);
+  ping_init(board_id);
+
   jump_to_application();
```
mitchellostler, line 59: Might be a good Idea to add a jump_to_application_init() function similar to ping_init, that can be added to the initial part of the main function and will call jump_to_application() from its own library, abstracting out the details from the main loop.

Would be good to have @ryandancy 's thoughts

mitchellostler (dismissed):
Looks good, just a few things to think about. Let me know if I've gotten anything wrong or if you have any questions

projects/bootloader/src/dispatcher.c
ShiCheng-Lu: will be fixed in the next commit

projects/bootloader/src/ping.c
ShiCheng-Lu: will be fixed in the next commit.

projects/bootloader/src/main.c
ryandancy: The `jump_to_application()` call will actually jump to the application, it won't return, so in order to actually process commands that call can't be there. Once the jump-to-application command module is complete it'll use the `jump_to_application` to actually do the jumping to application - it's probably good to take out the call right now so it's possible to test ping? It doesn't really matter though.

projects/bootloader/test/test_ping.c
```diff
@@ -0,0 +1,182 @@
...
+    .destination_nodes_len = 1,
+    .destination_nodes = &s_board_id,
+    .data_len = 0,
+    .data = NULL,
+    .tx_cmpl_cb = tx_cmpl_cb,
+  };
+  prv_send_tx_to_fifo(tx_config);
+
+  // this rx handler not working...
+  TEST_ASSERT_OK(can_register_rx_handler(s_board_id, prv_rx_to_fifo, NULL));
```
ryandancy, line 153: Ah, the issue is that we can't actually do `can_register_rx_handler` with datagram messages. Because datagram messages have a particular arbitration ID format that isn't the same as the one used by our regular CAN system (see https://uwmidsun.atlassian.net/l/c/76h1GfbH, "CAN ID allocation" section), we actually added a special case in the CAN library (can_fsm.c, lines 58-62) which pass bootloader messages through the `bootloader_can` module rather than the regular CAN RX handler system.

Currently there can only be one registered bootloader CAN message handler, which is unfortunate. Since you're already changing bootloader_can, you could switch it to allow multiple RX handlers to be registered, and then you could use `bootloader_can_register_handler` here instead.

(On a somewhat unrelated note, it's great you've got all this infrastructure for testing with datagrams! Since it's repeated in both test_ping.c and test_dispatcher.c, it would be great to factor it out into a separate ms-common library that supports testing with datagrams. We have a few test-only libraries in ms-common, like `test_helpers.h` and `ms_test_helpers.h`, so you could have a `can_datagram_test_helpers` library. But that's for another ticket, let's get this merged first before doing that.)

projects/bootloader/src/dispatcher.c
ShiCheng-Lu: what module is LOG_ERROR in? Is it "log.h" with LOG_WARN or LOG_DEBUG ?

projects/bootloader/src/main.c
ShiCheng-Lu: it doesn't matter for my testing, but if you need to test it on an actual board, it should probably be removed. I'll just register it to its datagram command with dispatcher, so it can still be called if it's needed. There should probably be a init function for that, but it's not in my ticket :P

projects/bootloader/test/test_ping.c
ShiCheng-Lu: bootloader can actually checks if the sender is the client_script_board (msg_id 0) so it won't work for listening to controller to client script datagrams. I've added a separate callback for debuging that doesnt' check for client scipt id. And a separate register debug hander.

projects/bootloader/test/test_dispatcher.c
```diff
@@ -0,0 +1,263 @@
...
+      send_tx = false;
+      tx_left--;
+    }
+    MS_TEST_HELPER_AWAIT_EVENT(e);
+    if (e.id == CAN_DATAGRAM_EVENT_TX) {
+      send_tx = true;
+    }
+    can_process_event(&e);
+    can_datagram_process_event(&e);
+  }
```
ryandancy, line 170: When you go to make the datagram testing library, you'll want to factor this out into its own function so we're not copy/pasting it in every test. But that's for the next ticket.

projects/bootloader/test/test_ping.c
```diff
@@ -0,0 +1,268 @@
...
+  TEST_ASSERT_EQUAL(s_board_id, rx_config.data[0]);
+}
+
+void test_ping_addressed_to_multiple(void) {
+  TEST_ASSERT_OK(ping_init(s_board_id));
+
+  CanDatagramTxConfig tx_config = {
+    .dgram_type = BOOTLOADER_DATAGRAM_PING_COMMAND,
+    .destination_nodes_len = 5,
+    .destination_nodes = &s_board_id,
```
ryandancy, line 211: This could cause a segfault eh? Since `&s_board_id` isn't a (pointer to an) array of length 5 - this could cause reading past the end of the `destination_nodes` array. Instead you should pass an actual array of length 5.

projects/bootloader/test/test_ping.c
```diff
@@ -0,0 +1,268 @@
...
+  can_datagram_process_event(&e);
+
+  while (can_datagram_get_status() == DATAGRAM_STATUS_ACTIVE) {
+    // listens from s_rx-fifo
+    TestCanDatagramMessage msg;
+    fifo_pop(&s_rx_fifo, &msg);
+    can_datagram_rx(msg.data, msg.len, msg.is_start_msg);
+    MS_TEST_HELPER_AWAIT_EVENT(e);
+    can_datagram_process_event(&e);
+  }
```
ryandancy, line 265: This stuff (for receiving a full datagram) should also be factored out into a function in the testing library in the next ticket.

projects/bootloader/test/test_ping.c
```diff
@@ -0,0 +1,268 @@
...
+  TEST_ASSERT_EQUAL(DATAGRAM_STATUS_RX_COMPLETE, can_datagram_get_status());
+  TEST_ASSERT_EQUAL(s_board_id, rx_config.data[0]);
```
ryandancy, line 267: Might also want to test that the received datagram's `dgram_type` is what you expect it to be (i.e. the ping response datagram type).

ryandancy (changes requested):
Just one comment about the possible read-past-end-of-array in the tests, then we should be good!

projects/bootloader/test/test_dispatcher.c
ShiCheng-Lu: yep, the read into fifo, send from fifo, rx into fifo, and rx from fifo will all be factored out a seperate functions.

projects/bootloader/test/test_ping.c
ShiCheng-Lu: ah that should have been a 1, idk how it became a 5

ryandancy (approved):

PR #419, opened 2021-07-16
[SOFT-275] MCI Fan Control

hewittmcg:
Fan control module for MCI.  The way this works is pretty simple -- there's a single pin, `MCI_FAN_EN_ADDR`, that is used to turn on/off the fans.  This is abstracted by `mci_fan_set_state()`.  Hardware wants the fan state based on the temperature measurements from the motor controllers, so that's for a future ticket.

There are also four pins that go high when the corresponding thermistor on the MCI board hits a certain threshold.  Eventually we should log these over CAN using `SYSTEM_CAN_MESSAGE_MOTOR_STATUS`, but for now we just track their status in a fault bitset that can be exported using `mci_fan_get_fault_bitset()`.  A fault callback can also be passed in on initialization, and will be called each time the fault bitset updates.

The test sequence for this is pretty short, as the module itself is quite simple.

### DIFF
--- a/projects/mci/src/mci_fan_control.c
+++ b/projects/mci/src/mci_fan_control.c
@@ -2,7 +2,6 @@
 
 #include "gpio.h"
 #include "gpio_it.h"
-#include "log.h"
 
 static MciFanControlStorage storage;
 
@@ -17,8 +16,7 @@ static const GpioAddress s_therm_addrs[NUM_MCI_FAN_CONTROL_THERMS] = {
   [MCI_THERM_Q3_OVERTEMP] = MCI_Q3_OVERTEMP_ADDR,
 };
 
-// TODO(SOFT-275): figure out a more elegant way of doing this
-// (these need to be mapped somewhere in memory to pass them as context to the interrupt)
+// Store thermistor indices to pass as context to interrupt callback
 static const uint8_t s_therm_indices[NUM_MCI_FAN_CONTROL_THERMS] = {
   [MCI_THERM_DISCHARGE_OVERTEMP] = MCI_THERM_DISCHARGE_OVERTEMP,
   [MCI_THERM_PRECHARGE_OVERTEMP] = MCI_THERM_PRECHARGE_OVERTEMP,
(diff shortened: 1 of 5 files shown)


projects/mci/test/test_mci_fan_control.c
```diff
@@ -0,0 +1,175 @@
...
+static void prv_init_fan(void) {
+  MciFanControlSettings settings = {
+    .fault_cb = NULL,
+    .fault_context = NULL,
+  };
+  TEST_ASSERT_OK(mci_fan_control_init(&settings));
+}
+
+// Used to make sure the fault callback is working as expected
+static volatile uint16_t s_times_cb_called;
```
AveryChiu64, line 18: Though these static variables are automatically initialized to 0, I think we should still initialize them ourselves.
Also, does this need to be volatile? It seems like this variable is only used within this test and not modified elsewhere (ie we aren't reading from hardware registers where something else could modify the contents stored in this variable)

projects/mci/src/mci_fan_control.c
```diff
@@ -0,0 +1,96 @@
...
+  // Configure all thermistor pins
+  for (uint8_t i = 0; i < NUM_MCI_FAN_CONTROL_THERMS; i++) {
+    status_ok_or_return(prv_configure_therm_it(&g_therm_addrs[i], i));
+  }
+
+  return STATUS_CODE_OK;
+}
+
+StatusCode mci_fan_set_state(MciFanState state) {
+  GpioState gpio_state = (state == MCI_FAN_STATE_ON);
```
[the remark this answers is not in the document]
hewittmcg: Yeah, that's definitely clearer, thanks

projects/mci/test/test_mci_fan_control.c
hewittmcg: I think the volatile got left in when I was trying to debug a weird testing bug I introduced a while back, whoops, good catch, thanks

AveryChiu64 (approved):

ryandancy (approved):
Looks good, just wondering where `mci_fan_set_state` / `mci_fan_get_fault_bitset` will be called in the rest of MCI? It doesn't look like that's integrated yet.

hewittmcg:
> Looks good, just wondering where `mci_fan_set_state` / `mci_fan_get_fault_bitset` will be called in the rest of MCI? It doesn't look like that's integrated yet.

Thanks -- `mci_fan_set_state` will be called when we broadcast the WaveSculptor temperature measurements in `mci_broadcast.c` once that gets added in SOFT-421.  If any reading is above a certain threshold, which I believe is 65 C, then the fan should be on. I talked to Frank a few days ago, and he's down to wrap this into SOFT-421.

I was initially planning to handle broadcasting the fan fault bitset in MCI fault handling (SOFT-139).  I might end up spinning a separate ticket off that to improve `SYSTEM_CAN_MESSAGE_MOTOR_STATUS`, since it currently just rebroadcasts the lower 32 bits of each WaveSculptor's status message.  As such, `mci_fan_get_fault_bitset` will probably end up getting called somewhere inside `prv_broadcast_status` in `mci_broadcast.c`.

PR #421, opened 2021-07-21
added firmware 102 hw

JeffreyLin39:
Firmware 102 homework


projects/hello_world/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
```
AveryChiu64, line 1: Please remove the hello world project. This should not be included in the PR

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include "log.h"
+#include "soft_timer.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void prv_counter_callback(SoftTimerId soft_timer_id, void *context) {
+  Counters *counter_ptr = (Counters *)context;
```
AveryChiu64, line 11: You can just do `Counters *counter_ptr = context`

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+  uint8_t *counter_a_ptr = &(counter_ptr->counter_a);
```
AveryChiu64, line 12: This seems sort of unneseccary.
Since we already have counter_ptr, which points to our  counter we created in main, we don't need a pointer to each section of the counter.
We just access/manipulate the data in there by using `counter_ptr->counter_a = ...`

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+  uint8_t *counter_b_ptr = &(counter_ptr->counter_b);
+
+  if ((*counter_a_ptr == 0) || ((*counter_a_ptr) % 2 == 1)) {
```
AveryChiu64, line 15: Since counter A is increment either way, we can simply take it out of the if else statement

AveryChiu64: We can then just increment counter B if counter A is divisible by 2

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+    LOG_DEBUG("Counter A: %d\n", *counter_a_ptr);
+
+  } else {
+    (*counter_a_ptr)++;
+    (*counter_b_ptr)++;
+    LOG_DEBUG("Counter B: %d\n", *counter_b_ptr);
+    LOG_DEBUG("Counter A: %d\n", *counter_a_ptr);
+  }
+
+  soft_timer_start_millis(500, prv_counter_callback, counter_ptr, NULL);
```
AveryChiu64, line 26: What is 500? Though I know this is 500 miliseconds, this isn't exactly clear
Normally, we define these constants that we use using `#define` so these numbers aren't hardcoded
See this guide I wrote: https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1591803936/C+Reference+for+Midnight+Sun+Firmware+Team

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counter = { 0 };
+  soft_timer_start_millis(500, prv_counter_callback, &counter, NULL);
+
+  while (true) {
+  }
```
AveryChiu64, line 37: Lets add a `wait();` in the while loop to put the processor to sleep when its not doing anything and just waiting for interrupts

projects/counters/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+  soft_timer_start_millis(500, prv_counter_callback, counter_ptr, NULL);
+}
+
+int main() {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counter = { 0 };
+  soft_timer_start_millis(500, prv_counter_callback, &counter, NULL);
```
AveryChiu64, line 34: Same thing here, lets define 500 somewhere
eg `#define TIMER_INTERVAL_MS 500`
To define what 500 means and its units

projects/counters/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+static void prv_counter_callback(SoftTimerId soft_timer_id, void *context) {
+  Counters *counter_ptr = context;
+
+  if (counter_ptr->counter_a % 2 == 0 && counter_ptr->counter_a != 0) {
```
AveryChiu64, line 16: We don't need to check `counter_ptr->counter_a != 0` since we want both a and b to increment when they start.

AveryChiu64 (commented):
Just one small change and you're good.

AveryChiu64 (commented):
Nice work Jeffrey! You can close this PR now.

PR #423, opened 2021-07-24
[SOFT-513] Fix steering pins

ryandancy:
These are the correct pins according to Aashmika - the pins are mislabelled on the schematic. Also updated smoke_steering to be a bit more consistent with steering itself in terms of naming.

### DIFF
--- a/projects/smoke_steering/src/main.c
+++ b/projects/smoke_steering/src/main.c
@@ -9,17 +9,17 @@
 #include "wait.h"
 
 #define HORN_GPIO_ADDR \
-  { .port = GPIO_PORT_B, .pin = 1 }
-#define LANE_ASSIST_GPIO_ADDR \
-  { .port = GPIO_PORT_A, .pin = 6 }
-#define HIGH_BEAM_FORWARD_GPIO_ADDR \
-  { .port = GPIO_PORT_A, .pin = 7 }
-#define HIGH_BEAM_REAR_GPIO_ADDR \
-  { .port = GPIO_PORT_B, .pin = 0 }
-#define REGEN_BRAKE_TOGGLE_GPIO_ADDR \
   { .port = GPIO_PORT_A, .pin = 4 }
-#define CC_TOGGLE_GPIO_ADDR \
+#define LANE_ASSIST_GPIO_ADDR \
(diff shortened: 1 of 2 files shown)


projects/smoke_steering/src/main.c
```diff
@@ -36,9 +36,9 @@
 
 typedef enum {
   STEERING_DIGITAL_INPUT_HORN = 0,
-  STEERING_DIGITAL_INPUT_LANE_ASSIST,
-  STEERING_DIGITAL_INPUT_HIGH_BEAM_FORWARD,
-  STEERING_DIGITAL_INPUT_HIGH_BEAM_REAR,
+  STEERING_DIGITAL_INPUT_RADIO_PTT,
+  STEERING_DIGITAL_INPUT_DRL_1,
```
AveryChiu64, line 40: I might be mistaken but I was told that moving the stalk forward would turn on high beams and moving it back would turn it off. It seems there are two daytime running lights here?

projects/smoke_steering/src/main.c
```diff
@@ -9,17 +9,17 @@
 #include "wait.h"
 
 #define HORN_GPIO_ADDR \
+  { .port = GPIO_PORT_A, .pin = 4 }
+#define RADIO_PTT_GPIO_ADDR \
   { .port = GPIO_PORT_B, .pin = 1 }
-#define LANE_ASSIST_GPIO_ADDR \
```
AveryChiu64, line 13: This should be okay. I assume we won't be adding lane assist. Its just strange since it seems to show up in the board
image

AveryChiu64 (approved):

projects/smoke_steering/src/main.c
ryandancy: Lane assist is "radio push-to-talk" now I believe, it was already changed in steering but this is just to make it match in smoke_steering.

projects/smoke_steering/src/main.c
ryandancy: The new requirements are that moving the stalk forward turns on DRL and moving it back turns them off - this was actually written before those requirements came out. This naming is already present in steering and this is just to keep smoke_steering consistent.

PR #424, opened 2021-07-30
SOFT-511 datagram node 0 behaviours

ShiCheng-Lu:
if 0 is in a destination_nodes of a datagram, the datagram is received by all

if 0 is the node id, it receives all datagrams


ryandancy (approved):
Looks great!

PR #425, opened 2021-07-31
update steering

JarvisWeng:


projects/steering/src/steering_can.c
```diff
@@ -7,16 +8,18 @@
...
+static bool turning = false;
+
 StatusCode steering_can_process_event(Event *e) {
   switch (e->id) {
     case STEERING_INPUT_HORN_EVENT:
       CAN_TRANSMIT_HORN((EEHornState)e->data);
       break;
     case STEERING_DRL_1_EVENT:
-      CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_DRL, (EELightState)e->data);
+      if (turning == false) CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_DRL, EE_LIGHT_STATE_ON);
```
ryandancy, line 19: I think the idea is we want to disable left/right signals when one DRL input is on, rather than disabling DRL inputs when left or right is on. (Also you can just do `!turning` instead of `turning == false`.)

projects/steering/src/steering_can.c
```diff
@@ -7,16 +8,18 @@
...
 #include "steering_events.h"
 
+static bool turning = false;
+
 StatusCode steering_can_process_event(Event *e) {
   switch (e->id) {
     case STEERING_INPUT_HORN_EVENT:
       CAN_TRANSMIT_HORN((EEHornState)e->data);
       break;
     case STEERING_DRL_1_EVENT:
```
ryandancy, line 18: Could you rename `DRL_1` to `DRL_ON` and `DRL_2` to `DRL_OFF` throughout steering+smoke_steering since that's what their purpose is now?

projects/steering/src/steering_can.c
```diff
@@ -29,9 +32,19 @@ StatusCode steering_can_process_event(Event *e) {
       break;
     case STEERING_CONTROL_STALK_EVENT_LEFT_SIGNAL:
       CAN_TRANSMIT_LIGHTS(EE_LIGHT_TYPE_SIGNAL_LEFT, (EELightState)e->data);
+      if ((EELightState)e->data == EE_LIGHT_STATE_ON) {
+        turning = true;
+      } else {
+        turning = false;
+      }
```
ryandancy, line 39: Boolean zen tip: you can just do `turning = ((EELightState)e->data = EE_LIGHT_STATE_ON)`.

ryandancy (changes requested):
Just a couple points

projects/steering/src/steering_can.c
JarvisWeng: Oh my bad, I must have read that backwards

projects/steering/src/steering_can.c
JarvisWeng: Sure

projects/steering/src/steering_can.c
JarvisWeng: Oh yeah I thought about doing a turnery operator, but I think I was told against that somewhere.... I can change it

projects/steering/src/steering_can.c
JarvisWeng: Actually, I talked to @aashmikamali , and she said it's the opposite. It's if left/right signal is on, then we disable DRL.

ryandancy: Ah okay

ryandancy (approved):

PR #426, opened 2021-08-19
[SOFT-495] Migrated from protobuf-c to nanopb

Kyle02:


Makefile
```diff
@@ -296,7 +297,11 @@ pyformat:
 bootloader_protos:
 	@echo "Compiling protos..."
 	@mkdir -p $(BOOTLOADER_DIR)/protogen
-	@protoc -I=$(BOOTLOADER_DIR)/protos --c_out=$(BOOTLOADER_DIR)/protogen $(BOOTLOADER_DIR)/protos/*.proto
+	@for i in $$(ls $(BOOTLOADER_DIR)/protos); do \
+		protoc -I=$(BOOTLOADER_DIR)/protos -ocommand.pb $$i; \
+		python $(PYTHONPATHNANO)/generator/nanopb_generator.py -I=$(BOOTLOADER_DIR)/protos command.pb; \
+	done
+	@mv *.pb *.pb.c *.pb.h $(BOOTLOADER_DIR)/protogen
```
ryandancy, line 304: You should probably remove `command.pb` after this, I think it's an intermediate file?

projects/bootloader/protogen/flash_application_code.pb.c
```diff
@@ -0,0 +1,9 @@
+/* Automatically generated nanopb constant definitions */
```
ryandancy, line 1: I think the generated files got autoformatted by the commit hook before you added the exception to `FIND_MOD_NEW` - could you regenerate them to avoid the autoformatting? Otherwise it'll cause issues with the generated code checking that Daniel Zeng is doing.

ryandancy (changes requested):
Looks great, just a few nits

ryandancy (approved):

PR #427, opened 2021-08-20
soft_999_logan_yu_fw_103

123Logan321:
103 onboarding homework (I couldn't finish the part 1 scavenger hunt I apologize)


projects/Buttons_and_ADC/README.md
```diff
@@ -0,0 +1,16 @@
...
+        - What problem does it solve?
+        - How do I use it? (with usage examples / example commands, etc)
+        - How does it work? (architectural overview)
+    A README for a board project (powering a hardware board, e.g. power distribution, centre console, charger, BMS carrier) should answer the following questions:
+        - What is the purpose of the board?
+        - What are all the things that the firmware needs to do?
+        - How does it fit into the overall system?
+        - How does it work? (architectural overview, e.g. what each module's purpose is or how data flows through the firmware)
+-->
+# Buttons_and_ADC
```
AveryChiu64, line 15: This is a bit nitpicky, but if you look at all our other projects we use `camel_case` and that no letters are capitalized

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,64 @@
+#include "adc.h"
+#include "delay.h"
```
AveryChiu64, line 2: I don't think you need to use the delay library here? Should probably remove this

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+  // Button setting
+  GpioSettings button_setting = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .alt_function = GPIO_ALTFN_NONE,
+    .resistor = GPIO_RES_NONE,
+  };
+
+  // Adc address
+  GpioAddress adc_addresses = {
```
AveryChiu64, line 36: The naming of this variable might be confusing since this is just one GPIO address. Considering just using "address" rather than "addresses"

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+    .alt_function = GPIO_ALTFN_ANALOG,
+    .resistor = GPIO_RES_NONE,
+  };
+
+  static InterruptSettings s_interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_NORMAL,
+  };
+
+  gpio_init_pin(&adc_addresses, &adc_setting);
```
AveryChiu64, line 54: The GPIO pin for the button must be initialized as well. It looks like you created `button_settings` but never actually used it.

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+static void prv_button_interrupt_handler(const GpioAddress *adc_address, void *context) {
+  GpioAddress *adc_position = context;
+  uint16_t adc = 0;
+  adc_read_converted_pin(*adc_position, &adc);
```
AveryChiu64, line 12: This is also a bit nitpicky but for production code, we always want to check what the output of the functions we call so that we know the data we are receiving is correct before outputting the data. If its not correct, we should LOG_DEBUG an error message
Eg
`if (adc_read_converted_pin(*adc_position, &adc) == STATUS_CODE_OK) {`

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+static void prv_button_interrupt_handler(const GpioAddress *adc_address, void *context) {
+  GpioAddress *adc_position = context;
+  uint16_t adc = 0;
```
AveryChiu64, line 11: This should be named more descriptively. What is `adc`? Is it the data, the channel, the address? Right now its too ambigious

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+#include "wait.h"
+
+static void prv_button_interrupt_handler(const GpioAddress *adc_address, void *context) {
+  GpioAddress *adc_position = context;
+  uint16_t adc_data = 0;
+  StatusCode adc_converted_data = adc_read_converted_pin(*adc_position, &adc_data);
+  if (adc_converted_data == STATUS_CODE_OK) {
+    LOG_DEBUG("ADC reading is: %d\n", adc_data);
+  } 
+  else {
```
AveryChiu64, line 15: The else should be placed beside the end of the } as shown in our coding standards
https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/9306218/Coding+Standards
image

projects/Buttons_and_ADC/README.md
```diff
@@ -0,0 +1,16 @@
...
+        - What problem does it solve?
+        - How do I use it? (with usage examples / example commands, etc)
+        - How does it work? (architectural overview)
+    A README for a board project (powering a hardware board, e.g. power distribution, centre console, charger, BMS carrier) should answer the following questions:
+        - What is the purpose of the board?
+        - What are all the things that the firmware needs to do?
+        - How does it fit into the overall system?
+        - How does it work? (architectural overview, e.g. what each module's purpose is or how data flows through the firmware)
+-->
+# buttons_and_adc
```
AveryChiu64, line 15: Looks like the folder name still shows up with capital letters. This is fine this time around since it is a bit of a pain to have to recreate the project and do all that over again. Just remember it for next time.

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+    LOG_DEBUG("an error has occured.");
```
AveryChiu64, line 16: Need to add a \n at the end of this string

AveryChiu64 (commented):
Make sure you git hooks are set up, it looks like there is just a formatting issue and it causes the build to fail too. With the git hooks set up, formatting should be done automatically when you commit and push your changes.
image

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+static void prv_button_interrupt_handler(const GpioAddress *adc_address, void *context) {
+  GpioAddress *adc_position = context;
+  uint16_t adc_data = 0;
+  StatusCode adc_converted_data = adc_read_converted_pin(*adc_position, &adc_data);
```
AveryChiu64, line 11: This is good. However, I would not name this variable adc_converted_data since this returns a status code and not the actual data so may confuse whoever is looking at your code. Also you don't actually need to declare a separate variable for this. You can toss this directly in the if(...) statement

eg `if(adc_read_converted_pin(*adc_position, &adc_data) == STATUS_CODE_OK)`

projects/Buttons_and_ADC/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_NORMAL,
+  };
+
+  gpio_init_pin(&button_address, &button_setting);
+  gpio_init_pin(&adc_address, &adc_setting);
+  adc_init(ADC_MODE_SINGLE);
+  adc_set_channel_pin(adc_address, true);
+  gpio_it_register_interrupt(&button_address, &s_interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             prv_button_interrupt_handler, &adc_address);
```
AveryChiu64, line 62: There seems to be some formatting issue here?

PR #429, opened 2021-08-26
[SOFT-524] race normal switch CAN messages

raiyansayeed:


ryandancy:
Hey Raiyan, this looks great, but...

The original goal of this ticket was to *send* a CAN message when the race/normal mode changes, so telemetry can pick it up. Regardless, I think having a way to control the race/normal mode over CAN is valuable, so we should keep it - you'll just need to add a different CAN message which gets sent on each transition between race and normal mode for telemetry.

ryandancy (approved):
Looks good, thanks for the fix!

PR #430, opened 2021-08-27
[Soft 530] datagram testing module

ShiCheng-Lu:


libraries/ms-common/src/ms_test_helper_datagram.c
```diff
@@ -0,0 +1,132 @@
...
+    do {  // until another tx event is processed
+      MS_TEST_HELPER_AWAIT_EVENT(e);
+      can_datagram_process_event(&e);
+      can_process_event(&e);
+    } while (e.id != s_can_tx_event);
+  }
+  return STATUS_CODE_OK;
+}
+
+// mock recieve a datagram as the client
```
ryandancy, line 92: Typo: `recieve -> receive`

libraries/ms-common/inc/ms_test_helper_datagram.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+/* Test helpers for datagrams
+ */
```
ryandancy, line 3: We don't usually use C-style /* comments */ - could you replace it with a `//` comment and then add a `// Requires ... to be initialized` comment?

libraries/ms-common/inc/ms_test_helper_datagram.h
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+// initialize the datagram helper and dependencies
+void init_datagram_helper(CanStorage *can_storage, CanSettings *can_settings, uint8_t board_id,
+                          CanDatagramSettings *can_datagram_settings);
+
+// mock send the datagram |tx_config| from the client script, with datagram id of 0
+// this function only process events until all can messages are sent
+// which may leave some datagram rx events unprocessed,
+// use DATAGRAM_PROCESS_ALL to finish processing all events
```
ryandancy, line 24: I notice you don't always call `DATAGRAM_PROCESS_ALL` in the tests after calling `mock_tx_datagram` - when exactly does it need to be called? Would it be better to just call it at the end of `mock_tx_datagram`?

libraries/ms-common/inc/ms_test_helper_datagram.h
```diff
@@ -0,0 +1,29 @@
...
+
+#define DATAGRAM_PROCESS_ALL(e)                                 \
+  while (can_datagram_get_status() == DATAGRAM_STATUS_ACTIVE) { \
+    MS_TEST_HELPER_AWAIT_EVENT(e);                              \
+    can_datagram_process_event(&e);                             \
+    can_process_event(&e);                                      \
+  }
+
+// initialize the datagram helper and dependencies
+void init_datagram_helper(CanStorage *can_storage, CanSettings *can_settings, uint8_t board_id,
```
ryandancy, line 18: The usual naming convention would be `ms_test_helper_datagram_init` just to make it clear what module it's from.

libraries/ms-common/inc/ms_test_helper_datagram.h
```diff
@@ -0,0 +1,29 @@
...
+#include "can_datagram.h"
+#include "ms_test_helpers.h"
+#include "status.h"
+
+#define DATAGRAM_PROCESS_ALL(e)                                 \
+  while (can_datagram_get_status() == DATAGRAM_STATUS_ACTIVE) { \
+    MS_TEST_HELPER_AWAIT_EVENT(e);                              \
+    can_datagram_process_event(&e);                             \
+    can_process_event(&e);                                      \
+  }
```
ryandancy, line 15: Could you wrap this in either `do { ... } while (0)` or `({ })` (GNU extension)? That way things don't get weird with scope and semicolons. Also consider adding a documentation comment above it.

libraries/ms-common/inc/ms_test_helper_datagram.h
```diff
@@ -0,0 +1,29 @@
...
+                          CanDatagramSettings *can_datagram_settings);
+
+// mock send the datagram |tx_config| from the client script, with datagram id of 0
+// this function only process events until all can messages are sent
+// which may leave some datagram rx events unprocessed,
+// use DATAGRAM_PROCESS_ALL to finish processing all events
+StatusCode mock_tx_datagram(CanDatagramTxConfig *tx_config);
+
+// mock recieve a datagram to |rx_config| as the client script
+// recieves the next datagram sent through can
```
ryandancy, line 28: Typo: `recieve -> receive`

ryandancy (changes requested):
Awesome job ShiCheng! Mostly just a few typos and clarifications, looks great!

libraries/ms-common/inc/ms_test_helper_datagram.h
ShiCheng-Lu: if the `rx_cmpl_cb` from the mock datagram that was received starts another datagram transmission, the datagram never enters the `RX_COMPLETE` state, so the `DATAGRAM_PROCESS_ALL` will process the incoming transmission from the board as well. So it can't be called if the response should be received by the test helper

ryandancy (approved):

PR #431, opened 2021-09-04
SOFT-512, setup tests for mcp2515

michael1ding:


ryandancy:
Did you manage to verify that the coverage is better?

PR #433, opened 2021-09-09
[SOFT-514] Improve MCI status broadcast

hewittmcg:
This improves the MCI status broadcast so it sends more useful info.

In short, `SYSTEM_CAN_MESSAGE_MOTOR_STATUS` now sends each WaveSculptor's status/limit flags, as well as any board overtemperatures and overtemperatures from the WaveSculptor's temperature messages (to be added in SOFT-534).


ryandancy (approved):
LGTM! Make sure to update the Confluence page and also perhaps tell #uxd-telemetry-dash about the changes and what each flag means.

PR #435, opened 2021-09-11
Soft 508 bootloader jump to application

farazkh80:
https://uwmidsun.atlassian.net/browse/SOFT-508 
the only problem encountered is that when testing the config crc32 does not match the computed crc32. Not sure if it is because of not testing on the actual hardware or something else.


projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,54 @@
...
+void test_bootloader_application_crc32() {
+  BootloaderConfig config = { 0 };
+  config_get(&config);
+
+  // compute crc32 code
+  uint32_t computed_crc32 = calculated_application_crc32((uintptr_t)BOOTLOADER_APPLICATION_START,
+                                                         BOOTLOADER_APPLICATION_SIZE);
+
+  // fails, not sure if it is neccessary becuase we don't have the actual device?
+  // TEST_ASSERT_EQUAL(config.application_crc32, computed_crc32);
```
farazkh80, line 53: this fails because the config crc32 is 0. Not sure if it is necessary to assert equal the codes here?

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,34 @@
...
+    // claculate crc32
+    crc_temp = crc32_arr((uint8_t *)&buffer, sizeof(buffer));
+
+    crc32_codes[crc32_code_number] = crc_temp;
+    crc32_code_number++;
+    curr_size += sizeof(buffer);
+  }
+
+  // calculate the final crc32 and return it
+  return crc32_arr((uint8_t *)&crc32_codes, sizeof(crc32_codes));
```
ryandancy, line 33: Similar thing here: use `(uint8_t *)crc32_codes` (no `&`) instead because `crc32_codes` is implicitly converted to a `uint32_t *`.

Also, the behaviour here will depend on the [endianness](https://en.wikipedia.org/wiki/Endianness) of the system: since you're treating `uint32_t`s as arrays of four `uint8_t`s, you'll get different results on a little-endian system, where a uint32_t like `0x12345678` is laid out like `78 56 34 12` in memory, than on a big-endian system, where `0x12345678` would be laid out like `12 34 56 78` in memory. This hopefully shouldn't be too much of an issue since most systems are little-endian, but to make it robust, I'd recommend creating a second array of `uint8_t`s and manually copying each byte of each uint32_t in `crc32_codes` to the uint8_t array (using bitwise operations to extract each byte). Since you're manually dealing with the bytes in this case, it should be the same regardless of endianness.

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,34 @@
...
+#include "crc32.h"
+#include "flash.h"
+
+// calculates the crc32 for the full address, with 2048 byte increments
+// at a time, returning the final crc32 code
+uint32_t calculated_application_crc32(uintptr_t address, size_t size) {
+  // code array for holding crc32 codes per 2048 bytes
+  // size of array is 54 since
+  // BOOTLOADER_APPLICATION_SIZE / 2048 = 54
+  uint32_t crc32_codes[54];
```
ryandancy, line 15: 54 shouldn't be hardcoded here - instead, use a #define for the length of the buffer (say `BUFFER_LEN`) and then you can declare `crc32_codes` as
```c
uint32_t crc32_codes[BOOTLOADER_APPLICATION_SIZE / BUFFER_LEN + 1];`
```
(with the +1 for safety in case it doesn't align exactly).

You should also check that `size / BUFFER_LEN <= SIZEOF_ARRAY(crc32_codes)` (`SIZEOF_ARRAY` is defined in misc.h and gives the length of the array), because otherwise you'll overflow the end of the array.

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include "stdio.h"
+
+#include "bootloader_mcu.h"
+#include "crc32.h"
+#include "flash.h"
+
+// calculates the crc32 for the full address, with 2048 byte increments
+// at a time, returning the final crc32 code
+uint32_t calculated_application_crc32(uintptr_t address, size_t size) {
```
ryandancy, line 11: I've just looked over the implementation of `flash_read`, and it looks like it just `memcpy`s data to the buffer, reading from the flash address it's passed directly. In that case, I think you might be able to just do
```c
crc32_arr(BOOTLOADER_APPLICATION_START, BOOTLOADER_APPLICATION_SIZE)
```
to compute the crc32 of the whole application (which sort of makes this function obsolete, sorry about that).

Test that and see if it works, and then if it does, you might want to just implement this function as the `crc32_arr` call above (and just don't take the `address` and `size` parameters, making this function just a shortcut to compute the crc32 of the application).

projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,55 @@
...
+
+// intialize memory with random numbers for testing
+void intialize_memory() {
+  size_t curr_size = 0;
+  // writing the memory using a buffer with 2048 bytes at a time
+  uint8_t buffer[2048];
+  unsigned int seed = 42;
+
+  // generating random values in range of uint8(0->255)
+  for (int i = 0; i < 2048; i++) buffer[i] = rand_r(&seed) % (255 + 1 - 0) + 0;
```
ryandancy, line 24: I'm pretty sure `rand_r` is a Linux thing (so not available on stm32), CI is choking over it.  Use `rand` from <stdlib> instead, and seed it with `srand`.

ryandancy, line 24: Use `uint16_t` or something instead of `int` and `unsigned int`.

projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,55 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "persist.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+uint32_t s_crc32_codes_size = 54;
+
+// intialize memory with random numbers for testing
+void intialize_memory() {
```
ryandancy, line 17: Nit: the parameter list should be `(void)`. Also, typo: `intialize -> initialize`.

projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,55 @@
...
+#include "config.h"
+#include "crc32.h"
+#include "flash.h"
+#include "interrupt.h"
+#include "log.h"
+#include "persist.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+uint32_t s_crc32_codes_size = 54;
```
ryandancy, line 14: This should be `static`.

projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,55 @@
...
+  // writing the memory using a buffer with 2048 bytes at a time
+  uint8_t buffer[2048];
+  unsigned int seed = 42;
+
+  // generating random values in range of uint8(0->255)
+  for (int i = 0; i < 2048; i++) buffer[i] = rand_r(&seed) % (255 + 1 - 0) + 0;
+
+  while (curr_size <= BOOTLOADER_APPLICATION_SIZE) {
+    // write flash
+    flash_write((uintptr_t)BOOTLOADER_APPLICATION_START + curr_size, (uint8_t *)&buffer,
```
ryandancy, line 28: Just pass `buffer` here (see above).

projects/bootloader/test/test_bootloader_crc32.c
```diff
@@ -0,0 +1,55 @@
...
+                sizeof(buffer));
+    curr_size += sizeof(buffer);
+  }
+}
+
+void setup_test(void) {
+  flash_init();
+  crc32_init();
+  config_init();
```
ryandancy, line 37: `config` also needs soft timers initialized (which in turn need interrupts initialized), see the comment at the top of config.h.

projects/bootloader/test/test_bootloader_crc32.c
ryandancy: The strategy for testing a crc32-like function like this would be to seed flash with data for which you know the crc32, then assert that the crc32 matches. In this case, using random data might not be the best idea - instead use something predictable (like set `buffer[i]` to `i % 231` or something). To find what crc32 it should have, I recommend creating the array in Python and passing it through the [zlib.crc32 function](https://docs.python.org/3/library/zlib.html#zlib.crc32) to find the crc32.

You don't need to do anything with `config` in this test, the config won't be set automatically or anything.

projects/bootloader/src/stm32f0xx/jump_to_application.c
```diff
@@ -19,8 +22,17 @@ static noreturn __attribute__((naked)) void prv_perform_jump(uint32_t sp, uint32
...
+  config_get(&config);
+
+  // get the computed crc32 code
+  uint32_t computed_crc32 = calculated_application_crc32((uintptr_t)BOOTLOADER_APPLICATION_START,
+                                                         BOOTLOADER_APPLICATION_SIZE);
+
+  if (config.application_crc32 != computed_crc32) {
+    LOG_DEBUG("CRC32 codes do not match returning");
+    return;
+  }
```
ryandancy, line 35: As specified above, this should be moved to the new module that handles the jump-to-application datagram (and instead of just logging and returning when the crc32 doesn't match, you should also send back a status datagram with a nonzero status).

projects/bootloader/src/stm32f0xx/jump_to_application.c
```diff
@@ -19,8 +22,17 @@ static noreturn __attribute__((naked)) void prv_perform_jump(uint32_t sp, uint32
...
-  // TODO(SOFT-413): deinitialize any libraries the bootloader uses before doing this
+  BootloaderConfig config = { 0 };
+  config_get(&config);
+
+  // get the computed crc32 code
+  uint32_t computed_crc32 = calculated_application_crc32((uintptr_t)BOOTLOADER_APPLICATION_START,
+                                                         BOOTLOADER_APPLICATION_SIZE);
+
+  if (config.application_crc32 != computed_crc32) {
+    LOG_DEBUG("CRC32 codes do not match returning");
```
ryandancy, line 33: There needs to be a `\n` at the end of log messages.

ryandancy (changes requested):
Nice, I can see you put a lot of work into the crc32 stuff.

There is a big other part of the ticket which you might have missed / I might not have explained well, in that implementing the jump-to-application *command* means we need a module that registers a callback with the `dispatcher` module for jump-to-application datagrams (format [here](https://uwmidsun.atlassian.net/l/c/f0gMdqTr)), and when a jump-to-application datagram is received, does all these checks before sending back a status datagram (format specified on the same page) and calling the `jump_to_application` function.

The new module should follow roughly the same format as the `ping` module, but in the callback you should put the code that gets the config, computes the application crc32, and then compares them. Look in ping.c to see how to register to receive a type of datagram and how to send a response datagram. Let me know if you want to have a quick call at some point to clear up any confusion!

projects/bootloader/inc/bootloader_crc32.h
```diff
@@ -0,0 +1,10 @@
+// Calculates the crc32 code for the flash memory for bootloader's
+// jump to application functionality
+#pragma once
+
+#include <stddef.h>
+#include <stdint.h>
+
+// calculates the crc32 for the full address, with 2048 byte increments
+// at a time, returning the final crc32 code
+uint32_t calculated_application_crc32(uintptr_t address, size_t size);
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,34 @@
...
+  uint8_t crc32_code_number = 0;
+  size_t curr_size = 0;  // memory size that has been read
+  uint8_t buffer[2048];  // buffer for holding 2048 bytes of flash
+  uint32_t crc_temp;
+
+  while (curr_size <= size) {
+    // read from flash
+    flash_read(address + curr_size, sizeof(buffer), (uint8_t *)&buffer, sizeof(buffer));
+    // claculate crc32
+    crc_temp = crc32_arr((uint8_t *)&buffer, sizeof(buffer));
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: removing it since it is not used

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: done

projects/bootloader/src/stm32f0xx/jump_to_application.c
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: removing it since config is not used

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,34 @@
+#include "bootloader_crc32.h"
+
+#include "stdio.h"
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: done

projects/bootloader/src/stm32f0xx/jump_to_application.c
farazkh80: done

projects/bootloader/test/test_bootloader_crc32.c
farazkh80: done

farazkh80: done

projects/bootloader/inc/jump_to_application_dispatcher.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+// This module registers a callback with the dispatcher
+// for jump_to_application operation on the bootloader
+#include "status.h"
+#include "stdint.h"
+
+#define FAILURE_STATUS 1
+#define SUCCESS_STATUS 0
```
ryandancy, line 9: Also put these in the .c file.

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,49 @@
...
+#include "bootloader_datagram_defs.h"
+#include "can_datagram.h"
+#include "config.h"
+#include "dispatcher.h"
+#include "log.h"
+
+// response config datagram setup
+static uint8_t s_status;
+static CanDatagramTxConfig s_response_config = {
+  .dgram_type = BOOTLOADER_DATAGRAM_JUMP_TO_APP,
```
ryandancy, line 18: This should just be `BOOTLOADER_DATAGRAM_STATUS_RESPONSE`

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,49 @@
...
+  printf("crc32 is %d\n\n", computed_crc32);
+  if (config.application_crc32 != computed_crc32) {
+    LOG_DEBUG("CRC32 codes do not match returning\n");
+    s_status = FAILURE_STATUS;
+    return can_datagram_start_tx(&s_response_config);
+  }
+
+  jump_to_application();
+  s_status = SUCCESS_STATUS;
+  return can_datagram_start_tx(&s_response_config);
```
ryandancy, line 43: You'll need to TX the datagram before calling `jump_to_application` since that function doesn't return. So what you'll need to do is put the `jump_to_application()` call in a TX completion callback that you pass as the `.tx_cmpl_cb` in the CanDatagramTxConfig instead of `tx_cmpl_cb`.

ryandancy (changes requested):
Just wanted to give a review while you're fixing the crc32/flash stuff.

projects/bootloader/inc/bootloader_crc32.h
```diff
@@ -0,0 +1,12 @@
+// Calculates the crc32 code for the flash memory for bootloader's
+// jump to application functionality
+#pragma once
+
+#include <stddef.h>
+#include <stdint.h>
+
+#define BUFFER_LEN 2048
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/inc/jump_to_application_dispatcher.h
farazkh80: done

projects/bootloader/src/jump_to_application_dispatcher.c
farazkh80: done

projects/bootloader/src/jump_to_application_dispatcher.c
farazkh80: done

projects/bootloader/src/stm32f0xx/jump_to_application.c
```diff
@@ -4,7 +4,10 @@
 #include <stdnoreturn.h>
 #include <string.h>
 
+#include "bootloader_crc32.h"
```
mitchellostler, line 7: Do you need these includes? it doesn't look like you use anything from them in any of these files

projects/bootloader/test/test_jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,163 @@
...
+  TEST_ASSERT_EQUAL(SUCCESS_STATUS, *(rx_config.data));
+  TEST_ASSERT_EQUAL(BOOTLOADER_DATAGRAM_STATUS_RESPONSE, rx_config.dgram_type);
+}
+
+// to be developed once flash bug is fixed
+void test_jump_to_application_failure(void) {
+  TEST_ASSERT_OK(initialize_memory(false));
+  TEST_ASSERT_OK(config_commit(&test_input_config));
+
+  TEST_ASSERT_OK(initialize_memory(true));
```
farazkh80, line 138: This is supposed to set the flash to all 1s but fails to do so.

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,56 @@
...
+};
+
+static StatusCode prv_jump_to_application_response(uint8_t *data, uint16_t data_len,
+                                                   void *context) {
+  BootloaderConfig config = { 0 };
+  config_get(&config);
+
+  // get the computed crc32 code
+  uint32_t computed_crc32 = calculate_application_crc32();
+  // printf("crc32 is %u vs %u\n\n", config.application_crc32, computed_crc32);
```
ryandancy, line 41: Remove this commented out code

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,56 @@
...
+#define FAILURE_STATUS 1
+#define SUCCESS_STATUS 0
+
+// response config datagram setup
+static uint8_t s_status;
+
+void prv_jump_to_application_tx_cmpl_cb() {
+  jump_to_application();
+}
+static CanDatagramTxConfig s_response_config = {
```
ryandancy, line 24: Nitpick: blank line here

ryandancy (changes requested):
Really minor nitpicks, otherwise looks good!

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,57 @@
...
+  // get the computed crc32 code
+  uint32_t computed_crc32 = calculate_application_crc32();
+  // printf("crc32 is %d\n\n", config.application_crc32);
+  if (config.application_crc32 != computed_crc32) {
+    LOG_DEBUG("CRC32 codes do not match returning\n");
+    s_status = FAILURE_STATUS;
+    return can_datagram_start_tx(&s_response_config);
+  }
+
+  jump_to_application();
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/src/bootloader_crc32.c
```diff
@@ -0,0 +1,51 @@
...
+uint32_t calculate_application_crc32() {
+  // code array for holding crc32 codes per 2048 bytes
+  // size of array is BOOTLOADER_APPLICATION_SIZE / BUFFER_LEN + 1
+  uint8_t crc32_codes[(BOOTLOADER_APPLICATION_SIZE / BUFFER_LEN + 1) * 4];
+  uint8_t buffer_bytes_size = 0;
+  uint8_t crc32_code_number = 0;
+  size_t curr_size = 0;        // memory size that has been read
+  uint8_t buffer[BUFFER_LEN];  // buffer for holding 2048 bytes of flash
+  uint32_t crc_temp = 0;
+  uint8_t firstByte = 0, secondByte = 0, thirdByte = 0, fourthByte = 0;
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/src/jump_to_application_dispatcher.c
```diff
@@ -0,0 +1,57 @@
...
+};
+
+static StatusCode prv_jump_to_application_response(uint8_t *data, uint16_t data_len,
+                                                   void *context) {
+  BootloaderConfig config = { 0 };
+  config_get(&config);
+
+  // get the computed crc32 code
+  uint32_t computed_crc32 = calculate_application_crc32();
+  // printf("crc32 is %d\n\n", config.application_crc32);
```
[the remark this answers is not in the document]
farazkh80: done

projects/bootloader/src/jump_to_application_dispatcher.c
farazkh80: done

ryandancy (approved):

PR #436, opened 2021-09-13
Soft 515 bts7004 uv vbat

benlin1221:
Not sure if the tests I wrote were correct, please let me know


libraries/ms-drivers/test/test_bts7040_load_switch.c
```diff
@@ -365,6 +365,88 @@ void test_bts7040_current_sense_get_measurement_pca9539r_valid(void) {
...
 }
 
+// Test that bts7040_get_measurement with the bts7004 scaling factor returns a different reading
+// than normal
+void test_bts7004_current_sense_get_measurement_stm32_valid(void) {
+  // this doesn't matter (adc isn't reading anything) but can't be null
+  GpioAddress test_sense_pin = { .port = GPIO_PORT_A, .pin = 0 };
+  GpioAddress test_enable_pin = { .port = GPIO_PORT_A, .pin = 1 };  // EN pin
+  uint32_t interval_us = 500;                                       // 0.5 ms
+  Bts7040Stm32Settings bts7004settings = {
```
ryandancy, line 375: Nitpick: could you change these names like `btsXXXXsettings` to `btsXXXX_settings` (both here and in test_output.c)? That way they better comply with snake_case and look less smushed together (we tend to treat `bts7040` as one word for snake_case purposes).

projects/power_distribution/inc/output.h
```diff
@@ -105,6 +105,7 @@ typedef struct OutputBts7200Spec {
 typedef struct OutputBts7040Spec {
   Pca9539rGpioAddress enable_pin;
   uint8_t mux_selection;  // what should we select on the mux to read current from the BTS7040?
+  bool use_bts7004_scaling;
```
ryandancy, line 108: In output_config.c, in the spec for `FRONT_OUTPUT_UV_VBAT` (around line 300 or so), could you set this attribute to `true`? UV VBAT is on a BTS7004, and changing its scaling was the reason we needed this ticket.

ryandancy (changes requested):
This looks amazing, great job Ben! One minor nitpick and a tiny thing to add and then we can get this merged.

ryandancy:
Also, one of the tests you added seems to be failing - please figure out why and fix it.

ryandancy:
Just the one change to add the line in `output_config.c` please, then we're good to go.

ryandancy (approved):

PR #441, opened 2021-09-18
[SOFT-999] This is my check-in for Project 102.

Lynx101:


projects/hello_world/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
```
AveryChiu64, line 1: The hello world project should not be included in this PR.

projects/leds/src/main.c
```diff
@@ -4,9 +4,10 @@
 #include <stddef.h>
 #include <stdint.h>
```
AveryChiu64, line 5: Make sure to remove any extra files that are not part of the PR. We normally want to clean our pull request clean and focused on one ticket without extraneous files. In this case we would want to focus on only the FW 102 hwk

projects/soft_999_Linux_Heller_fw_102/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+  // Stuff
+  interrupt_init();
+  soft_timer_init();
+
+  Counter counter;
+
+  counter.counter_a = 0;
+  counter.counter_b = 0;
+
+  soft_timer_start_millis(500, counter_timer_callback, &counter, NULL);
```
AveryChiu64, line 45: What is 500 here? (In this case since I know what the tasks is I know what this means). However, someone who is skimming through this code without much background knowledge wouldn't know what 500 means. Does this mean 500 seconds? hours? We normally use macros (check out #define https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/1591803936/C+Reference+for+Midnight+Sun+Firmware+Team)

projects/soft_999_Linux_Heller_fw_102/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+  while (true) {
```
AveryChiu64, line 47: Add a `wait()` in here and include the wait library so that we do not waste processor resources as we are waiting for interrupts to fire.

projects/soft_999_Linux_Heller_fw_102/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+  if (value == 0) {
+    counter_storage->counter_b++;
+    LOG_DEBUG("Counter B: %u\n", counter_storage->counter_b);
+  }
+
+  soft_timer_start_millis(500, counter_timer_callback, counter_storage, NULL);
+}
+
+int main(void) {
+  // Stuff
```
AveryChiu64, line 36: Though commenting is generally good. In such a case comments like these may clutter our code and not provide any insightful information.

projects/soft_999_Linux_Heller_fw_102/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+  interrupt_init();
+  soft_timer_init();
+
+  Counter counter;
+
+  counter.counter_a = 0;
```
AveryChiu64, line 42: You can actually initialize all values to 0 when you declare it by using
`Counters counter = { 0 };`

projects/soft_999_Linux_Heller_fw_102/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+void counter_timer_callback(SoftTimerId timer_id, void *context) {
+  Counter *counter_storage = context;
+
+  uint8_t value = counter_storage->counter_a % 2;
+
+  counter_storage->counter_a++;
+  LOG_DEBUG("Counter A: %u\n", counter_storage->counter_a);
+
+  if (value == 0) {
```
AveryChiu64, line 27: You probably don't need the `value` variable here. Just toss in `counter_storage->counter_a % 2` into the if statement

PR #442, opened 2021-09-18
Soft 535 bootloader query

ShiCheng-Lu:


projects/bootloader/src/main.c
```diff
@@ -54,7 +54,9 @@ int main(void) {
   can_datagram_init(&s_datagram_settings);
 
   dispatcher_init(board_id);
+
   ping_init(board_id);
+  // query_init(blconfig);
```
ryandancy, line 59: Any reason this is commented out? We'll need to initialize the module...

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+  }
+  // compare the field to the board field
+  if (strncmp(target, stream->state, stream->bytes_left) == 0) {
+    s_results[field->index] = FOUND;
+  }
+  return true;
+}
+
+static StatusCode prv_check_query(uint8_t *data, uint16_t data_len, void *context) {
+  for (int i = 0; i < NUM_QUERY_FIELDS; ++i) {
```
ryandancy, line 79: Use `uint8_t` here instead of `int` please

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+    s_results[i] = NON_EXISTANT;
+  }
+
+  pb_istream_t pb_istream = pb_istream_from_buffer(data, data_len);
+  pb_decode(&pb_istream, Querying_fields, &s_querying);
+
+  for (int i = 0; i < NUM_QUERY_FIELDS; ++i) {
```
ryandancy, line 86: Also here

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+    return true;
+  }
+  s_results[field->index] = NOT_FOUND;
+
+  char *target = (char *)s_targets[field->index];
+  if (target == NULL) {
+    return true;
+  }
+  // compare the field to the board field
+  if (strncmp(target, stream->state, stream->bytes_left) == 0) {
```
ryandancy, line 72: To be extra safe you might want to pass `MIN(stream->bytes_left, MAX_STRING_SIZE)` as the third parameter to strncmp (`MIN` is from misc.h).

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// compares the query field to the board field directly without storying the query
+static bool prv_decode_cmp_string(pb_istream_t *stream, const pb_field_iter_t *field, void **arg) {
+  if (s_results[field->index] == FOUND) {
+    return true;
+  }
+  s_results[field->index] = NOT_FOUND;
+
+  char *target = (char *)s_targets[field->index];
```
ryandancy, line 67: You shouldn't need the cast here since it's a void pointer.

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+  return true;
+}
+
+static StatusCode prv_check_query(uint8_t *data, uint16_t data_len, void *context) {
+  for (int i = 0; i < NUM_QUERY_FIELDS; ++i) {
+    s_results[i] = NON_EXISTANT;
+  }
+
+  pb_istream_t pb_istream = pb_istream_from_buffer(data, data_len);
+  pb_decode(&pb_istream, Querying_fields, &s_querying);
```
ryandancy, line 84: Should probably check the return value of `pb_decode` here

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+      // do not match the query
+      return STATUS_CODE_OK;
+    }
+  }
+  return can_datagram_start_tx(&s_response_config);
+}
+
+// encode a string to a pb_ostream_t
+static bool prv_encode_string(pb_ostream_t *stream, const pb_field_iter_t *field,
+                              void *const *arg) {
```
ryandancy, line 97: You might want to include a comment here explaining what exactly `void *const *arg` means ("pointer to const pointer to void" according to [cdecl](https://cdecl.org)) since C's types get nigh unreadable at this level of nesting.

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+  const char *str = (const char *)(*arg);
```
ryandancy, line 98: I don't think the cast is needed here, but I might be wrong.

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+  // add to stream
+  if (!pb_encode_tag_for_field(stream, field)) {  // write tag and wire type
+    return false;
+  }
+  return pb_encode_string(stream, (uint8_t *)str, strlen(str));  // write sting
```
ryandancy, line 103: Just to be safe, you should probably pass `MIN(strlen(str), MAX_STRING_SIZE)` here too (what if `str` wasn't null-terminated somehow and strlen iterated through way too much memory?).

projects/bootloader/test/test_ping.c
```diff
@@ -106,3 +106,25 @@ void test_ping_addressed_to_multiple(void) {
   TEST_ASSERT_EQUAL(s_board_id, rx_config.data[0]);
   TEST_ASSERT_EQUAL(BOOTLOADER_DATAGRAM_PING_RESPONSE, rx_config.dgram_type);
 }
+
+void test_ping_no_resopnse(void) {
```
ryandancy, line 110: Typo: `resopnse -> response`

projects/bootloader/test/test_query.c
```diff
@@ -0,0 +1,280 @@
...
+    0x71, 0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79, 0x7A, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36,
+    0x37, 0x38, 0x39, 0x31, 0x30, 0x2A, 0x3F, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68, 0x69,
+    0x6A, 0x6B, 0x6C, 0x6D, 0x6E, 0x6F, 0x70, 0x71, 0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79,
+    0x7A, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68, 0x69, 0x6A, 0x6B, 0x6C, 0x6D, 0x6E, 0x6F,
+    0x70, 0x71, 0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79, 0x7A, 0x31, 0x32, 0x33, 0x34, 0x35,
+    0x36, 0x37, 0x38, 0x39, 0x31, 0x30,
+  };
+  // response
+  size_t expected_len = len;
+  uint8_t *expected_response = data;
```
ryandancy, line 140: You set up all these variables and didn't do anything with them, did you forget to finish this test?

projects/bootloader/test/test_query.c
```diff
@@ -0,0 +1,280 @@
...
+  };
+
+  dgram_helper_mock_tx_datagram(&tx_config);
+  dgram_helper_mock_rx_datagram(&rx_config);
+
+  TEST_ASSERT_EQUAL(DATAGRAM_STATUS_RX_COMPLETE, can_datagram_get_status());
+
+  TEST_ASSERT_EQUAL(BOOTLOADER_DATAGRAM_QUERY_RESPONSE, rx_config.dgram_type);
+  TEST_ASSERT_EQUAL(rx_config.data_len, expected_len);
+  for (int i = 0; i < rx_config.data_len; ++i) {
```
ryandancy, line 66: `uint16_t` or `uint32_t` here please instead of `int`

projects/bootloader/test/test_query.c
```diff
@@ -0,0 +1,280 @@
...
+  // "current_project" : [ "pedal board", "centre console" ]
+  // "project_info" : ["testing"]
+  // "git_version" : []
+  size_t query_len = 58;
+  uint8_t query[58] = {
+    0x08, 0x01, 0x08, 0x02, 0x08, 0x03, 0x12, 0x05, 0x68, 0x65, 0x6C, 0x6C, 0x6F, 0x12, 0x05,
+    0x77, 0x6F, 0x72, 0x6C, 0x64, 0x1A, 0x0B, 0x70, 0x65, 0x64, 0x61, 0x6C, 0x20, 0x62, 0x6F,
+    0x61, 0x72, 0x64, 0x1A, 0x0E, 0x63, 0x65, 0x6E, 0x74, 0x72, 0x65, 0x20, 0x63, 0x6F, 0x6E,
+    0x73, 0x6F, 0x6C, 0x65, 0x22, 0x07, 0x74, 0x65, 0x73, 0x74, 0x69, 0x6E, 0x67,
+  };
```
ryandancy, line 165: How did you generate these hexdumps? Could you generate them with nanopb inside the test instead of declaring these raw hexdumps? Currently it's impossible to edit by hand or verify that the data is what the comments say it is. (Also, pro tip: you can just declare `uint8_t query[] = {...}` and C will fill in the length for you.)

ryandancy (changes requested):
This is incredible ShiCheng, the module and tests all look completely correct! The comments are mostly formatting nitpicks and safety adjustments, great job!

projects/bootloader/src/query.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+StatusCode query_init(BootloaderConfig *config) {
+  // set up the search variables to compare the query to
+  // [tag - 1] as the tags start at index 0
```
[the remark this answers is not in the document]
ShiCheng-Lu: protobufs must have "positive" tags, so 0 is not allowed. it has to be that way...

projects/bootloader/src/main.c
ShiCheng-Lu: forgot lol

projects/bootloader/src/query.c
```diff
@@ -39,6 +39,16 @@ static CanDatagramTxConfig s_response_config = {
   .tx_cmpl_cb = tx_cmpl_cb,
 };
 
+static size_t prv_strnlen(const char *str, size_t maxlen) {
```
ryandancy, line 42: This looks nice and clean, if you find yourself using this again in another file it might be worth moving it to the libcore library.

ryandancy (approved):
LGTM, I love how readable the tests are now!

PR #444, opened 2021-09-19
Firmware 102 HW and Module 2

aadiamazing5:
added hello world and soft timers projs


projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+int main(void) {
+  // Library initializations:
+  interrupt_init();
+  soft_timer_init();
+
+  // Context initialization:
+  Counters storage = { 0 };
+
+  // Initialize timer loop:
+  soft_timer_start(10000,
```
mitchellostler, line 44: start_millis

projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+} Counters;
+
+// Counter functions:
+void counter_a_func(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  uint8_t counter_a = 1;
+
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a++);
+
+  soft_timer_start(500000, counter_a_func, storage, NULL);
```
mitchellostler, line 21: Usually we use soft_timer_start_millis() to make the code more readable. It's essentially the same thing, but then we can take off 3 of the zeros.

projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+// Counters Struct
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+// Counter functions:
+void counter_a_func(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  uint8_t counter_a = 1;
```
mitchellostler, line 17: This variable is not used, you can get rid of it

projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a++);
+
+  soft_timer_start(500000, counter_a_func, storage, NULL);
+}
+
+void counter_b_func(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+  uint8_t counter_b = 1;
```
mitchellostler, line 26: Same here, can also get rid of it

projects/mod2_fw_102/src/main.c
mitchellostler, line 44: It also looks like your values here are off by a factor of 10 (too small), but you can fix this when you use start_millis

projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+  soft_timer_init();
+
+  // Context initialization:
+  Counters storage = { 0 };
+
+  // Initialize timer loop:
+  soft_timer_start(10000,
+                   counter_a_func,  // function callback
+                   &storage, NULL);
+
```
mitchellostler, line 47: start_millis

projects/mod2_fw_102/src/main.c
```diff
@@ -0,0 +1,56 @@
...
+  soft_timer_start(50000,
+                   counter_b_func,  // functioncallback
+                   &storage, NULL);
+
+  while (true) {
+    wait();  // let functions run themselves indefinetly
```
mitchellostler, line 53: Nit: typo

mitchellostler (changes requested):
Looks really good! just a few changes to be made to make it perfect

mitchellostler:
Part of the homework criteria was to use only 1 soft-timer. Your code looks good, so I'm happy to approve your FW 102, but maybe think about how you could do it with 1, and ask me if you don't know 😁

projects/mod2_fw_102/src/main.c
mitchellostler, line 21: Also, for the constant values (number of milliseconds/microseconds) it's better to not just have a number

Use a #define statement at the top of your code with some context and the units to make your code more readable, and then pass this in lieu of the constant:

`#define COUNTER_A_PERIOD_MS 500`
...
`soft_timer_start_millis(COUNTER_A_PERIOD_MS...`

mitchellostler:
Just a note, in the future its a good idea to click the "resolve comment" button on each thing in Github. Other than that, good job, you can move on to FW 103

PR #445, opened 2021-09-20
firmware 102 homework

EC28:


projects/102_homework/src/main.c
```diff
@@ -0,0 +1,45 @@
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#include <stdint.h>
```
mitchellostler, line 6: Nit: It's generally good practice to include system libraries first

projects/102_homework/src/main.c
```diff
@@ -0,0 +1,45 @@
...
+  increment->counter_a++;
+
+  LOG_DEBUG("Counter A: %i\n", increment->counter_a);
+
+  if (increment->counter_a % 2 == 0) {
+    increment->counter_b++;
+
+    LOG_DEBUG("Counter B: %i\n", increment->counter_b);
+  }
+
```
mitchellostler, line 28: Nit: there seems to be a lot of whitespace, which isn't necessarily a problem but makes things harder to read. I would put a bit more thought into how you group things together (Fix if you'd like)

mitchellostler (changes requested):
Looks really good! You can fix the nits if you'd like or just take note of them. 
The only real issue is that the CI build is not passing. It looks like you need to lint your code (ie format it in a way that github likes) this should be easy enough.

`make format` -> formats your code
`git add .` ->add the changes 
`make lint` -> should pass, if not it will tell you the issues. Only worry about the things to do with your code

mitchellostler:
Well done, you are good to move on to FW 103

PR #446, opened 2021-09-23
Firmware 102 Homework

ericzhao625:


projects/soft_999_eric_zhao_fw_102/src/main.c
```diff
@@ -0,0 +1,42 @@
+#include <stdbool.h>
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define INCREMENT_HALF_SECOND 0.5
```
mitchellostler, line 8: Convention is to not use decimal values in the soft_timer function. Better to use 500 millis

mitchellostler (commented):
other than the one point i made, it looks good, well done!

mitchellostler:
You're good to move on to FW103

PR #447, opened 2021-09-23
[SOFT-233] Sudo openocd

jessm:
Ok I know it's not ideal, but in the original box there's some chef recipe that sets up some rules so that devices can be opened by all users. I couldn't figure it out with backpack, but this is a super simple solution that would improve (mostly just my own) quality of life a lot and is pretty harmless.


ryandancy (approved):
LGTM, I don't think there's a root password in the regular box so this shouldn't cause password prompts or anything like that

PR #449, opened 2021-09-24
firmware tutorial 102

ShahriarAhnaf:
I could not get git push working in my VM unfortnately so I manually uploaded this.


mitchellostler:
Looks good, you are good to move on to FW 103. Also, if you're still having trouble with git reach out to me today :)

PR #450, opened 2021-09-25
[SOFT-999] Firmware 102 Homework

Bafran:


projects/fw_102/src/main.c
```diff
@@ -0,0 +1,49 @@
+#include <stdbool.h>
+#include "gpio.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define ITERATION_TIMEOUT 0.5
```
mitchellostler, line 8: Usually we try not to use floating point numbers for our delays. Try defining it as 500 and using soft_timer_start_millis() instead

mitchellostler (commented):
Other than the one minor point, this looks good! You can fix if you'd like, or just move on to FW 103. Sorry for the late feedback

projects/fw_102/src/main.c
mitchellostler: Also, it's a good idea to add units to the define for improved readability

PR #451, opened 2021-09-25
[SOFT-999] Firmware 102 hw

BENJIGAO:


mitchellostler (commented):
Hey, sorry for the late feedback!
This solution works, but part of the criteria was that the program would only use 1 soft timer. 
If you feel like fixing it go ahead, otherwise feel free to move on to 103!

PR #452, opened 2021-09-26
firmware 103 homework

EC28:


projects/103_homework/src/main.c
```diff
@@ -0,0 +1,60 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// interrupt call back function
+void interrupt_callback(const GpioAddress *address, void *context) {
```
mitchellostler, line 10: If a function is used only in the file it is declared in, ie by other methods, it is important to declare it static and prefix it with `prv_` so that we can know that it is a 'private function', only visible in the scope of the program. The concept of static in C allows us to better structure our code, and retain the scope of any function to only that which it needs. See [here](https://www.tutorialspoint.com/static-functions-in-c)

projects/103_homework/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+  uint16_t ADC_reading = 0;
+  if (adc_read_raw_pin(*ADC_address, &ADC_reading) == STATUS_CODE_OK) {
+    LOG_DEBUG("Output: &d\n", ADC_reading);
+  } else {
+    LOG_DEBUG("Failed to read.");
+  }
+}
+
+int main(void) {
+  gpio_it_init();
```
mitchellostler, line 22: If you have a look at each library you will see that there is a hierarchy of calling init functions.
For example, in gpio_it.h: 
`// Requires GPIO and interrupts to be initialized.`
So you must call gpio_init() and interrupt_init() before gpio_it_init()

projects/103_homework/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+#include "gpio_it.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+// interrupt call back function
+void interrupt_callback(const GpioAddress *address, void *context) {
+  GpioAddress *ADC_address = context;
+
+  uint16_t ADC_reading = 0;
```
mitchellostler, line 13: nit: We generally don't use capitals in variable names. `adc_reading` would be much more compliant.

mitchellostler (changes requested):
Hey Elvis, good work! just a few things to fix. Also, I noticed the CI build is failing, so if you could take a look at that and see if you can fix that (most likely lint error). Let me know if you have any questions, and I apologize about the late review

mitchellostler:
Hey Elvis, sorry I took so long to get to your review. As you know the teams been going through a bit of a transitory period, but we're slowly getting back on track. 
It looks good, well done! You are ready to move on to FW 104, and get a task. Contact me or Ryan if you are looking for one

PR #454, opened 2021-09-26
Correct I2C port in smoke_pca9539r

jessm:
This has caused too much grief to not be corrected.

The schematics for rev 3.0 are unclear, but this is what works in hardware right now.


ryandancy (approved):
Oof, this must have been rough to debug...

PR #455, opened 2021-09-27
Soft 540 Added motor current continuity tests for brake and throttle

danielq987:
The motor current continuity tests are implemented for throttle + brake, with cruise control both on and off. The continuity test is implemented such that the pedal value is increased by 0.95 each iteration, and compared with the results of the previous iteration. From testing, the increment value had to be below 1 to ensure small enough intervals, so a value of 0.95 was used.

There was an issue I faced with segfaults - I added a comment below to describe it in more detail.

In all, the time to run all the tests takes quite long, over two minutes just for the continuity tests. One idea I had was increasing the `TEST_ASSERT_FLOAT_WITHIN` threshhold from 0.01, thoughts?


projects/mci/test/test_mci_output.c
```diff
@@ -168,6 +174,15 @@ EEDriveOutput TEST_MOCK(drive_fsm_get_drive_state)() {
   return s_drive_state;
 }
 
+bool TEST_MOCK(drive_fsm_is_cruise)() {
+  return s_is_cruise;
+}
+
+// Mocking this function is somehow required to prevent a Segfault.
+bool TEST_MOCK(drive_fsm_toggle_cruise)() {
```
danielq987, line 182: If this function is not mocked, reading the current+velocity from line 681 of this file produces a SEGFAULT - will only occur for this test and no other tests. I hope this doesn't happen outside of a test setting?

ryandancy (commented):
Not a full review, but a suggestion on how to decrease the time requirements. You can see in mci_output.c that `prv_mci_output` is running in a soft timer loop, currently running every `MOTOR_CONTROLLER_DRIVE_TX_PERIOD_MS` ms, which is 200ms currently. The vast majority of our time is spent waiting for this.

Currently this soft timer loop delay is a #define'd macro. I suggest modifying mci_output to take this parameter as input to `mci_output_init` instead. You can take it as a parameter and store it in a static variable, and then initialize it with the current #define in main.c and with a much lower value in your test file (like 5 or 10ms rather than 200ms). Then you can decrease the delay in `prv_do_tx_rx_pedal_values` and the tests should run much faster.

PR #456, opened 2021-09-30
fw 103 homework

ericzhao625:


projects/soft_999_eric_zhao_fw_103/src/main.c
```diff
@@ -0,0 +1,57 @@
...
+
+  if (adc_read_raw_pin(*a6_addr, &a6_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("A6 data: %d\n", a6_data);
+  } else {
+    LOG_DEBUG("Failed to read A6 data");
+  }
+}
+
+int main(void) {
+  GpioAddress a6_reading_addr = {
```
mitchellostler, line 19: An alternate way of doing these gpio declarations that we use frequently to reduce clutter is
`#define ADC_ADDRESS {.port = GPIO_PORT_A, .pin = 6}`
This allows us to in main.c simply just do the assignment as 
`GpioAddress adc_ad = ADC_ADDRESS` which will insert the struct we defined above. 

The way you've done it is fine, just a thought :)

mitchellostler (commented):
Looks good! you are good to move on to FW 104

projects/soft_999_eric_zhao_fw_103/src/main.c
```diff
@@ -0,0 +1,57 @@
+#include "adc.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+void prv_analog_read_a6(const GpioAddress *address, void *context) {
```
mitchellostler, line 7: Private functions should be declared static, ie `static void`. This means that it will only be able to be accessed in the file in which it is declared, and lets us know that it will only be called by other methods

PR #457, opened 2021-10-01
[SOFT-552] PD current sense smoke test mode

ryandancy:
I figured it would be nice to have a more convenient smoke test for RPD2/FPD2 than the aging smoke_bts7200/smoke_bts7040 smoke tests. This outputs current sense readings for every output that supports it (regardless of bts7200/bts7040), and it can be customized to output whatever subset you want.


projects/power_distribution/src/main.c
```diff
@@ -1,6 +1,14 @@
 // Uncomment this line to force firmware to run as front or rear power distribution.
 // #define FORCE_IS_FRONT_POWER_DISTRIBUTION true
 
+// SMOKE TEST MODES: each macro is a different smoke test you can run.
+
+#define SMOKE_TEST_MODE_NORMAL 0               // run as normal PD
+#define SMOKE_TEST_MODE_CURRENT_MEASUREMENT 1  // see smoke_current_measurement.c
```
mitchellostler, line 7: So if I'm reading correctly, the idea that as we expand upon smoke testing infrastructure for PD that the user can adjust the "mode" by changing the defined macro here. 

Since we are looking at developing these smoke tests for each project, I am kind of thinking that this PR will represent the de-facto for how we go about it, and as new ones are written they will follow how this is setup, so I want to discuss a few things.

I like the idea of being able to change smoke tests by just adjusting the macro, but I am worried about cluttering the program as we increase to more than one smoke test.  If we are thinking multiple smoke tests per project, the same way that it is set up here, then it might make sense to have all smoke tests of the same function type (no return, no params), and have a lookup table defined in a `smoke_<project>.h` header file along with the macros defining each mode.

Then all we need in main is the one line which sets `SMOKE_TEST_MODE` and the one check against MODE_NORMAL in main, otherwise call whatever is in the lookup table. (effectively the same as what is happening now, just scales with more smoke tests)

This is more a factor in PD and other big projects, as increasing complexity just makes them harder to read and debug.

projects/power_distribution/src/main.c
```diff
@@ -1,6 +1,14 @@
 // Uncomment this line to force firmware to run as front or rear power distribution.
 // #define FORCE_IS_FRONT_POWER_DISTRIBUTION true
 
+// SMOKE TEST MODES: each macro is a different smoke test you can run.
+
+#define SMOKE_TEST_MODE_NORMAL 0               // run as normal PD
```
mitchellostler, line 6: Calling normal operation a smoke test mode makes it a bit harder to understand. Maybe `OP_MODE_NORMAL` vs `OP_MODE_SMOKE_CURRENT_MEASUREMENT` or something like that?

projects/power_distribution/src/smoke_current_measurement.c
```diff
@@ -0,0 +1,152 @@
...
+#define USE_FRONT_PD false
+
+// Delay between minicom dumps of current measurement data
+#define MEASUREMENT_DELAY_MS 1000
+
+// If you want to only measure current from certain outputs, uncomment the outputs you want to
+// read from. If this array is empty, we output everything; otherwise, we only output the currents
+// for the outputs in this array.
+// Note that we can't measure front outputs when initialized to rear or vice versa.
+static Output s_outputs_to_measure[] = {
```
mitchellostler, line 31: Really like this method of making it configurable, super easy to use

projects/power_distribution/inc/output.h
```diff
@@ -144,7 +150,7 @@ StatusCode output_init(OutputConfig *config, bool is_front_power_distro);
 StatusCode output_set_state(Output output, OutputState state);
 
 // Read the current that the output is drawing into |*current| in mA.
-// STATUS_CODE_INVALID_ARGS is returned if current sense isn't supported by this output.
+// STATUS_CODE_UNIMPLEMENTED is returned if current sense isn't supported by this output.
```
mitchellostler, line 153: nice catch

mitchellostler (commented):
Looks good, this is going to be super useful. Just a couple of comments about how we want to structure our smoke test within the project code. Since we are going to be building up our smoke tests, I'm thinking this is the example that we can start with to show how they will be structured.

projects/power_distribution/src/main.c
ryandancy: This makes a lot of sense, I'll change to this. I am a bit worried that `smoke_<project>.h` might look like another smoke test instead of a smoke test control header, so I think the name `smoketests_<project>.h` might fit better.

jessm:
Looks good and works when flashing, just need to get hardware to connect some headers so I can properly do eload stuff to verify full functionality

PR #458, opened 2021-10-01
FW103 HW - Aadi

aadiamazing5:


projects/fw103/src/main.c
```diff
@@ -0,0 +1,61 @@
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "wait.h"
+
+// Pin A6 Reader Function
+void read_pin_a6(const GpioAddress *address, void *context) {
```
mitchellostler, line 9: Any function that is only used in the file it is declared should be declared as `static void` and prefixed with `prv_`, and it's a good idea to put `_cb` or `_callback` at the end so that we know that it will be registered as a callback

This helps us keep track which functions are part of an interface and which are used only by other methods in the same file.

projects/fw103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  GpioAddress pin_a6 = { .port = GPIO_PORT_A, .pin = 6 };
+
+  GpioSettings a6_settings = {
+    GPIO_DIR_IN,
+    GPIO_STATE_LOW,
+    GPIO_RES_NONE,
+    GPIO_ALTFN_ANALOG,
+  };
+
+  GpioAddress button_pin = { .port = GPIO_PORT_B, .pin = 2 };
```
mitchellostler, line 32: An alternate way of doing these gpio declarations that we use frequently to reduce clutter is the use of preprocessor statements at the top of the file i.e.:
#define ADC_ADDRESS {.port = GPIO_PORT_A, .pin = 6}
This allows us to in main.c simply just do the assignment as
GpioAddress adc_ad = ADC_ADDRESS which will insert the struct we defined above.

The way you've done it is fine, just a thought :)

projects/fw103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+
+  // Interrupt Initializations
+  interrupt_init();
```
mitchellostler, line 35: Its a good idea to do all your initializations at the start of your main function, all together

mitchellostler (commented):
Looks good, just take note of the things that I've mentioned and you are good to move on!

PR #459, opened 2021-10-02
Soft 468 jumping back to bootloader

vaaranan-y:


projects/bootloader/src/stm32f0xx/jump_to_bootloader.c
```diff
@@ -0,0 +1,40 @@
...
+      : [sp] "r"(sp), [pc] "r"(pc));
+}
+
+void jump_to_bootloader(void) {
+  // Disable all interrupts so their is no interference when working with vector tables
+  __disable_irq();
+
+  // Reset the vector table to bootloader's vector table
+  SYSCFG_MemoryRemapConfig(
+      SYSCFG_MemoryRemap_SRAM);  // change where the vector table is stored to beginning of flash
```
ryandancy, line 28: This will set the vector table to the beginning of SRAM (which is where it is already since we're in the application) - you'll need to use `SYSCFG_MemoryRemap_Flash` instead (see stm32f0xx_syscfg.h).

projects/bootloader/src/stm32f0xx/jump_to_bootloader.c
```diff
@@ -0,0 +1,40 @@
...
+static noreturn __attribute__((naked)) void prv_perform_jump(uint32_t sp, uint32_t pc) {
+  __asm(
+      "msr msp, %[sp] \n"  // reset the main stack pointer (msp) to sp
+      "bx %[pc] \n"        // jump to pc
+
+      // this bizarre syntax associates the "sp" and "pc" in asm with the sp and pc parameters
+      // see http://www.ethernut.de/en/documents/arm-inline-asm.html
+      :
+      : [sp] "r"(sp), [pc] "r"(pc));
+}
```
ryandancy, line 20: We might want to factor this function out from this and jump_to_application.c into an ms-common (or ms-bootloader) library called `jump.h` since it just jumps to a particular location.

projects/bootloader/src/main.c
```diff
@@ -8,6 +8,7 @@
 #include "flash.h"
 #include "interrupt.h"
 #include "jump_to_application.h"
+#include "jump_to_bootloader.h"
```
ryandancy, line 11: So the next steps for this module is that we want to jump from the bootloader to the application when we receive a datagram message while in an application. This will be in can_fsm.c, in the `if (rx_msg.source_id == SYSTEM_CAN_DEVICE_BOOTLOADER)` if statement. Essentially, when that if statement is reached:
* if we're in an application running on the bootloader, in which case we should jump to the bootloader so that the bootloader can receive subsequent datagram messages
* oherwise, we should call `bootloader_can_receive` there like we do now so that the bootloader can process the message (if necessary)

So there are a couple of parts to doing this:
1. We'll need to move `jump_to_bootloader` to a library, perhaps called `ms-bootloader`, so can_fsm.c can reach it. Perhaps move bootloader_mcu.h to that library too.
2. We'll need a way to determine which of the three scenarios above we're in. This is a bit tricky: we'll need a preprocessor symbol that's defined when we're compiling for an application running on the bootloader. You can do this by adding a `-D<symbolname>` argument to the C compiler flags when compiling for the application, which currently you can do in platform/stm32f0xx/platform.mk by adding `CFLAGS += -D<symbolname>` to the `ifeq ($(MAKECMDGOALS),temp-bootloader-write)` if statement (replace `<symbolname> with the symbol you want defined when on the application only). Then you can test with `#ifdef` in can_fsm.c.

ryandancy (changes requested):
Looks pretty good, I'll test this in a bit, and I've left some action items in the comments.

projects/bootloader/inc/stm32f0xx/bootloader_mcu.h
```diff
@@ -23,3 +25,5 @@ extern uint32_t _vector_table_size;
 #define BOOTLOADER_RAM_START ((void *)&_ram_start)
 #define BOOTLOADER_RAM_SIZE ((size_t)&_ram_size)
 #define BOOTLOADER_VECTOR_TABLE_SIZE ((size_t)&_vector_table_size)
+#define BOOTLOADER_DEFAULT_LOCATION ((void *)&_bootloader_start)
```
ryandancy, line 28: I think `BOOTLOADER_START` might be a clearer name for this, especially since we'll likely never have the bootloader at a different location (so not "default").

projects/bootloader/src/stm32f0xx/jump_to_bootloader.c
```diff
@@ -0,0 +1,30 @@
+#include "jump_to_bootloader.h"
```
ryandancy, line 1: jump_to_application, jump_to_bootloader, and jump shouldn't be duplicated between the bootloader project and the ms-bootloader libraries, they should only appear in ms-bootloader. So remove the copies in the bootloader project / move all the functionality to ms-bootloader.

platform/stm32f0xx/platform.mk
```diff
@@ -44,6 +44,7 @@ endif
 DEFAULT_LINKER_SCRIPT ?= stm32f0_default.ld
 ifeq ($(MAKECMDGOALS),temp-bootloader-write)
 DEFAULT_LINKER_SCRIPT := stm32f0_application.ld
+CFLAGS += -DAPP_COMP_ON_BOOT
```
ryandancy, line 47: What does `APP_COMP_ON_BOOT` stand for? `BOOTLOADER_APPLICATION` might be clearer.

libraries/ms-bootloader/src/jump.c
```diff
@@ -0,0 +1,12 @@
+#include "jump.h"
+
+noreturn void prv_perform_jump(uint32_t sp, uint32_t pc) {
```
ryandancy, line 3: You'll need to add back the `__attribute__((naked))` here - it tells the compiler not to generate the standard entry/exit code for functions and just to use the raw assembly. It also will stop CI from complaining about `noreturn`.

Also, this is STM32-specific - you'll need to move it to an `stm32f0xx` folder and add a (no-op) `x86` implementation.

libraries/ms-bootloader/README.md
```diff
@@ -0,0 +1,16 @@
...
+        - What problem does it solve?
+        - How do I use it? (with usage examples / example commands, etc)
+        - How does it work? (architectural overview)
+    A README for a board project (powering a hardware board, e.g. power distribution, centre console, charger, BMS carrier) should answer the following questions:
+        - What is the purpose of the board?
+        - What are all the things that the firmware needs to do?
+        - How does it fit into the overall system?
+        - How does it work? (architectural overview, e.g. what each module's purpose is or how data flows through the firmware)
+-->
+# ms-bootloader
```
ryandancy, line 15: You should write a brief description of what goes in ms-bootloader here.

ryandancy (changes requested):

ryandancy (changes requested):
Looks good (barring tiny changes)! I'll approve once you validate it onsite :)

The big thing to validate is if we can successful jump back to the bootloader from an application a) on demand (i.e. using a test project) and b) when we receive a bootloader CAN message. Use `make temp-bootloader-write` as described in https://github.com/uw-midsun/firmware_xiv/pull/476 to flash (you'll need that PR on your branch). Mitchell can show you how to use `cansend` to send a bootloader CAN message manually.

PR #461, opened 2021-10-02
Soft 550 uv detector smoke test mode for pd

123Logan321:


projects/power_distribution/inc/smoketests_pd.h
```diff
@@ -8,6 +8,7 @@
 #define PD_SMOKE_TEST SMOKE_TEST_CURRENT_MEASUREMENT
 
 typedef enum {
+  SMOKE_TEST_UV_CUTOFF = 0,            // see smoke_uv_cutoff.c
   SMOKE_TEST_CURRENT_MEASUREMENT = 0,  // see smoke_current_measurement.c
```
ryandancy, line 12: It won't work having `= 0` on both - how C enums work is that they're numbered with sequential integers starting at 0, unless you put an `= N` (where `N` is some number), in which case that enum member gets that value. Our convention is that we explicitly set the first value to 0, then let C auto-number the rest of them. So the first value should have `= 0` and the rest shouldn't have anything.

projects/power_distribution/src/main.c
```diff
@@ -29,6 +29,7 @@
 #include "publish_data.h"
 #include "publish_data_config.h"
 #include "rear_strobe_blinker.h"
+#include "smoke_uv_cutoff.h"
```
ryandancy, line 32: You don't actually need to #include this since main.c doesn't use any of the symbols (functions, etc) it declares.

projects/power_distribution/inc/smoke_uv_cutoff.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+// A smoke test mode for the UV detector
+void smoke_uv_cutoff_perform(void);
```
ryandancy, line 4: For the last step in hooking your smoketest into the system, you'll need to add a line to the `s_smoke_tests` array in smoketests_pd.c. The purpose of that array is to map the `SmokeTestMode` enum values from smoketests_pd.h to functions like this, which are the actual functions called when somebody sets `PD_SMOKE_TEST` to a certain value in smoketests_pd.h.

Currently `SMOKE_TEST_CURRENT_MEASUREMENT` is mapped to the `smoke_current_measurement_perform`  function from smoke_current_measurement.h; you'll need to additionally map `SMOKE_TEST_UV_CUTOFF` to `smoke_uv_cutoff_perform` using the same syntax (it's called "designated inits" and it's a GNU extension to C, see https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html).

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,46 @@
...
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_NORMAL,
+  };
+
+  // callback occurs when pin turns low i.e falling
+  status_ok_or_return(gpio_it_register_interrupt(
+      detector_pin, &interrupt_settings, INTERRUPT_EDGE_FALLING, prv_pin_interrupt_handler, NULL));
+
+  return STATUS_CODE_OK;
+}
```
ryandancy, line 30: It looks like you've copy-pasted front_uv_cutoff.c here - you don't need to do that, you just need to call `front_uv_detector_init` in `smoke_uv_cutoff_perform` to initialize the module, and it'll do the rest.

Use `FRONT_UV_COMPARATOR_PIN` from pin_defs.h as the GpioAddress parameter - you might have to assign it to a variable and pass in its address though, or use the syntax `&(GpioAddress)FRONT_UV_COMPARATOR_PIN`, which works due to another GNU extension called "compound literals" (google it).

When you call `front_uv_detector_init`, you'll also want to wrap its call in `BUG()` - it's a macro from bug.h that will cause a big fancy error message to be printed if the wrapped call returns a StatusCode which isn't `STATUS_CODE_OK` (any non-OK value means that the initialization failed). Since this is a smoke test, we want lots of logging and to make it easy to figure out if something went wrong in software.

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,46 @@
...
+  gpio_init();
+  gpio_it_init();
+  event_queue_init();
+  interrupt_init();
+
+  Event e = { 0 };
+  while (true) {
+    while (event_process(&e) == STATUS_CODE_OK) {
+      can_process_event(&e);
+      pd_gpio_process_event(&e);
```
ryandancy, line 42: You don't need these calls here, just an empty while loop body will do for this smoke test.

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include "can_transmit.h"
+#include "front_uv_detector.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "pd_events.h"
+#include "pin_defs.h"
+#include "status.h"
+#include "wait.h"
```
ryandancy, line 11: Once you've made the changes, remove any #includes that you don't use - the general rule is that you should #include a header if and only if you use some symbol declared in it in the file (like a function name, typedef, struct name, macro, global variable, enum value, etc).

projects/power_distribution/inc/smoketests_pd.h
```diff
@@ -8,6 +8,7 @@
 #define PD_SMOKE_TEST SMOKE_TEST_CURRENT_MEASUREMENT
```
ryandancy, line 8: Oops, looks like I accidentally left this uncommented. Could you comment this out so that by default power distribution works normally rather than going into the current measurement smoke test?

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,46 @@
...
+      detector_pin, &interrupt_settings, INTERRUPT_EDGE_FALLING, prv_pin_interrupt_handler, NULL));
+
+  return STATUS_CODE_OK;
+}
+
+void smoke_uv_cutoff_perform(void) {
+  gpio_init();
+  gpio_it_init();
+  event_queue_init();
+  interrupt_init();
```
ryandancy, line 36: You need to initialize interrupts *before* initializing gpio interrupts with `gpio_it_init()`, because gpio interrupts depend on interrupts already being initialized.

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,46 @@
...
+
```
ryandancy, line 37: Could you add a `LOG_DEBUG` right after you initialize `front_uv_detector` saying something about how the UV cutoff smoke test has successfully initialized? Then people using the smoke test know that it's started successfully.

ryandancy (changes requested):
Nice first pass, Logan! Two general things: first, you can test if you've set everything up correctly by uncommenting `PD_SMOKE_TEST` in smoketests_pd.h and setting it to your `SMOKE_TEST_UV_CUTOFF` enum value.

Second, next time you can just create a branch instead of forking the repo - I'll add you to the github repo so you can push branches.

ryandancy (approved):
Looks good!

ryandancy:
But one thing you will need to fix: you can see that CI is failing. Click on "details" and find what the error is (currently it's a compiler error since `LOG_WARNING` isn't defined, you'll need to include log.h), and fix it so CI passes.

projects/power_distribution/src/smoke_uv_cutoff.c
```diff
@@ -0,0 +1,27 @@
...
+#include "wait.h"
+
+void smoke_uv_cutoff_perform(void) {
+  interrupt_init();
+  gpio_init();
+  gpio_it_init();
+  event_queue_init();
+
+  BUG(front_uv_detector_init(&(GpioAddress)FRONT_UV_COMPARATOR_PIN));
+  LOG_DEBUG("The front UV cutoff smoke test has been successfully initialized.");
```
ryandancy, line 19: One last thing I didn't catch last time: put a `\n` at the end of this string. Otherwise it won't insert a line break at the end of the line (it's like printf).

PR #462, opened 2021-10-05
Update Mcp2515 with sanity check changes

mitchellostler:
One issue that we've encountered with both mci and charger is that there wasn't a way in the smoke test to verify that the mcp2515 even has a valid connection. 

This is a direct check to provide a sort of sanity check that the config of the mcp is correct and it's not bricked

I tested it on hardware and it works


ryandancy (approved):
Looks good, seems super useful

PR #463, opened 2021-10-06
Soft 999 adel chinoy fw 103

Bafran:


projects/fw_103/src/main.c
```diff
@@ -0,0 +1,50 @@
...
+  GpioAddress *adc_addr = context;
+  uint16_t adc_reading = 0;
+  adc_read_converted_pin(*adc_addr, &adc_reading);
+
+  LOG_DEBUG("A6 ADC converted reading: %d\n", adc_reading);
+}
+
+int main(void) {
+  interrupt_init();
+  gpio_it_init();
```
AveryChiu64, line 18: I think gpio has to be initialized before gpio interrupts

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,50 @@
...
+  gpio_init();
+
+  GpioAddress adc_addr = { .port = GPIO_PORT_A, .pin = 6 };
```
AveryChiu64, line 21: Just a small formatting thing. We normally add a comma at the end so that it is formated better
(ie just add a comma like this ... `.pin = 6,` so that the code will be formated properly when we run make format

AveryChiu64: Same goes for the structs below

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+  gpio_init_pin(&adc_addr, &adc_settings);
+  gpio_init_pin(&button_addr, &button_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+
+  AdcChannel adc_channel = NUM_ADC_CHANNELS;
+
+  adc_get_channel(adc_addr, &adc_channel);
+  adc_set_channel(adc_channel, true);
```
AveryChiu64, line 38: This is fine, but we have new libraries that can automatically handle the channels for us. Check out `tutorial_board_adc`

PR #467, opened 2021-10-15
Fw 103 Hw

BENJIGAO:


projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+
+  gpio_init_pin(&pot_address, &pot_settings);
+
+  adc_init(ADC_MODE_SINGLE);
+  adc_set_channel_pin(pot_address, true);
+
+  gpio_it_register_interrupt(&button_address, &interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             &prv_button_interrupt_handler, &pot_address);
+
+  while (true) {
```
mitchellostler, line 57: can optionally add a wait() function call in the loop to save power

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+
+static GpioAddress button_address = {
+  .port = GPIO_PORT_B,
+  .pin = 2,
+};
+
+static GpioSettings pot_settings = {
```
mitchellostler, line 22: Generally we define these settings only in the function which we use them (main) to avoid cluttering the program. If you use them in more than one place in the program, like the GPIO addresses, then this is ok.

mitchellostler (commented):
Looks good! Added a few comments, but you are ready to move on to FW 104

PR #468, opened 2021-10-15
[SOFT-556] Charger Smoke-Test

123Logan321:


projects/charger/src/smoketests_charger.c
```diff
@@ -0,0 +1,33 @@
...
+  if (smoke_test >= NUM_SMOKE_TESTS) {
+    LOG_CRITICAL(
+        "Invalid smoke test! "
+        "Please set CHARGER_SMOKE_TEST to a valid value from the SmokeTest enum, "
+        "or comment it out to run Charger normally. (Name=%s, value=%d)\n",
+        smoke_test_name, smoke_test);
+    return;
+  }
+  if (s_smoke_tests[smoke_test] == NULL) {
+    LOG_CRITICAL(
```
mitchellostler, line 24: Probably don't need to log critical here unless there's something i'm missing, log_debug would be fine, since you return from the function anyways.

mitchellostler (approved):
Looks good! Well done. I noticed that the program doesn't run on x86 (this is a charger issue, not you) but luckily we don't need to deal with this now. Thanks for putting the work into this

PR #469, opened 2021-10-17
104 Homework Eric Zhao

ericzhao625:


mitchellostler (commented):
Hey Eric, Great work, sorry it took so long to get to your review. 

Part 1 and 2 look ok, but there are some issues with the I2C pins and the SPI transmission and receiving values.

The CAN part looks good. I tried running it and I got the ack parts, but none of the message callbacks were printing, so you should look into that. I tried to answer your inline questions as best I could, but if you have more please reach out and we can address them

mitchellostler (commented):
Good work!

PR #471, opened 2021-11-04
Soft 534 mci fan temp toggle

ShiCheng-Lu:


projects/mci/src/mci_broadcast.c
```diff
@@ -30,6 +29,7 @@ static const uint32_t s_offset_lookup[NUM_MOTOR_CONTROLLER_BROADCAST_MEASUREMENT
   [MOTOR_CONTROLLER_BROADCAST_DSP_TEMP] = WAVESCULPTOR_MEASUREMENT_ID_DSP_BOARD_TEMPERATURE,
 };
 
+static bool should_turn_off_fan = true;
```
mitchellostler, line 32: It looks like this isn't used anywhere? if so, you can rid of it.
If it is needed, it should be prefixed with s_.

projects/mci/src/mci_broadcast.c
```diff
@@ -80,6 +80,27 @@ static void prv_broadcast_dsp_temp(MotorControllerBroadcastStorage *storage) {
                                (uint32_t)measurements[RIGHT_MOTOR_CONTROLLER].dsp_temp_c);
 }
 
+static void prv_temperature_fan_toggle(MotorControllerBroadcastStorage *storage) {
```
mitchellostler, line 83: I think that it might be better if the fan activation check (overtemp conditions) is handled in the functions where the data itself is processed (ie. `prv_handle_dsp_temp_rx` and `prv_handle_motor_sink_temp_rx`). I think this helps us to better track the cause and effect of data we're reading, and activate the fan right away as we reach the temp condition. 

The deactivation logic makes sense to have in this function, called periodically, as you need to know that all are below the off threshold to make the change

That being said, this solution works fine, and if there are other variables, such as this makes testing easier, then this is fine. Just something to think about, don't worry about changing

mitchellostler, line 83: Also, 'toggle' implies that the function is guaranteed to switch the state. Something like prv_update_fan_state might be better

mitchellostler (changes requested):
Looks good Shicheng. Only a few small things, mainly checking if you need 'should_turn_off_fan'. 

Also, really impressed with your tests, they look very well done

mitchellostler (approved):
Looks Good 👍

PR #472, opened 2021-11-04
Fix github workflow MutableSet error

ShiCheng-Lu:
pin python vertion to 3.9

this should fix the pipline as MutableSet is no longer importable from collections.
This is a issue with pylint using deprecated imports.


ryandancy (approved):
Thanks ShiCheng!

PR #473, opened 2021-11-10
Soft 560 jump to app client script

BENJIGAO:
Main concern is whether the script handles response datagrams correctly. A DatagramListener is initialized and connected to CAN, but during testing, only the jump to application datagram is handled by the callback (no instances of response datagrams) as far as I can tell.


projects/bootloader/scripts/test_jump_application.py
```diff
@@ -0,0 +1,54 @@
...
+            node_ids=TEST_NODES,
+            data=bytearray(TEST_DATA))
+        sender.send(message)
+
+        timeout = time.time() + 10
+        listener_message = listener.get_message()
+        while listener_message is not None:
+            if time.time() > timeout:
+                break
+            listener_message = listener.get_message()
```
ryandancy, line 43: The purpose of the unit test is to test that the code you wrote works (i.e. the `jump_to_application` function). To do this, you'll need to call your function. Specifically, what I'd recommend is setting up your sender, listener, and notifier as you've done, then calling `jump_to_application` with the test nodes and listener. Then in the callback, you should assert that the datagram type is `TEST_DATAGRAM_TYPE_ID` and respond with the appropriate response datagrams, to simulate controller boards responding to the datagram.

Finally, once `jump_to_application` returns, you should verify that it returned the correct value. You should also check that the `callback` was called! (You can do this by setting a variable from `callback` and checking it in the test function - could be easiest to define `callback` inside the test function.)

ryandancy (changes requested):
Awesome Ben! Sorry for the delayed review, the core of it looks good.

projects/bootloader/scripts/jump_application.py
```diff
@@ -0,0 +1,40 @@
...
+        data=DATAGRAM_DATA)
+    sender.send(message)
+
+    timeout = time.time() + 10
+    listener_message = listener.get_message()
+    # Retrieves response datagrams until there are none or timeout
+    while listener_message is not None:
+        if time.time() > timeout:
+            break
+        listener_message = listener.get_message()
```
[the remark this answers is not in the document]
ryandancy: For getting the `board_id` into the callback, also take a look at Bafran's PR, looks like he's already implemented it. Might be worth merging his branch into yours for those changes.

projects/bootloader/scripts/test_jump_application.py
ryandancy: Also look at Bafran's PR for sending a datagram from an arbitrary board ID, he's got that implemented too (with the `sender_id` parameter of `Datagram.send`).

projects/bootloader/scripts/jump_application.py
```diff
@@ -0,0 +1,40 @@
...
+DATAGRAM_DATA = []
+
+
+def jump_to_application(node_ids):
+    """This function sends a jump-to-application datagram to specified boards,
+    receives a response, then returns the status"""
+    sender = DatagramSender(receive_own_messages=True)
+    listener = DatagramListener(trigger_callback)
+
+    can.Notifier(sender.bus, [listener])
```
ryandancy, line 20: One other thing: you'll probably want to assign this to a variable to make sure it doesn't get garbage-collected.

BENJIGAO:
I made changes based on comments for both my function and test file. Let me know if there's anything I should change!

projects/bootloader/scripts/test_jump_application.py
```diff
@@ -0,0 +1,68 @@
...
+        """Test the jump to application feature"""
+        self.callback_triggered = False
+        self.sender = DatagramSender(channel=TEST_CHANNEL, receive_own_messages=True)
+        listener = DatagramListener(self.callback)
+
+        notifier = can.Notifier(self.sender.bus, [listener])
+
+        statuses = jump_to_application(TEST_NODE_IDS, self.sender)
+
+        self.assertEqual(self.callback_triggered, True)
```
ryandancy, line 46: This assertion is failing on my machine. Is it working on yours? If so, let's have a call to see why it's inconsistent; if not, run this test with `make pytest PLATFORM=x86 PROJECT=bootloader TEST=jump_application` and debug why.

ryandancy (changes requested):
Sorry for the super delayed feedback, just a few more comments!

ryandancy (approved):

PR #474, opened 2021-11-11
Soft 561 flash application code operation

Bafran:


projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+# Import protobuf
+from protogen import flash_application_code_pb2 as pb2
+from can_datagram import Datagram, DatagramListener, DatagramSender
+import can
+import time
+import os
+import sys
+currentdir = os.path.dirname(os.path.realpath(__file__))
+parentdir = os.path.dirname(currentdir)
+sys.path.append(parentdir)
```
ryandancy, line 13: It's unfortunate due to our hacky project setup, but you'll need to put these lines above the `from protogen import ...` line for python to resolve the imports. (You can test with `make pytest PROJECT=bootloader TEST=flash_application_code`, and you can ignore the pylint errors it'll generate line-by-line.)

projects/bootloader/scripts/test_flash_application_code.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+TEST_PROTOCOL_VERSION = 1
+TEST_DATAGRAM_TYPE_ID = 1
+TEST_NODES = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+TEST_DATA = [3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8, 9, 7, 9, 3, 2, 3, 8, 4, 6, 2, 6, 4, 3, 3]
+TEST_CHANNEL = "vcan0"
+
+
+class TestFlashApplication(unittest.TestCase):
```
ryandancy, line 27: The purpose of a unit test is to test the actual production code, meaning your tests have to use the `FlashApplicationCode` class from the flash_application_code.py. Otherwise, it's not actually verifying that what your code does is correct - you could change the production code however you want and the tests would still pass!

You'll have to find a way to mock the responses you need with the existing code (you might have to modify it to make it more testable). One way to do this would be to have `receive_own_messages` set to True in the test and have your test function instantiate another DatagramListener and DatagramSender which provides the mock messages.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+        message = Datagram(
+            datagram_type_id=id,
+            node_ids=self.board_ids,
+            data=bytearray(data))
+        return message
+
+    def flash_protobuf(self):
+        """Flash Protobuf"""
+
+        message = self.create_message(8, pb2.DESCRIPTOR.serialized_pb)
```
ryandancy, line 49: Looks like you're sending a serialized description of the protobuf format itself here. Instead what you'll need to do is instantiate the protobuf and fill out its fields, which you can find in `flash_application_code.proto`:
```py
pb = pb2.FlashApplicationCode()
pb.name = ...
pb.git_version = ...
pb.application_crc = ...
pb.size = ...

# then to serialize:
serialized = pb.SerializeToString()
```
See https://developers.google.com/protocol-buffers/docs/pythontutorial#the-protocol-buffer-api. `name` and `git_version` will have to be passed into your class (that's the name of the project and the git commit hash it's from respectively). `size` is the size in bytes of the application code data, and `application_crc` is a hash that will have to be computed specially - talk to Faraz about how to compute it.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+        self.recv_boards = set()
+        while not set(self.board_ids) == self.recv_boards:
+            self.recv_boards = set(self.listener.datagram_messages.keys())
+            if time.time() > timeout:
+                self.status_code = 2
+                break
+
+        # Return if status code fails
+        if self.status_code != 0:
+            return
```
ryandancy, line 63: Did you mean to return something here? I think it would be more Pythonic to raise an exception in the timeout case and then not return anything for success.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+    def flash_protobuf(self):
+        """Flash Protobuf"""
+
+        message = self.create_message(8, pb2.DESCRIPTOR.serialized_pb)
+        self.sender.send(message)
+
+        self.status_code = 0
+        timeout = time.time() + 5
+        self.recv_boards = set()
+        while not set(self.board_ids) == self.recv_boards:
```
ryandancy, line 55: Use `a != b` instead of `not a == b`.

projects/bootloader/scripts/flash_application_code.py
ryandancy, line 49: Instead of using magic numbers like 8 and 9 for the ids, factor them out into constants with descriptive names. (In fact, it may be best to gather all the datagram-id constants into an enum in a separate file.)

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+
+            message_chunk = self.create_message(9, chunk)
+            self.sender.send(message_chunk)
+
+            self.status_code = 0
+            timeout = time.time() + 5
+            while not set(self.board_ids) == self.recv_boards:
+                self.recv_boards = set(self.listener.datagram_messages.keys())
+                if time.time() > timeout:
+                    self.status_code = 2
```
ryandancy, line 83: Similar, you raise an exception in the timeout case rather than relying on a `status_code` to be checked.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+                    return
+        return
```
ryandancy, line 85: This return is useless?

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,85 @@
...
+import can
+import time
+import os
+import sys
+currentdir = os.path.dirname(os.path.realpath(__file__))
+parentdir = os.path.dirname(currentdir)
+sys.path.append(parentdir)
+
+
+class FlashApplication():
```
ryandancy, line 16: It's great you're doing this in an object-oriented style! For consistency with other operations, please add a top-level `flash_application_code` function that wraps the whole protocol (which would mean instantiating FlashApplication and calling flash_protobuf and flash_application).

ryandancy (changes requested):
Nice job!

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,153 @@
...
+        return message
+
+    def flash_protobuf(self):
+        """Flash Protobuf"""
+
+        # Construct and serialize protobuf
+        protobuf = pb2.FlashApplicationCode()
+        protobuf.name = self.name
+        protobuf.git_version = bytes(self.git_version)
+        protobuf.application_crc = zl.crc32(self.app_data)
```
ryandancy, line 96: Did you talk to Faraz and/or ShiCheng about how to compute this? We have a somewhat convoluted custom way of computing the crc32 for the application code because of constraints on the C side - ask Faraz or ShiCheng about how to do it. (Unless they told you to do it this way, in which case never mind.)

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,153 @@
...
+        self.listener = DatagramListener(self.listener_callback)
+        self.notifier = can.Notifier(self.sender.bus, [self.listener])
+
+        self.recv_boards = []
+        self.status_code = 0
+
+    def listener_callback(self, datagram: Datagram, board_id):
+        """Check whether the datagram has an OK status code"""
+        # Ignores non-zero codes from the client
+        if datagram._data != bytearray(0) and board_id != 0:
```
ryandancy, line 61: `bytearray(0)` creates an empty bytearray (length 0) - you probably want `bytearray([0])` to check for a length-1 bytearray containing a 0.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Send protobuf
+        message = self.create_message(OUTGOING_PROTOBUF_ID, serialized)
+        self.sender.send(message)
+
+        self.status_code = 0
+        timeout = time.time() + 5
+        self.recv_boards = set()
+        while set(self.board_ids) != self.recv_boards:
+            self.recv_boards = set(self.listener.datagram_messages.keys())
```
ryandancy, line 108: It's valid but tricky that you're relying on the listener's `datagram_messages` field to get updated when it receives a message - you might want to put a comment to that effect somewhere.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,153 @@
...
+
+        message_chunk = self.create_message(OUTGOING_DATA_ID, chunk)
+        self.sender.send(message_chunk)
+
+        self.status_code = 0
+        timeout = time.time() + 5
+        while set(self.board_ids) != self.recv_boards:
+            self.recv_boards = set(self.listener.datagram_messages.keys())
+            if time.time() > timeout:
+                self.status_code = 2
```
ryandancy, line 127: `2` is a magic number here which is hard to understand - factor this and the other status code constants out into constants like `STATUS_CODE_TIMEOUT = 2`.

projects/bootloader/scripts/test_flash_application_code.py
```diff
@@ -0,0 +1,54 @@
...
+        flash = FlashApplication(
+            channel=TEST_CHANNEL,
+            board_ids=TEST_NODES,
+            application_code_data=bytes(TEST_DATA),
+            name=TEST_NAME,
+            git_version=TEST_GIT_VERSION
+        )
+
+        # Edit DatagramSender to receive own messages for testing
+        flash.sender = DatagramSender(channel=TEST_CHANNEL, receive_own_messages=True)
```
ryandancy, line 32: It would be more elegant to take `receive_own_messages` as an optional keyword argument in `FlashApplication.__init__`, then you don't have to edit it like this.

projects/bootloader/scripts/flash_application_code.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # pylint: disable=too-many-instance-attributes
+    def __init__(self, **kwargs):
+        self._check_kwargs(**kwargs)
+
+        self.channel = kwargs["channel"]
+        self.app_data = kwargs["application_code_data"]
+        self.board_ids = kwargs["board_ids"]
+        self.name = kwargs["name"]
+        self.git_version = kwargs["git_version"]
```
ryandancy, line 49: It might be cleaner to just pass the arguments as regular positional and/or keyword arguments instead of parsing through `**kwargs`.

ryandancy (changes requested):
Looks like it works great! There's a few nitpicks, but really it's just the one correctness thing with bytearray and then we're good to go.

projects/bootloader/scripts/flash_application_code.py
Bafran: I asked Faraz about it and he said zlib would work

projects/bootloader/scripts/flash_application_code.py
Bafran: I changed it to use the board_id received by the datagram listener since I'd made that change earlier anyway

ryandancy (approved):

PR #476, opened 2021-11-16
[SOFT-413] Fix sizes in `make temp-bootloader-write`

ryandancy:
We updated the bootloader size from 16K to 32K at some point but neglected to update `make temp-bootloader-write`, which would cause impossible-to-debug linker script issues. Updated the size to fix it.

@mitchellostler, @vaaranan-y, and anyone else testing bootloader onsite this weekend: you'll have to make sure this PR is on your branch before testing. Otherwise, you should be able to do `make temp-bootloader-write PROJECT=leds` to flash the `leds` project on top of the bootloader (although the config will *not* be populated). Also, if you're testing the flash-application-code operation, you might have to dig into the linker scripts to generate a properly offsetted application-only binary (though it *should* just be as simple as using the .bin file for the particular application generated during `make temp-bootloader-write`).


mitchellostler (approved):
Looks good, sorry I forgot to get to this

PR #477, opened 2021-11-24
Upgrade DBC releaser script from ubuntu-16.04

abdullahs26:


ryandancy:
This seems to be on the main CI script? We need the changes on `make_dbc_release.yml`.

PR #479, opened 2021-11-28
Fw 104 Hw

BENJIGAO:


projects/fw_104_hw/src/p1p2.c
```diff
@@ -0,0 +1,117 @@
...
+  .mode = SPI_MODE_0,
+  .mosi = { .port = GPIO_PORT_B, .pin = 15 },
+  .miso = { .port = GPIO_PORT_B, .pin = 14 },
+  .sclk = { .port = GPIO_PORT_B, .pin = 13 },
+  .cs = { .port = GPIO_PORT_B, .pin = 12 },
+};
+
+// Confused on how to write to CANCTRL register, because unlike i2c,
+// there's no spi write-to-register function where an address is used
+int spi(void) {
```
mitchellostler, line 87: The way that it works with spi, is that all the information is contained in whatever message you are sending. That is, the operation you want to perform, along with where you want to perform it, and which data you want to perform it with. These values are outlined in mcp2515_defs.h, which we use for our mcp2515 driver. The MCP2515 provides shortcuts for some, but in this case, the operation would be {Write: 0x02, MCP_CANCTRL_REG: 0x0F, Value: 0x49}

projects/fw_104_hw/src/p1p2.c
```diff
@@ -0,0 +1,117 @@
...
+              (READ_I2C_PORT == I2C_PORT_2) ? 2 : 1);
+  } else {
+    LOG_DEBUG("Read failed: status code %d\n", status);
+  }
+  return 0;
+}
+
+#include "spi.h"
+
+static const uint8_t canctrl_info_tx[] = { 0b01001001 };
```
mitchellostler, line 69: Always use hex, not binary, makes things much easier to read

mitchellostler (commented):
Well done, just a few small points to mention

mitchellostler:
I'm going to go ahead and close this since it's been open so long. Feel free to take a look at the comments I listed though

PR #483, opened 2022-01-13
Makefile changes to check MISRA compliance

CheranMahalingam:
To use the tool run,
```
make install_misra
```
to install the dependencies and run,
```
make misra PROJECT=X
```
to check compliance for a project

>Note: The MISRA addon for cppcheck was not working for cppcheck 1.82 which is the latest version for our machine. So I copied the scripts and threw them in the misra folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ codegen/system_can.dbc
 *.gcda
 *.gcno
 codecov/
+
+# The MISRA script generates these
+*.dump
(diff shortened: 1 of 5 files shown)


ryandancy:
Code looks fine, my concern is copyright - I'm fairly certain the `misra.txt` file (containing the rule texts) is proprietary to MISRA and we need a license from MISRA to distribute it. In fact, it may be wise to remove it from this PR (and the git history - do `git reset --soft HEAD~2` to undo the last two commits but leave the changes, remove the text, and then force push). My suggestion is to replace the actual rule texts with the rule number, and then we'll try and get a license from MISRA as a sponsorship, but we'll at least have the rule numbers.

ryandancy (approved):

ryandancy:
It does seem the script has a lot of false positives - we can talk about it at standup tomorrow but we'll probably end up merging this and then later going through and fixing the false positives.

PR #485, opened 2022-01-18
[SOFT-999] FW_103 HW

AdrianJendo:


projects/fw_103/src/main.c
```diff
@@ -0,0 +1,43 @@
...
+
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+void readADC(const GpioAddress *btn_addr_ptr, void *context) {
```
mitchellostler, line 11: For functions that are only used in the scope of the file, we make them private (ie declare them static). We also prefix the function name with `prv_` to let everyone know that they should only be used here. 

Nit: We use the convention of lower case in function prototypes and variables with underscores (ie `prv_read_adc` instead.
Structs and types generally use Camel case (GpioAddress, etc)

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,43 @@
...
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+  gpio_it_init();
+
+  // Potentiometer
+  GpioAddress pot_addr = { .port = GPIO_PORT_A, .pin = 6 };
+  GpioSettings pot_settings = { GPIO_DIR_IN, GPIO_STATE_LOW, GPIO_RES_NONE, GPIO_ALTFN_ANALOG };
```
mitchellostler, line 26: Nit: usually do each parameter defined in the struct as a new line

mitchellostler (changes requested):
A couple of small changes, mostly superficial, but otherwise well done. Also not sure if you noticed but CI didn't pass. It doesn't seem to be an issue with your code, and it's a bit buggy sometimes, so I gave it a rerun and we'll see what comes of it

PR #486, opened 2022-01-20
[SOFT-999] Implement periodic counter using soft timers - FW102 Homework

AbhinavA10:


projects/fw_102/src/main.c
```diff
@@ -0,0 +1,42 @@
+#include <stdint.h>  // for integer types
+
+#include "interrupt.h"   // interrupts are required for soft timers
+#include "log.h"         // for printing
+#include "soft_timer.h"  // for soft timers
+#include "wait.h"        // for wait function
+
+#define COUNTER_PERIOD_MS 500  // ms
```
mitchellostler, line 8: Nit: you don't need to specify the units in a comment since it's already in the constant name

projects/fw_102/src/main.c
```diff
@@ -0,0 +1,42 @@
+#include <stdint.h>  // for integer types
+
+#include "interrupt.h"   // interrupts are required for soft timers
```
mitchellostler, line 3: Good work with the comments, but the ones for the libraries are a bit overkill. If they are for your own reference in the homework that's totally fine, but not something we would do in production code 🙂

mitchellostler (commented):
a couple of nitpick comments, but otherwise well done! I am marking you complete for FW102, and closing this PR

PR #487, opened 2022-01-21
[SOFT-999] fw102 - nourish-cherish

nourishnew:


projects/fw_102/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_DURATION 500
+
+typedef struct {
+  uint8_t counter_a;
+  uint8_t counter_b;
+  int turn;
```
mitchellostler, line 12: You don't need to have this variable in the struct here. Since you are incrementing counter_a on every timer callback, you can use it's value (the exact same way as you are using 'turn') to determine whether or not to increment counter_b

projects/fw_102/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
mitchellostler, line 15: Using the change I mentioned above, you can do this section in quite a few less lines

projects/fw_102/src/main.c
```diff
@@ -0,0 +1,38 @@
...
+    LOG_DEBUG("Counter B: %d\n", counters->counter_b);
+  }
+  counters->turn++;
+  soft_timer_start_millis(COUNTER_DURATION, prv_timer_callback, counters, NULL);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  Counters c = { 0, 0, 0 };
```
mitchellostler, line 33: This initialization to zero can also be done by just specifying `c = { 0 }`. For structs, it interprets this implicitly as initialize everything to zero

projects/fw_102/src/main.c
mitchellostler, line 15: also, if a function is only going to be used in the file it is declared, you should declare it as `static void`. You've got the `prv_` naming convention correct though 🙂

mitchellostler (changes requested):
Looks pretty good, just a few changes to make it a bit cleaner. Good work, and sorry it took so long to get a review

projects/fw_102/src/main.c
```diff
@@ -0,0 +1,38 @@
+
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_DURATION 500
```
mitchellostler, line 7: Constants should have units if applicable (ie `DURATION_MS` for milliseconds)

mitchellostler (commented):
Well Done, you're good to go!

PR #488, opened 2022-01-22
Completed Firmware 102 homework, request for code review

nehemiah-negussie:
First test commit.

### DIFF
--- a/projects/timer_hw/src/main.c
+++ b/projects/timer_hw/src/main.c
@@ -1,0 +1,61 @@
+#include "interrupt.h"
+#include "soft_timer.h"
+#include "log.h"
+#include "wait.h"
+
+#include <stdlib.h>
+#include <stdint.h>
+
+#define COUNTER_PERIOD_MS 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback_2(SoftTimerId timer_id, void *context);
+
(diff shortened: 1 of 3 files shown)


projects/timer_hw/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+#include "wait.h"
+
+#define COUNTER_PERIOD_MS 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback_2(SoftTimerId timer_id, void *context);
```
mitchellostler, line 16: Don't worry about forward declaring your functions, just declare them where you write the function bodies.
This line can go

projects/timer_hw/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
mitchellostler, line 18: You can (and should) do all the soft-timer functionality in one timer callback function instead of 2, see if you can figure it out, but if not feel free to ask for help 🙂

mitchellostler: Additionally, all private functions in a file should be declared as static, ie `static void prv_*`

mitchellostler (changes requested):
Looks really well done! There is a simpler design however, that uses only one timer callback function. See if you can figure it out, but if not feel free to ask for help

PR #489, opened 2022-01-23
Completed fw 103 homework

ShahriarAhnaf:
please review for accuracy.

### DIFF
--- a/projects/soft999_ahnaf_shahriar_fw_103/src/main.c
+++ b/projects/soft999_ahnaf_shahriar_fw_103/src/main.c
@@ -1,0 +1,61 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static GpioAddress potentiometer_addr = {
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+// function to call when the button gets hit
+static void button_interrupt_handler(const GpioAddress *addy, void *context) {
+  uint16_t potentiometer_data = 0;
+  if (adc_read_raw_pin(potentiometer_addr, &potentiometer_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("potentiometer data: %d\n", potentiometer_data);
+  } else {
+    LOG_DEBUG("Failed to read potentiomter data");
+  }
+  delay_ms(200);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+
+  GpioAddress btn_addr = {
+    .port = GPIO_PORT_B,
+    .pin = 2,
+  };
+
+  GpioSettings pot_settings = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_NONE,
+    .alt_function = GPIO_ALTFN_ANALOG,
+  };
+  InterruptSettings interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,       //
+    .priority = INTERRUPT_PRIORITY_NORMAL,  //
+  };
+
+  gpio_init_pin(&potentiometer_addr, &pot_settings);
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(potentiometer_addr, true);
+  // initialize the button interrupt.
+  gpio_it_register_interrupt(&btn_addr,            // address for button
+                             &interrupt_settings,  // what type of interrupt
+                             INTERRUPT_EDGE_FALLING,
+                             &button_interrupt_handler,  // what function to call
+                             &potentiometer_addr         // what to send the function
+    );
+  while (true) {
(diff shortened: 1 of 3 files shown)


projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+    LOG_DEBUG("Failed to read potentiomter data");
+  }
+  delay_ms(200);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+  gpio_init();
+
```
ShiCheng-Lu, line 28: gpio_it_init() needs to be called to enable interrupts from gpio pins

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+};
+// function to call when the button gets hit
+static void button_interrupt_handler(const GpioAddress *addy, void *context) {
+  uint16_t potentiometer_data = 0;
+  if (adc_read_raw_pin(potentiometer_addr, &potentiometer_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("potentiometer data: %d\n", potentiometer_data);
+  } else {
+    LOG_DEBUG("Failed to read potentiomter data");
+  }
+  delay_ms(200);
```
ShiCheng-Lu, line 21: what is this delay for?

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_NONE,
+    .alt_function = GPIO_ALTFN_ANALOG,
+  };
+  InterruptSettings interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,       //
+    .priority = INTERRUPT_PRIORITY_NORMAL,  //
+  };
+
+  gpio_init_pin(&potentiometer_addr, &pot_settings);
```
ShiCheng-Lu, line 45: you also need to init the button pin to use it

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static GpioAddress potentiometer_addr = {
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+// function to call when the button gets hit
+static void button_interrupt_handler(const GpioAddress *addy, void *context) {
```
ShiCheng-Lu, line 14: we prefix static methods with `prv_` and global static variables with `s_`.

(addy is an interesting variable name, typo?)

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(potentiometer_addr, true);
+  // initialize the button interrupt.
+  gpio_it_register_interrupt(&btn_addr,            // address for button
+                             &interrupt_settings,  // what type of interrupt
+                             INTERRUPT_EDGE_FALLING,
+                             &button_interrupt_handler,  // what function to call
+                             &potentiometer_addr         // what to send the function
+    );
```
ShiCheng-Lu, line 55: here you are not using the `void * context` in your function anyway so its ok to pass in `NULL`. 

format is also complaining about this ending parenthesis, I would put it straight after the last argument.

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+  while (true) {
+    // let cpu run wild until interrupt.
```
ShiCheng-Lu, line 57: you can add a `wait()` from `wait.h` here to save on processing time.

projects/soft999_ahnaf_shahriar_fw_103/src/main.c
```diff
@@ -0,0 +1,61 @@
...
+    LOG_DEBUG("potentiometer data: %d\n", potentiometer_data);
+  } else {
+    LOG_DEBUG("Failed to read potentiomter data");
+  }
+  delay_ms(200);
+}
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
```
ShiCheng-Lu, line 26: I don't think soft timer is used in this project, you don't need to include and init it.

ShiCheng-Lu (commented):
There are 2 errors that would stop this project from doing what is intended, but apart from that, there are only minor problems. 

you should run formatting and linting before you push. If you've set up the github hooks, it should format and lint before every commit.

your branch name should be `soft_999_...` with a underscore.

you can move on to 104 after fixing the errors, and probably take on a ticket if you are interested.

mitchellostler (changes requested):
Take a look at and address Shicheng's requested changes when you get the chance then we can move you on to a task

PR #490, opened 2022-01-23
[SOFT-999] Complete fw103

AbhinavA10:
FW 103 Homework

# Part 1 - Schematics and datasheets scavenger hunt
## Datasheet scavenger hunt.

| Question      | Answer |
| ----------- | ----------- |
| What bit in what register should be set on the MCP2515 CAN controller to enable one-shot mode?      | `CANCTRL[3]` |
| In the MCP2515, what are the 3 bytes you’d send over spi if you wanted to write 0x17 to register 0x0f?   | Format is ` WRITE, ADDR_BYTE, DATA_BYTE`. This becomes `0x02 0x0f 0x17`        |
| What is the digital output code (16 bits) for the MCP3427 when the input voltage is 0 | `0000000000000000` |
| What is the first Opcode byte for to start a conversion on the  ADS1259 | In binary, `0000 100x` |
| What is the address (hex) for GPIO in the MCP23008? | `0x09` |
| What are the fault codes for the LTC6811 when a fault is detected in bits 15-12? | `0b1XXX` |

## Schematic scavenger hunt.

| Question      | Answer |
| ----------- | ----------- |
|On the MSXIV power select board, what port and pin should be set to what state to shut down the LTC4417? | `PB15`, pulled low |
|On the MSXIV pedal board, does the controller board talk to the ADC via SPI, CAN, I2C, or UART? | `I2C` |
|What communication protocol is used by the MSXIV BMS carrier to communicate with the AFE and current sense board? | `SPI` |
|What port and pin are used for the MSXIV CAN interrupt for the charger interface board? | `PA8` |
|What port and pin control the horn in the MSXIV steering wheel board? | `PB1` |
|What protocol is used by the MSXIV MCI to communicate with the MCP2515? What ports and pins does this use? | `SPI`, `PB` `pins 12-15` |

# Part 2 - Buttons and ADCs
Please see FW_103 project code in PR.

### DIFF
--- a/projects/fw_103/src/main.c
+++ b/projects/fw_103/src/main.c
@@ -1,0 +1,59 @@
+#include <stdint.h>  // for integer types
+
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"  // interrupts are required for soft timers
+#include "log.h"        // for printing
+#include "wait.h"       // for wait function
+
+GpioAddress adc_addr = {
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+
+static void prv_read_adc(const GpioAddress *btn_addr_ptr, void *context) {
+  uint16_t adc_data = 0;
+  if (adc_read_converted_pin(adc_addr, &adc_data) == STATUS_CODE_OK) {
+    LOG_DEBUG("ADC data: %d\n", adc_data);
+  } else {
+    LOG_DEBUG("Failed to read adc data");
+  }
+}
+
+int main(void) {
+  gpio_init();
+  gpio_it_init();
+  interrupt_init();
+
+  GpioAddress btn_addr = {
+    .port = GPIO_PORT_B,
+    .pin = 2,
+  };
+  GpioSettings adc_gpio_settings = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_NONE,
+    .alt_function = GPIO_ALTFN_ANALOG,
+  };
+  GpioSettings btn_gpio_settings = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_PULLDOWN,
+  };
+
+  gpio_init_pin(&adc_addr, &adc_gpio_settings);
+  gpio_init_pin(&btn_addr, &btn_gpio_settings);
+  adc_init(ADC_MODE_SINGLE);
+
+  adc_set_channel_pin(adc_addr, true);
+
+  InterruptSettings settings = { .priority = INTERRUPT_PRIORITY_NORMAL,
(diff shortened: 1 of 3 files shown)


projects/fw_103/src/main.c
```diff
@@ -0,0 +1,59 @@
+#include <stdint.h>  // for integer types
+
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"  // interrupts are required for soft timers
+#include "log.h"        // for printing
+#include "wait.h"       // for wait function
+
+GpioAddress adc_addr = {
```
ShiCheng-Lu, line 10: when we have global variables like this that should only be visible within one file, we define them as static and prefix the variable name with s_{name}.

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+    LOG_DEBUG("ADC data: %d\n", adc_data);
+  } else {
+    LOG_DEBUG("Failed to read adc data");
+  }
+}
+
+int main(void) {
+  gpio_init();
+  gpio_it_init();
+  interrupt_init();
```
ShiCheng-Lu, line 27: I think interrupt needs to be initialized before gpio interrupt

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,59 @@
...
+  GpioSettings adc_gpio_settings = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_NONE,
+    .alt_function = GPIO_ALTFN_ANALOG,
+  };
+  GpioSettings btn_gpio_settings = {
+    .direction = GPIO_DIR_IN,
+    .state = GPIO_STATE_LOW,
+    .resistor = GPIO_RES_PULLDOWN,
```
ShiCheng-Lu, line 42: I think this should be non instead of pulldown? but tbh, I'm not exactly sure myself.

ShiCheng-Lu (commented):
Looks good, just some small changes. You work on 104 home work or take on a task if you'd like, attend a standup and ask for a task or msg Mitchell or Ryan. 

for scavenger hunt, q2, you switched up the addr and data. it should be 0x02 0x0f 0x17

(was definitely better than my 103)

projects/fw_103/src/main.c
AbhinavA10: Right, thanks

projects/fw_103/src/main.c
AbhinavA10: I did not have a reference schematic, and so assumed an internal pull-down resistor. But `None` would also work for a external pull-down. I'll update this.

projects/fw_103/src/main.c
AbhinavA10: Noted, thank you

AbhinavA10:
Thank you for reviewing this. I have fixed the comments you outlined.

ShiCheng-Lu:
looks good!

PR #491, opened 2022-01-27
[SOFT-999] fw 102 homework matthew Keller

m6keller:
homework for firmware 102

### DIFF
--- a/projects/fw102_example/src/main.c
+++ b/projects/fw102_example/src/main.c
@@ -1,0 +1,51 @@
+//
+// Created by Matthew Keller on 2022-01-22.
+//
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "delay.h"       // For real-time delays.
+#include "interrupt.h"   // For interrupting stuff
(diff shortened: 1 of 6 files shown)


projects/fw102_example/src/main.c
```diff
@@ -0,0 +1,51 @@
...
+#include "misc.h"  // Various helper functions/macros.
+
+#define COUNTER_LENGTH 500
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void increment_a(SoftTimerId timer_id, void *context) {
```
mitchellostler, line 24: Functions that are only used in one file should use the static label and be prefixed with prv_, (ie. `static void prv_increment_a`). This lets us know that it is limited to the scope of the file that it is in

mitchellostler (changes requested):
Looks good! You got the logic down. Just a few formatting details. 
Also, I noticed that the build is failing, and this can be fixed by running `make format && make lint` on your branch, and committing the changes. Let me know if you have any questions :)

PR #493, opened 2022-01-29
[SOFT-999] 102 hmwrk

andrewseo2001:
Finished my 102 hmwrk


projects/102_hmwrk/src/main.c
```diff
@@ -0,0 +1,40 @@
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+#define COUNTER_A_PERIOD 500
+#define COUNTER_B_PERIOD 1000
```
mitchellostler, line 7: Since you're not using this you can get rid of it

projects/debug_test/src/main.c
```diff
@@ -0,0 +1,69 @@
+#include <stdint.h>
```
mitchellostler, line 1: Not to sure what's going on with these projects? this pull request should just be the FW 102 content

mitchellostler (changes requested):
Looks good, seems like you got all the logic down, just one nitpick. 
Also, if you could clean up your PR so that only the files from FW102 are on it that would be great. 
Last thing, if you look on github right now, the build is failing. If you run `make format && make lint` and commit the changes, this should fix the issue

mitchellostler:
Looks like your build is passing, good work!

PR #495, opened 2022-02-01
Soft 581 flash empty config with bootloader

123Logan321:


projects/bootloader/src/config.c
```diff
@@ -54,8 +44,18 @@ StatusCode config_verify(void) {
...
   status_ok_or_return(persist_ctrl_periodic(&s_config_2_persist, false));
 
-  uint32_t config_1_check_crc = prv_compute_crc32(&s_config_1_blob);
-  uint32_t config_2_check_crc = prv_compute_crc32(&s_config_2_blob);
+  uint32_t config_1_check_crc = 0;
+  uint32_t config_2_check_crc = 0;
+
+  // Check for if the bootloader config is still zero. If it is not, then calculate the crc of the
+  // config to fill in the static configs
+  if (memcmp(&s_config_1_blob, &s_config_empty, sizeof(BootloaderConfig)) != 0) {
```
mitchellostler, line 52: I would say you can just declare `BootloaderConfig config_empty = { 0 }` here, so it's clear that you're just using it to compare. (also just a tip, if you're declaring something static, then you don't need to initialize it to zero, since static zeroes it by default)

mitchellostler (changes requested):
Just the one note, otherwise it looks good! I'll add ryan too in case there's anything he wants to add

projects/bootloader/src/config.c
```diff
@@ -11,41 +11,31 @@ static BootloaderConfig s_config_1_blob = { 0 };
 static PersistStorage s_config_2_persist = { 0 };
 static BootloaderConfig s_config_2_blob = { 0 };
```
ryandancy, line 12: Initialize the `BootloaderConfig` blobs to sensible defaults of the parameters - I'd say these should be the values of the fields:
- `controller_board_id` should be `63` so that it's easy for us to tell that it's invalid, since 63 is the maximum controller board ID
- `controller_board_name` should be the empty string (or maybe `"<unset>"`)
- `project_present` should be false
- `project_name`, `project_info`, and `git_version` should be the empty string
- `application_crc32` and `application_size` should be zero

Most of these are the same as the defaults, but it's good to specify them explicitly, and we don't want to use the default controller board ID of 0 since that might cause errors (ID 0 is reserved for the client script). It might in fact be a good idea to factor out the default controller board ID into a `#define` with a comment so people know why 63 was chosen.

The `crc32` field is what I was talking about with how we should dynamically calculate it. At the very start of `config_init`, just call `prv_compute_crc32` on both of the config blobs and set their crc32 fields to the result. That will be more robust than handling the default config after having loaded the config - it's best to have a correct config regardless and then we don't need special cases.

projects/bootloader/test/test_config.c
```diff
@@ -103,7 +103,7 @@ void setup_test(void) {
   s_test_config_2.crc32 = crc32_arr((uint8_t *)&s_test_config_2, sizeof(BootloaderConfig));
   persist_commit(&s_test_persist_storage_2);
 
-  TEST_ASSERT_OK(config_verify());
+  TEST_ASSERT_OK(config_init());
 }
 void teardown_test(void) {}
```
ryandancy, line 109: I would like one test for the default config changes - essentially, use the `flash` module to clear the two flash pages `BOOTLOADER_CONFIG_PAGE_1_FLASH_PAGE` and `BOOTLOADER_CONFIG_PAGE_2_FLASH_PAGE`, call `config_init()`, and verify that a) the config you then get from `config_get()` is the default config and b) the default config was correctly recorded on flash. You can do this by using the `persist` module to read from flash, or you can just call `config_init()` again and make sure that the config from `config_get()` is still the default, because it will read the config pages from flash.

ryandancy (changes requested):
Great job Logan, just a few tweaks.

ryandancy (changes requested):
Looks great! Just one minor nitpick, then you'll be good to go.

ryandancy (approved):

mitchellostler (approved):

PR #501, opened 2022-02-12
[SOFT-999] button and adc hmwrk

andrewseo2001:
First draft of button and adc assignment for 103 hmwrk

### DIFF
--- a/projects/103_hmwrk/src/main.c
+++ b/projects/103_hmwrk/src/main.c
@@ -1,0 +1,55 @@
+#include "adc.h"
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static GpioAddress adc_addr = {
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+
+static GpioAddress button_addr = {
+  .port = GPIO_PORT_B,
+  .pin = 2,
+};
+
+static InterruptSettings s_interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+void read_from_adc(const GpioAddress *address, void *context) {
+  uint16_t data = 0;
+  if (adc_read_raw_pin(*(GpioAddress *)context, &data) == STATUS_CODE_OK) {
+    LOG_DEBUG("pin A6 data: %d\n", data);
+  } else {
+    LOG_DEBUG("Failed to read data from pin A6");
+  }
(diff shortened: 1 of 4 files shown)


projects/103_hmwrk/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  gpio_init_pin(&adc_addr, &adc_read_settings);
+  gpio_it_register_interrupt(&button_addr, &s_interrupt_settings, INTERRUPT_EDGE_FALLING,
+                             read_from_adc, &adc_addr);
+
+  adc_init(ADC_MODE_SINGLE);
+  adc_set_channel_pin(adc_addr, true);
+
+  while (true) {
```
mitchellostler, line 53: Nit: good to use a wait() function call here

projects/103_hmwrk/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+};
+
+static InterruptSettings s_interrupt_settings = {
+  .type = INTERRUPT_TYPE_INTERRUPT,
+  .priority = INTERRUPT_PRIORITY_NORMAL,
+};
+
+void read_from_adc(const GpioAddress *address, void *context) {
+  uint16_t data = 0;
+  if (adc_read_raw_pin(*(GpioAddress *)context, &data) == STATUS_CODE_OK) {
```
mitchellostler, line 26: This cast is a bit awkward. The format we generally use is at the top of the function:
```
GpioAddress *address = (GpioAddress*)context;
```
and then access all fields from the pointer, ie. `address->port`
This isn't a functional problem, more of a best coding practices.

projects/103_hmwrk/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+static GpioAddress adc_addr = {
+  .port = GPIO_PORT_A,
+  .pin = 6,
+};
+
+static GpioAddress button_addr = {
```
mitchellostler, line 14: Nit: all static variables should be prefixed with `s_`. Alternatively, the button and interrupts could be declared in main, since this is the only place you use them.

Also, since they won't be changed it is good practice to declare them as `const`.

mitchellostler (commented):
Overall your code looks really good, and I'm sure it would work fine, so good work getting this done. There's a couple of things that I mentioned that would be good to take a look at, but you can choose whether or not you want to fix them. I am just waiting on getting access to the scavenger hunt answer key, and then I will give you feedback on that 😁

mitchellostler (commented):
Scavenger hunt looks good! you are ready to take on a ticket!

PR #502, opened 2022-02-12
[SOFT-999] Did Firmware 102 homework.

laure-nn:
My firmware 102 homework.

### DIFF
--- a/projects/fw_102_homework/src/main.c
+++ b/projects/fw_102_homework/src/main.c
@@ -1,0 +1,39 @@
+#include <stdint.h>
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
+  Counters *storage = context;
+
+  ++storage->counter_a;
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  if (storage->counter_a % 2 == 0) {
+    ++storage->counter_b;
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b);
+  }
+
+  soft_timer_start_millis(500, prv_timer_callback, storage, NULL);
+}
+
+int main(void) {
+  interrupt_init();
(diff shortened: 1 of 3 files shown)


projects/fw_102_homework/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+  ++storage->counter_a;
+  LOG_DEBUG("Counter A: %i\n", storage->counter_a);
+
+  if (storage->counter_a % 2 == 0) {
+    ++storage->counter_b;
+    LOG_DEBUG("Counter B: %i\n", storage->counter_b);
+  }
+
+  soft_timer_start_millis(500, prv_timer_callback, storage, NULL);
```
ryandancy, line 23: Generally it's a good idea to factor out the `500` constant from here and the `soft_timer_start_millis` call below into a `#define`'d constant with an appropriate name, so we know what it is and so we don't have to change it in multiple places if we ever need to change it. It should be suffixed with the unit name (i.e. `_MS`) so we know what the units are.

projects/fw_102_homework/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include "log.h"
+#include "soft_timer.h"
+#include "wait.h"
+
+typedef struct Counters {
+  uint8_t counter_a;
+  uint8_t counter_b;
+} Counters;
+
+void prv_timer_callback(SoftTimerId timer_id, void *context) {
```
ryandancy, line 12: Since this is a private function, it should be marked `static`. That makes it private to a file, so other files can't see it. This is important in bigger projects because if you don't have `static`, then for example if some other file defined a `prv_timer_callback` function as well, then you might get linker errors.

projects/fw_102_homework/src/main.c
```diff
@@ -0,0 +1,39 @@
+#include <stdint.h>
+#include "interrupt.h"
```
ryandancy, line 2: Nitpick: there should be a blank line between the standard imports and the MS-specific imports.

ryandancy (changes requested):
Good job Lauren! Looks like it works, just a couple minor things to clean up before moving on to FW 103.

mitchellostler:
Looks good! Sorry for the delay, good work getting this done.

PR #504, opened 2022-03-02
[SOFT 999] FW 103 HW

m6keller:
FW 103 HW

### DIFF
--- a/projects/fw102_example/src/main.c
+++ b/projects/fw102_example/src/main.c
@@ -2,36 +2,29 @@
 // Created by Matthew Keller on 2022-01-22.
 //
 #include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
-
-#include "delay.h"       // For real-time delays.
 #include "interrupt.h"   // For interrupting stuff
 #include "log.h"         // For printing stuff.
+#include "misc.h"        // Various helper functions/macros.
 #include "soft_timer.h"  // Software timers for scheduling future events.
-#include "wait.h"  //
+#include "wait.h"        //
 
-#include "gpio.h"  // General Purpose I/O control.
-#include "misc.h"  // Various helper functions/macros.
-
-#define COUNTER_LENGTH 500
+#define COUNTER_LENGTH_MS 500
 
 typedef struct Counters {
   uint8_t counter_a;
   uint8_t counter_b;
 } Counters;
 
-void increment_a(SoftTimerId timer_id, void *context) {
+static void prv_increment_a(SoftTimerId timer_id, void *context) {
   Counters *counting = context;
   counting->counter_a++;
-  printf("Counter A: %d\n", counting->counter_a);
+  LOG_DEBUG("Counter A: %d\n", counting->counter_a);
 
-  if( counting->counter_a % 2 == 0 ) {
-      counting->counter_b ++;
-      printf("Counter B: %d\n", counting->counter_b);
-  }
+  if (counting->counter_a % 2 == 0) counting->counter_b++;
+  LOG_DEBUG("Counter B: %d\n", counting->counter_b);
+}
 
-  soft_timer_start_millis(COUNTER_LENGTH, increment_a, context, NULL);
+soft_timer_start_millis(COUNTER_LENGTH_MS, prv_increment_a, context, NULL);
 }
 
 int main(void) {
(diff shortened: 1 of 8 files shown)


projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,67 @@
...
+  InterruptSettings button_interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_HIGH,
+  };
+
+  adc_set_channel_pin(button_addr, true);
+
+  gpio_it_get_edge(&button_addr, INTERRUPT_EDGE_RISING);
+
+  uint16_t button_data = 0;
```
mitchellostler, line 56: Since you are not using this value, it won't be changed and the log debug below will always print out 0

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,67 @@
...
+  void prv_button_interrupt_callback(const GpioAddress *button_addr, void *context) {
```
mitchellostler, line 57: You can't actually declare a function inside of another one in C. You'll need to move this outside of the main function.

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  gpio_init_pin(&button_addr, &pot_settings);
+
+  InterruptSettings button_interrupt_settings = {
+    .type = INTERRUPT_TYPE_INTERRUPT,
+    .priority = INTERRUPT_PRIORITY_HIGH,
+  };
+
+  adc_set_channel_pin(button_addr, true);
```
mitchellostler, line 52: The adc should be set for the potentiometer pin, instead of the button

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,67 @@
...
+
+  gpio_it_get_edge(&button_addr, INTERRUPT_EDGE_RISING);
```
mitchellostler, line 54: You actually don't need to call this function. This is only to check for it once you've set the interrupt. Also it is expecting you to pass a pointer to an interrupt edge type (so that it can set the value, and you can read it after the function has been called), not the edge type itself.

mitchellostler (commented):
Looks good, but there are a few issues that will prevent the program from compiling. You should run `make run PROJECT=fw_103_hw PLATFORM=x86`

projects/fw102_example/src/main.c
```diff
@@ -0,0 +1,42 @@
...
+
+int main(void) {
+  interrupt_init();
+  soft_timer_init();
+
+  Counters counting = { 0 };
+  soft_timer_start_millis(COUNTER_LENGTH_MS, prv_increment_a, &counting, NULL);
+
+  while (true) {
+    wait();
```
mitchellostler, line 38: gpio_trigger_interrupt();

mitchellostler: Delay_ms(1000);

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+#include "delay.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "interrupt.h"
+#include "log.h"
+#include "soft_timer.h"
+
+
+// GPIO address that we will allow us to read button data
+const GpioAddress potentiometer_addr_A6 = { // could be passed as context
```
mitchellostler, line 14: Good programming practice is to not to use global variables (ie any variable that is non-static declared outside a function). The const is good since they won't be changed, but declare them static as well `const static`

projects/fw_103_hw/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+
+int main(void) {
+    interrupt_init();
+    gpio_init();
+    gpio_it_init();
+    soft_timer_init();
+    adc_init(ADC_MODE_SINGLE);
+
+    // GPIO Address where button is pressed
+    GpioAddress potentiometer_addr_B2 = {
```
mitchellostler, line 38: This wouldn't be a potentiometer address, this would be a button address, so it would be good to label it as such

mitchellostler (commented):
FW 103 looks good, just a couple of fixes and you should be good to go. Were you able to get it to run?

Also a note: You should really put your name in your branch name, since there are other people using the same ticket number for homework

PR #507, opened 2022-03-14
Arvind firmware 103

A3sanjay:
Completed fw 103 hw. I'm not sure if it is correct though.

### DIFF
--- a/fw_103/src/main.c
+++ b/fw_103/src/main.c
@@ -23,22 +23,22 @@ int main(void) {
   gpio_init();
   gpio_it_init();
 
-  static GpioAddress button_addr = {
+  GpioAddress button_addr = {
     .port = GPIO_PORT_B,
     .pin = 2,
   };
 
-  static GpioAddress adc_address = {
+  GpioAddress adc_address = {
     .port = GPIO_PORT_A,
     .pin = 6,
   };
 
-  static InterruptSettings s_interrupt_settings = {
+  InterruptSettings s_interrupt_settings = {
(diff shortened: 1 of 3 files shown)


fw_103/src/main.c
```diff
@@ -0,0 +1,73 @@
...
+  LOG_DEBUG("ADC Data: %d\n", adc_data);
+  delay_ms(200);
+}
+
+int main(void) {
+  interrupt_init();
+  gpio_init();
+  gpio_it_init();
+
+  static GpioAddress button_addr = {
```
mitchellostler: static declarations should only be used outside of functions, unless you have a good reason for using it in a function

fw_103/src/main.c
```diff
@@ -0,0 +1,73 @@
...
+    GPIO_DIR_IN,
+    GPIO_STATE_HIGH,
+    GPIO_RES_NONE,
+    GPIO_ALTFN_ANALOG,
+  };
+
+  gpio_init_pin(&button_addr, &button_settings);
+  gpio_init_pin(&adc_address, &adc_settings);
+  adc_set_channel_pin(adc_address, true);
+  adc_init(ADC_MODE_SINGLE);
```
mitchellostler, line 57: Adc init should be called before any adc methods

mitchellostler (changes requested):
Just the two comments, otherwise looks good

mitchellostler:
Looks Good! I'll see if I can find you a task today

